Issue #333, opened 2019-11-11
schema_merge does not quite correctly validate merged schemas correctly


aygold92:
In the schema flattener, when there is a combiner (`anyOf`, `allOf`, or `oneOf`), we combine the schemas together in order to make one unified schema.

We currently validate that multiple $refs cant be declared and that multiple types cannot be declared.  However, this misses a few possibilites:
1. both type and $ref are defined.  type can be defined as string but $ref to an object, and we will accept it.  So this schema is valid when it should NOT be:
```
{
       "typeName": "AWS::Valid::TypeName",
        "definitions": {"obj": {"properties": {"Foo": {"type": "object"}}}},
        "properties": {
            "p": {anyOf: [
                     {"type": "string"}, 
                     {"$ref": "#definitions/obj"}
              ]}
        }
}
```
2. Multiple refs go to compatible types.  they can declare different properties, or the same type for the same property.  Actually, these refs need to be merged as well.
So this schema is invalid, when it should be:
```
{
        "typeName": "AWS::Valid::TypeName",
        "definitions": {
            "obj": {"properties": {"a": {"type": "object"}}},
            "obj2": {"properties": {"a": {"type": "object"}}}
        },
        "properties": {
            "p": {
                "allOf": [
                    {"$ref": "#/definitions/obj"},
                    {"$ref": "#/definitions/obj2"}
                ]
            }
        },
    }
```

Oddly enough, it seems the solution is to completely unflatten the schemas in order to validate the constraints properly, and then flatten them back again.

ammokhov:
closing as being tracked in #242

PR #334, opened 2019-11-11
don't allow insertionOrder and uniqueItems to be declared differently for a single property

aygold92:
*Issue #, if available:* N/A

*Description of changes:* When flattening the schema, if we see a combiner (`anyOf`, `oneOf`, or `allOf`), we will merge the schemas together in order to create a unified POJO.  We currently validate that you can't declare multiple types for the same property, or use multiple $refs, as they could be to incompatible types.  Because `uniqueItems` and `insertionOrder` are used to declare the type of an array, they should not be allowed to have different values either.

In addition, there are some unit tests in this PR to assert the behavior described in #333 


tobywf:
Closing in favour of #384 - I wasn't able to update this PR to merge in the Travis CI changes, since the upstream has gone.

PR #344, opened 2019-11-14
Use $id if defined in schema

tobywf:
*Issue #, if available:* N/A

*Description of changes:* If/once we allow `$id` in schemas, we should make use of it.


tobywf:
I think it's because `$id` is a bit like `$ref` - a "language" feature that tells you how to parse a schema, but is sort-of outside the schema itself

tobywf:
I can't update this PR, not sure why - please see #373

Issue #355, opened 2019-11-22
No language plugins found


autarchprinceps:
Whenever I try to do a cfn init, I just get "no language plugins found". I have installed rust, python, scala and nodejs however. Increasing verbosity doesn't give me any more details either
```
╰─ cfn init -v    
Initializing new project
What's the name of your resource type?
(Organization::Service::Resource)
>> Test::Test::Test
No language plugins found

aborted

```

wbingli:
Have you installed any language plugin? e.g. Java plugin,  `pip3 install cloudformation-cli-java-plugin`

jotompki:
Hi, if you look at the pypi pages for those languages, they are all in the planning phase. Currently the [java](https://pypi.org/project/cloudformation-cli-java-plugin/) and [go](https://pypi.org/project/cloudformation-cli-go-plugin/) plugins are functional, and the python plugin is a developer preview. You can check that out [here](https://github.com/aws-cloudformation/aws-cloudformation-rpdk-python-plugin/)

rjlohan:
@jotompki - let's update the `README.md` for this repo to explicitly call out the available language plugins and their development status in a little table. That'll help reduce confusion.

PR #357, opened 2019-11-22
Adding supported plugins to readme

jotompki:
*Issue #, if available:* #355 

*Description of changes:*  Adding in a list of supported plugins under the plugin section for some clarity.


jotompki:
Do you think it's worth moving Usage up before Development? I feel like the first thing I'd want to see on a readme would be how to install

tobywf:
Now that  we're public, I think that makes eminent sense

PR #367, opened 2019-12-12
Add documentation generation functionality

iann0036:
Looking for feedback on functionality to add documentation generation during a `cfn generate`. Will generate only initial README.md without correct types for now.


rjlohan:
Tried running this branch on one of my local repos but hit a bug;

```
Traceback (most recent call last):
  File "/Users/ryanloha/workspaces/cloudformation-cli/src/rpdk/core/cli.py", line 98, in main
    args.command(args)
  File "/Users/ryanloha/workspaces/cloudformation-cli/src/rpdk/core/generate.py", line 16, in generate
    project.generate_docs()
  File "/Users/ryanloha/workspaces/cloudformation-cli/src/rpdk/core/project.py", line 317, in generate_docs
    contents = template.render(type_name=self.type_name, schema=self.schema)
  File "/Users/ryanloha/workspaces/cloudformation-cli/env/lib/python3.7/site-packages/jinja2/asyncsupport.py", line 76, in render
    return original_render(self, *args, **kwargs)
  File "/Users/ryanloha/workspaces/cloudformation-cli/env/lib/python3.7/site-packages/jinja2/environment.py", line 1008, in render
    return self.environment.handle_exception(exc_info, True)
  File "/Users/ryanloha/workspaces/cloudformation-cli/env/lib/python3.7/site-packages/jinja2/environment.py", line 780, in handle_exception
    reraise(exc_type, exc_value, tb)
  File "/Users/ryanloha/workspaces/cloudformation-cli/env/lib/python3.7/site-packages/jinja2/_compat.py", line 37, in reraise
    raise value.with_traceback(tb)
  File "/Users/ryanloha/workspaces/cloudformation-cli/src/rpdk/core/templates/docs-readme.yml", line 63, in top-level template code
    {% if len(identifierptr) == 1 %}
jinja2.exceptions.UndefinedError: 'len' is undefined
```

This is running against the schema from this PR: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4

It's possible that my schema is broken as I haven't finished that, but at the least we probably want to be more tolerant to bad schemas in this doc generation. I think it's a valid schema though, as validation should happen as part of `cfn generate` before it gets to this new doc generation path.

iann0036:
TIL you can use composite keys for the primaryIdentifier...will cater for that.

Working on the sub-property pages and type detection now.

iann0036:
@rjlohan Docs generation is now complete. Seeking full review.

src/rpdk/core/project.py
```diff
@@ -293,6 +296,141 @@ def submit(
...
 
+    def generate_docs(self):
+        # generate the docs folder that contains documentation based on the schema
+        docs_path = "{}/docs".format(self.root)
+
+        if not self.type_info or not self.schema or "properties" not in self.schema:
+            LOG.warning(
+                "Could not generate schema docs due to missing type info or schema"
+            )
+            return
```
tobywf: can this actually happen once the schema has been validated?

src/rpdk/core/project.py
```diff
@@ -293,6 +296,141 @@ def submit(
...
+        if os.path.isdir(docs_path):
+            LOG.debug("Docs directory already exists, recreating...")
+            shutil.rmtree(docs_path)
+        os.mkdir(docs_path)
+
+        LOG.debug("Writing generated docs")
+
+        docs_schema = copy.deepcopy(
+            self.schema
+        )  # take care not to overwrite master schema
```
tobywf: we've been debating this, the safer approach might be to serialize and deserialize to json. this is because some objects may not support deep copying (ref proxies), although we don't currently use them. but it's a fairly common idiom in the codebase:

```bash
$ git grep deepcopy
$ git grep 'json.loads(json.dumps('
src/rpdk/core/contract/resource_client.py:        schema = json.loads(json.dumps(self._schema))
src/rpdk/core/contract/resource_client.py:        schema = json.loads(json.dumps(self._schema))
tests/jsonutils/test_inliner.py:    copy = json.loads(json.dumps(schema))
tests/test_data_loaders.py:    copy = json.loads(json.dumps(BASIC_SCHEMA))
```

tests/test_project.py
```diff
@@ -171,9 +172,158 @@ def test_generate_no_handlers(project):
...
+    docs_dir = Path(project.root / "docs")
+    readme_file = Path(project.root / "docs" / "README.md")
+
+    assert docs_dir.is_dir()
+    assert readme_file.is_file()
+    with patch.object(project, "_plugin", mock_plugin):
+        project.generate()
+        project.generate_docs()
+    with readme_file.open("r", encoding="utf-8") as f:
+        readme_contents = f.read()
```
tobywf: can also use

```python
readme_contents = readme_file.read_text(encoding="utf-8")
```

tests/test_project.py
```diff
@@ -171,9 +172,158 @@ def test_generate_no_handlers(project):
...
+    # tmpdir conflicts with other tests, make a unique one
+    project.root = tmpdir_factory.mktemp("generate_with_docs")
+    mock_plugin = MagicMock(spec=["generate"])
+    with patch.object(project, "_plugin", mock_plugin):
+        project.generate()
+        project.generate_docs()
+    mock_plugin.generate.assert_called_once_with(project)
+
+    docs_dir = Path(project.root / "docs")
+    readme_file = Path(project.root / "docs" / "README.md")
```
tobywf: don't think the conversion is needed? `project.root` should be a `Path` already

tests/test_project.py
```diff
@@ -171,9 +172,158 @@ def test_generate_no_handlers(project):
     mock_plugin = MagicMock(spec=["generate"])
     with patch.object(project, "_plugin", mock_plugin):
         project.generate()
+        project.generate_docs()
     mock_plugin.generate.assert_called_once_with(project)
 
 
+def test_generate_with_docs(project, tmpdir_factory):
```
tobywf: if you use `tmp_path_factory`, it returns a `Path`

iann0036: Changed per your suggestion.

tests/test_project.py
iann0036: Changed per your suggestion.

tests/test_project.py
iann0036: Changed per your suggestion.

src/rpdk/core/project.py
iann0036: Neat trick! Changed per your suggestion.

src/rpdk/core/project.py
```diff
@@ -293,6 +295,29 @@ def submit(
...
+            LOG.debug("Docs directory already exists, recreating...")
+            shutil.rmtree(docs_path)
+        os.mkdir(docs_path)
+
+        LOG.debug("Writing generated docs")
+
+        template = self.env.get_template("docs-readme.yml")
+
+        contents = template.render(type_name=self.type_name, schema=self.schema)
+        readme_path = Path("{}/README.md".format(docs_path))
```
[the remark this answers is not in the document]
iann0036: Changed per your suggestion.

src/rpdk/core/project.py
```diff
@@ -293,6 +295,29 @@ def submit(
                     f, endpoint_url, region_name, role_arn, use_role, set_default
                 )
 
+    def generate_docs(self):
+        # generate the docs folder that contains documentation based on the schema
+        docs_path = "{}/docs".format(self.root)
```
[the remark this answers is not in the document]
iann0036: Changed per your suggestion.

src/rpdk/core/project.py
iann0036: Probably not, but I wanted this for a specific test to fail gracefully. Docs generation shouldn't stop a `cfn generate` in theory if it's passed validation.

src/rpdk/core/project.py
```diff
@@ -293,6 +295,29 @@ def submit(
...
+        if not self.type_info or not self.schema:
+            LOG.warning(
+                "Could not generate schema docs due to missing type info or schema"
+            )
+            return
+
+        if os.path.isdir(docs_path):
+            LOG.debug("Docs directory already exists, recreating...")
+            shutil.rmtree(docs_path)
+        os.mkdir(docs_path)
```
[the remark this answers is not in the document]
iann0036: Changed per your suggestion.

iann0036:
Thanks @tobywf!

Is there a cleaner way to do [this](https://github.com/aws-cloudformation/cloudformation-cli/blob/2c4e92cd480b2937d61e7447f18576ce3d00096a/src/rpdk/core/project.py#L360-L368)? Looked at JsonSchemaFlattener and it seemed close but not quite what I needed to resolve $ref's.

tobywf:
if i'm reading it right, i think John did something similar in #315 with `RefResolver` from `jsonschema`:

src/rpdk/core/project.py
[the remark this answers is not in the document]
tobywf: this still won't work on windows. think this would: `readme_path = docs_path / "README.md"`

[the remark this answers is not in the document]
iann0036: Apologies for missing those other ones. Should have it covered via https://github.com/aws-cloudformation/cloudformation-cli/pull/367/commits/bafad874cf96c6f4dcaa479661789f5d10b03cb8 now.

iann0036:
> if i'm reading it right, i think John did something similar in #315 with `RefResolver` from `jsonschema`:

That `RefResolver` class was perfect! Replaced in https://github.com/aws-cloudformation/cloudformation-cli/pull/367/commits/1a3124859d822b2991bf7b8cd59ebf62aa2c3f11.

iann0036:
@tobywf Fine for you to make direct edits.

Also happy for the docs generation to be opt-in (cfn generate flag?).

From memory, when generating sub-property pages I had it generate AWS::Foo::Bar Spam Eggs format (and not just the last element) in order to avoid having two pages of the exact same title - I believe the AWS:: types always have explicit names for these definitions, but this may not be the case in other schemas. Agree the double up is a bug though.

tobywf:
I'm just going to run it against a bunch of schemas and see what happens. If it works reliably, no need to make it opt-in.

tobywf:
Okay, fixed the issue with different data types, debugging the duplication. It happens with arrays:
```
propname: Foos, proppath: ['Foos'] | prop: {'type': 'array', 'items': {'$ref': '#/definitions/Foo'}}
```

```python
elif prop["type"] == "array":
    prop["arrayitems"] = self._set_docs_properties(
        propname, prop["items"], proppath
    )
```

```
propname: Foos, proppath: ['Foos', 'Foos'] | prop: {'$ref': '#/definitions/Foo'}
```

An idea is using `"Items"` as `propname`? For [AccessAnalyzer](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-accessanalyzer/blob/master/aws-accessanalyzer-analyzer/aws-accessanalyzer-analyzer.json), that results in:

```
archiverules-items-filter-items.md
archiverules-items.md
tags-items.md
```

But not doing the duplication is what I've gone with for now.

iann0036:
If you appended `-items-` to those you'd have to do similar for actual properties (e.g. `someobj-properties-key.md`) so there's no issue when the actual key `Items:` exists.

tobywf:
awesome, I tried it against a bunch of schemas, and it works great. got a bit carried away, we have functions to parse the JSON pointers. i think we'll try and get this out ASAP in the next release.

src/rpdk/core/project.py
```diff
@@ -317,6 +328,126 @@ def submit(
...
+        contents = template.render(
+            type_name=self.type_name, schema=docs_schema, ref=ref, getatt=getatt
+        )
+        self.safewrite(readme_path, contents)
+
+    @staticmethod
+    def _get_docs_primary_identifier(docs_schema):
+        try:
+            primary_id = docs_schema["primaryIdentifier"]
+            if len(primary_id) == 1:
```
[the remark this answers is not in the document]
tobywf: (how) does `!Ref` in CFN work for compound IDs? i'm not sure, so think that's fine for now. you still get to see your documentation

src/rpdk/core/project.py
```diff
@@ -317,6 +328,126 @@ def submit(
...
+                if len(primary_id_path) == 1:
+                    return primary_id_path[0]
+                LOG.warning("Nested primaryIdentifier found")
+        except (KeyError, ValueError):
+            pass
+        return None
+
+    @staticmethod
+    def _get_docs_additional_identifiers(docs_schema):
+        getatt = []
```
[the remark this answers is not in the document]
tobywf: yeah, good shout, missed this cause i was blindly porting it. will fix shortly.

tobywf:
Found another issue: read-only properties should really be excluded, since they can't be specified.

PatMyron:
It could be interesting to include [regional availability information](https://github.com/awsdocs/aws-cloudformation-user-guide/issues/490) here eventually as well

iann0036:
True-up'd with master.

wbingli:
Nice work!

src/rpdk/core/templates/docs-readme.md
```diff
@@ -0,0 +1,105 @@
...
+
+_Pattern_: <code>{{ prop.pattern }}</code>
+{% endif %}
+{% if prop.createonly %}
+
+_Update requires_: [Replacement](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-replacement)
+{% else %}
+
+_Update requires_: [No interruption](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-updating-stacks-update-behaviors.html#update-no-interrupt)
+{% endif %}
```
PatMyron: If we're generating documentation, it's probably worth noting nuance for something as important as potential resource replacement. For example, a version property might be upgradable in-place but not downgradable in-place, so neither of these definitions would be entirely accurate. My assumption would be while schema definitions like `required` and `createOnlyProperties` guarantee a property is required or requires resource replacement to update respectively, other properties not specified as `required` or `createOnlyProperties` may in fact be required or cause resource replacement under some circumstances. [Immutability is important enough that its nuance should probably be captured in the resource schema itself in the future](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/30)

PatMyron: Real world example:
[`Updating AWS::EC2::VPC.InstanceTenancy requires no replacement only if you are updating its value from "dedicated" to "default". Updating AWS::EC2::VPC.InstanceTenancy from "default" to "dedicated" requires replacement`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-ec2-vpc.html#cfn-aws-ec2-vpc-instancetenancy)

src/rpdk/core/project.py
```diff
@@ -67,6 +68,14 @@
 )
 
 
+BASIC_TYPE_MAPPINGS = {
+    "string": "String",
+    "number": "Double",
+    "integer": "Double",
```
PatMyron, line 74: This should be `Integer`, right?
```suggestion
    "integer": "Integer",
```

src/rpdk/core/project.py
```diff
@@ -320,6 +329,128 @@ def submit(
...
+                    return primary_id_path[0]
+                LOG.warning("Nested primaryIdentifier found")
+        except (KeyError, ValueError):
+            pass
+        return None
+
+    @staticmethod
+    def _get_docs_gettable_atts(docs_schema):
+        def _get_property_description(prop):
+            path = fragment_decode(prop, prefix="")
```
PatMyron, line 387: https://github.com/aws-cloudformation/cloudformation-cli/issues/581

PR #373, opened 2019-12-19
Use $id if defined in schema

tobywf:
*Issue #, if available:* Please see #344 , which hasn't fared too well due to the rename.

*Description of changes:* Now that `$id` is allowed in schemas, make use of it. This will make testing `$ref` to other schemas easier, as they don't need to be present on the local filesystem.

### DIFF
--- a/src/rpdk/core/data/schema/provider.definition.schema.v1.json
+++ b/src/rpdk/core/data/schema/provider.definition.schema.v1.json
@@ -210,6 +210,11 @@
         }
     },
     "type": "object",
+    "patternProperties": {
+        "^\\$id$": {
+            "$ref": "https://json-schema.org/draft-07/schema#/properties/$id"
+        }
+    },
     "properties": {
         "$schema": {
             "$ref": "https://json-schema.org/draft-07/schema#/properties/$schema"
--- a/src/rpdk/core/data/schema/provider.definition.schema.v1.json
+++ b/src/rpdk/core/data/schema/provider.definition.schema.v1.json
@@ -1,6 +1,6 @@
 {
     "$schema": "https://json-schema.org/draft-07/schema#",
-    "$id": "provider.definition.schema.v1.json",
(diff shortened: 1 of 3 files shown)


tests/test_data_loaders.py
```diff
@@ -153,6 +163,22 @@ def test_load_resource_spec_file_object_has_name(tmpdir):
...
+def test_load_resource_spec_uses_id_if_id_is_set():
+    @Request.application
+    def application(request):  # pylint: disable=unused-argument
+        return Response(json.dumps({"type": "string"}), mimetype="application/json")
+
+    with wsgi_serve(application) as server:
+        schema = {
+            **BASIC_SCHEMA,
+            "$id": server.url + "/foo",
+            "properties": {"foo": {"$ref": server.url + "/bar"}},
```
vladtsir: Could you add a test to show that relative refs are supported? As in:
 
```
"$ref" : "./remoteschema.json/bar"
```

src/rpdk/core/data/schema/provider.definition.schema.v1.json
```diff
@@ -1,6 +1,6 @@
 {
     "$schema": "https://json-schema.org/draft-07/schema#",
-    "$id": "provider.definition.schema.v1.json",
+    "$id": "https://schema.cloudformation.us-east-1.amazonaws.com/provider.definition.schema.v1.json",
```
jotompki: Is this link valid? Getting access denied when I try to view. Might not be related to this PR, but we should really depend on a central source for the schema just so we don't have to keep copying it over whenever there's changes.

tobywf: Ugh, yep if it's in the schema it should: https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/1fd31cdb3ca76f8418c0209753e204e4e12c019a/src/main/resources/schema/provider.definition.schema.v1.json#L3

PR #375, opened 2019-12-23
increase IAM role max session timeouts to support custom stabilization up to 12 hours

aygold92:
*Issue #, if available:*

This increases the managed provider IAM role to 12 hours, and generates a value for the resource provider role timeout based on the max handler timeout.  This allows for timeouts up to 12 hours

Related: https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/61

### DIFF
--- a/src/rpdk/core/data/managed-upload-infrastructure.yaml
+++ b/src/rpdk/core/data/managed-upload-infrastructure.yaml
@@ -68,7 +68,7 @@ Resources:
   LogAndMetricsDeliveryRole:
     Type: AWS::IAM::Role
     Properties:
-      MaxSessionDuration: 8400
+      MaxSessionDuration: 43200
       AssumeRolePolicyDocument:
         Version: '2012-10-17'
         Statement:
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -236,15 +236,11 @@ class Project:  # pylint: disable=too-many-instance-attributes
 
             # calculate IAM role max session timeout based on highest handler timeout
             # with some buffer (70 seconds per minute)
-            cud_handlers = [
-                handlers.get(handler, {})
-                for handler in ["create", "update", "delete"]
-                if handlers.get(handler, None)
-            ]
             max_handler_timeout = max(
                 (
                     handler.get("timeoutInMinutes", DEFAULT_ROLE_TIMEOUT_MINUTES)
-                    for handler in cud_handlers
+                    for operation, handler in handlers.items()
+                    if operation in ["create", "update", "delete"]
                 ),
                 default=DEFAULT_ROLE_TIMEOUT_MINUTES,
             )
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -30,7 +30,9 @@ ROLE_TEMPLATE_FILENAME = "resource-role.yaml"
 TYPE_NAME_REGEX = "^[a-zA-Z0-9]{2,64}::[a-zA-Z0-9]{2,64}::[a-zA-Z0-9]{2,64}$"
 
 DEFAULT_ROLE_TIMEOUT_MINUTES = 120  # 2 hours
-MIN_ROLE_TIMEOUT_SECONDS = 900  # 15 minutes
+# min and max are according to CreateRole API restrictions
+# https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html
+MIN_ROLE_TIMEOUT_SECONDS = 3600  # 1 hour
 MAX_ROLE_TIMEOUT_SECONDS = 43200  # 12 hours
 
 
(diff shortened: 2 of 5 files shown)


src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -68,7 +68,7 @@ Resources:
   LogAndMetricsDeliveryRole:
     Type: AWS::IAM::Role
     Properties:
-      MaxSessionDuration: 8400
+      MaxSessionDuration: 43200
```
tobywf: think this one is pretty uncontroversial

src/rpdk/core/templates/resource-role.yml
```diff
@@ -7,7 +7,7 @@ Resources:
   ExecutionRole:
     Type: AWS::IAM::Role
     Properties:
-      MaxSessionDuration: 8400
+      MaxSessionDuration: 43200
```
tobywf: couldn't this be templated from the value in the schema?

aygold92: will try that out

src/rpdk/core/project.py
```diff
@@ -217,15 +221,37 @@ def generate(self):
...
+            ]
+            max_handler_timeout = max(
+                (
+                    handler.get("timeoutInMinutes", DEFAULT_ROLE_TIMEOUT_MINUTES)
+                    for handler in cud_handlers
+                ),
+                default=DEFAULT_ROLE_TIMEOUT_MINUTES,
+            )
+            # max role session timeout must be between 15 minutes and 12 hours
+            role_session_timeout = min(
```
wbingli: I would suggest to validate and give error if the `timeoutInMinutes` configuration is not between 15 ~ 12*60.  Translate the invalid values to a min value and max value would confuse customer thought their configuration accepted, however, the actual value is not as same as configured.

src/rpdk/core/project.py
```diff
@@ -217,15 +221,37 @@ def generate(self):
...
+                MAX_ROLE_TIMEOUT_SECONDS,
+                max(MIN_ROLE_TIMEOUT_SECONDS, 70 * max_handler_timeout),
```
wbingli: Why it's `70` not `60`?

src/rpdk/core/project.py
aygold92: this value will be validated by the schema itself so I don't think that's necessary, and that includes with the buffer we give (multiplication by 70)

src/rpdk/core/project.py
aygold92: for buffer time -- we have the exact same logic on CloudFormation side when setting the session duration for the role

src/rpdk/core/project.py
aygold92: also the schema allows the timeout to be 2 minutes, so we don't want to validate that.  Actually the max role session timeout [must be between 1 hour and 12 hours](https://docs.aws.amazon.com/IAM/latest/APIReference/API_CreateRole.html) so I will have to update this

src/rpdk/core/project.py
```diff
@@ -217,15 +223,37 @@ def generate(self):
...
+            # with some buffer (70 seconds per minute)
+            cud_handlers = [
+                handlers.get(handler, {})
+                for handler in ["create", "update", "delete"]
+                if handlers.get(handler, None)
+            ]
+            max_handler_timeout = max(
+                (
+                    handler.get("timeoutInMinutes", DEFAULT_ROLE_TIMEOUT_MINUTES)
+                    for handler in cud_handlers
```
jotompki: if you just loop over all handler values you can avoid the filtering above with the cud_handlers. The read and list would just default to the default time since it wont be found.

aygold92: we don't want it to be the default time if all values specified are less than the default time

src/rpdk/core/project.py
```diff
@@ -217,15 +223,37 @@ def generate(self):
...
+                for handler in handlers.values()
                 for action in handler.get("permissions", [])
             }
+
+            # calculate IAM role max session timeout based on highest handler timeout
+            # with some buffer (70 seconds per minute)
+            cud_handlers = [
+                handlers.get(handler, {})
+                for handler in ["create", "update", "delete"]
+                if handlers.get(handler, None)
```
jotompki: if we are going to only use the cud handlers, could we change to something like this? something about the current one seems a little roundabout. not a huge deal but seems cleaner
```
max(
(
handler.get("timeoutInMinutes", DEFAULT_ROLE_TIMEOUT_MINUTES)
for operation, handler in handlers.items() if operation in ["create", "update", "delete]
)
, 
default.....
)
```

aygold92: ah yeah, that is better 👍

PR #380, opened 2020-01-10
Add resource model pruning function and fix contract test

jotompki:
*Issue #, if available:*

*Description of changes:* Contract tests for certain providers were failing the delete_create duplicate test. It turns out the prune_properties function used to try to remove the readOnly properties from both the deleted model and the newly created model wasn't actually pruning anything because the paths included the "/properties" prefix. This adds a function for pruning models (that do not have a "properties" key) and uses that function in the contract test. 

The test is also skipped if the schema does not contain any writable identifiers/

### DIFF
--- a/src/rpdk/core/contract/suite/handler_delete.py
+++ b/src/rpdk/core/contract/suite/handler_delete.py
@@ -79,7 +79,7 @@ def contract_delete_delete(resource_client, deleted_resource):
 @pytest.mark.create
 @pytest.mark.delete
 def contract_delete_create(resource_client, deleted_resource):
-    if resource_client.has_writable_identifiers():
+    if resource_client.has_writable_identifier():
         deleted_model, request = deleted_resource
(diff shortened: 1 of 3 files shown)


src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -79,11 +79,16 @@ def contract_delete_delete(resource_client, deleted_resource):
 @pytest.mark.create
 @pytest.mark.delete
 def contract_delete_create(resource_client, deleted_resource):
-    deleted_model, request = deleted_resource
-    response = test_create_success(resource_client, request)
-
-    # read-only properties should be excluded from the comparison
-    prune_properties(deleted_model, resource_client.read_only_paths)
-    prune_properties(response["resourceModel"], resource_client.read_only_paths)
+    if resource_client.has_writable_identifiers():
```
jaymccon: should be `has_writable_identifier`

PR #386, opened 2020-01-30
Allow external use of write_settings

stilvoid:
*Issue #, if available:* N#A

*Description of changes:* Allows plugins to call `write_settings` on the project. We need this in the Go plugin as we're intending to store plugin version information in `.rpdk-config`. See https://github.com/aws-cloudformation/cloudformation-cli-go-plugin/pull/115

### DIFF
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -73,6 +73,7 @@ class Project:  # pylint: disable=too-many-instance-attributes
         self.root = Path(root) if root else Path.cwd()
         self.settings_path = self.root / SETTINGS_FILENAME
         self.type_info = None
+        self.language = None
         self._plugin = None
         self.settings = None
         self.schema = None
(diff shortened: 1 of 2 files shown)


stilvoid:
The test failure reported by Travis is also failing in the master branch and is not introduced by this PR. I'll rebase when it's fixed.

stilvoid:
> The test failure reported by Travis is also failing in the master branch and is not introduced by this PR. I'll rebase when it's fixed.

Fix in #388

tobywf:
I can't think of/remember a reason not to do this, but I also can't figure out why we didn't do this in the first place - and there must of been a reason for us to go out of our way to not have `language` be a project attribute in such a blatant way. Let me try and sync up with John about this.

tobywf:
Yeah, I really can't remember. Oh well, ship it :D

Issue #387, opened 2020-01-30
Fix Jinja env test


stilvoid:
`test_language_plugin_setup_jinja_env_defaults` in `tests/test_plugin_base.py` is currently failing in the master branch. This impacts all PRs for now.

PR #388, opened 2020-01-30
Use PEP451 loader to work around pytest/jinja2 disagreements (#387)

stilvoid:
*Issue #, if available:* 387

*Description of changes:* Work around update to Jinja2 that breaks pytest's mangling of module loaders.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -7,8 +7,7 @@ docutils>=0.14
 # testing tools
 pylint>=2.3.1
 coverage>=4.5.4
-# https://github.com/pytest-dev/pytest/issues/5392
-pytest==4.5.0
+pytest>=5.3.5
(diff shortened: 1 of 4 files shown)


tobywf:
Possibly it was this commit: https://github.com/pallets/jinja/pull/1113/

This seems to fix the lower pytest version we were needing to use (https://github.com/pytest-dev/pytest/issues/5392), this was always a bit hacky it seems. I don't quite understand in which cases `spec` or `spec.origin` can be `None`, and did toy around with the idea of removing `PackageLoader` entirely.

The only other things I did was split the inline `if`, it seems to not give branch coverage, and use `pathlib` - I don't think we use `os` anywhere in the core lib, so it just keeps things consistent.

PR #389, opened 2020-02-05
Enforce secure transport policy on artifacts bucket

vladtsir:
*Issue #, if available:* n/a

*Description of changes:*

Enforce "secure transport" bucket policy on the artifacts bucket created in the user account

### DIFF
--- a/src/rpdk/core/data/managed-upload-infrastructure.yaml
+++ b/src/rpdk/core/data/managed-upload-infrastructure.yaml
@@ -38,6 +38,24 @@ Resources:
               - !Sub "arn:${AWS::Partition}:s3:::${ArtifactBucket}"
               - !Sub "arn:${AWS::Partition}:s3:::${ArtifactBucket}/*"
 
+  ArtificatBucketSecureCopyPolicy:
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref ArtifactBucket
+      PolicyDocument:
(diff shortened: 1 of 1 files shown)


jotompki:
Have you tested end to end submitting a resource type? Want to make sure registration still works with this condition

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -38,6 +38,24 @@ Resources:
               - !Sub "arn:${AWS::Partition}:s3:::${ArtifactBucket}"
               - !Sub "arn:${AWS::Partition}:s3:::${ArtifactBucket}/*"
 
+  ArtificatBucketSecureCopyPolicy:
+    Type: AWS::S3::BucketPolicy
+    Properties:
+      Bucket: !Ref ArtifactBucket
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
```
jotompki: This could also be added to the above bucket policy as another statement

vladtsir: Folded "secure transport" policy into ArtifactCopyPolicy.

vladtsir:
> Have you tested end to end submitting a resource type? Want to make sure registration still works with this condition

Yes, I was able to perform `cfn submit` and confirmed that the bucket policy was updated

PR #390, opened 2020-02-05
Bring schema/README.md in sync with aws-cloudformation-resource-schema

vladtsir:
*Issue #, if available:* n/a

*Description of changes:*

README.md is updated to stay in sync with the master version in aws-cloudformation-resource-schema repository

### DIFF
--- a/src/rpdk/core/data/schema/README.md
+++ b/src/rpdk/core/data/schema/README.md
@@ -1,16 +1,18 @@
-## Provider Definition Schema
+## AWS CloudFormation Resource Schema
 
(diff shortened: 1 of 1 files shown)


tobywf:
So we've discussed this. First, the `README` in this repo could be a link. Second, the fact that it's manual is actually good - this way, we never release an updated meta-schema before the backend has received it. Probably not good to release a new CloudFormation CLI version that allows for new schema elements that the upload then rejects.

rjlohan:
> The `README` in this repo could be a link to https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/master/README.md ?
> 
> So Vlad and me discussed this (offline). The fact that it's manual is actually good - this way, we never release an updated meta-schema before the backend has received it. Probably not good to release a new CloudFormation CLI version that allows for new schema elements that the upload then rejects.

That makes sense, can it be achieved with submodules or dependencies somehow though?

tobywf:
Do we want this merged?

PR #396, opened 2020-02-19
Enable versioning and server access logging for ArtifactBucket

knassef:
*Description of changes:* For security reasons;
- Created new bucket `AccessLogsBucket` to store s3 server access logs of `ArtifactBucket`
- Enabled versioning in both buckets (`ArtifactBucket` and `AccessLogsBucket`)
- Set `LifecycleConfiguration` to maximum value

### DIFF
--- a/src/rpdk/core/data/managed-upload-infrastructure.yaml
+++ b/src/rpdk/core/data/managed-upload-infrastructure.yaml
@@ -19,6 +19,27 @@ Resources:
             Status: Enabled
             AbortIncompleteMultipartUpload:
               DaysAfterInitiation: 1
+      VersioningConfiguration:
+        Status: Enabled
+      LoggingConfiguration:
+        DestinationBucketName: !Ref AccessLogsBucket
+        LogFilePrefix: ArtifactBucket
+
+  AccessLogsBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: LogDeliveryWrite
+      BucketEncryption:
+        ServerSideEncryptionConfiguration:
+        - ServerSideEncryptionByDefault:
+            SSEAlgorithm: aws:kms
+            KMSMasterKeyID: !Ref EncryptionKey
+      LifecycleConfiguration:
+        Rules:
+          - Status: Enabled
+            ExpirationInDays: 3653
+      VersioningConfiguration:
+        Status: Enabled
 
   ArtifactCopyPolicy:
     Type: AWS::S3::BucketPolicy
@@ -55,7 +76,7 @@ Resources:
       Description: KMS key used to encrypt the resource type artifacts
       EnableKeyRotation: true
       KeyPolicy:
-        Version: 2012-10-17
+        Version: "2012-10-17"
         Statement:
         - Sid: Enable full access for owning account
           Effect: Allow
@@ -80,7 +101,7 @@ Resources:
     Properties:
       MaxSessionDuration: 43200
       AssumeRolePolicyDocument:
-        Version: '2012-10-17'
+        Version: "2012-10-17"
         Statement:
         - Effect: Allow
           Principal:
@@ -92,7 +113,7 @@ Resources:
       Policies:
       - PolicyName: LogAndMetricsDeliveryRolePolicy
         PolicyDocument:
-          Version: '2012-10-17'
+          Version: "2012-10-17"
           Statement:
           - Effect: Allow
             Action:


src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -19,6 +19,31 @@ Resources:
             Status: Enabled
             AbortIncompleteMultipartUpload:
               DaysAfterInitiation: 1
+      VersioningConfiguration:
+        Status: Enabled
+      LoggingConfiguration:
+        DestinationBucketName: { 'Fn::Sub' : "artifact-bucket-access-logs-${AWS::AccountId}-${AWS::Region}" }
```
ikben: This should probably be !Ref AccessLogsBucket ?
- Will setting this work if the AccessLogBucket does not exist?
- Will setting this work if the permissions are wrong (if not, it will also need a DependsOn)

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -19,6 +19,31 @@ Resources:
...
+        - ServerSideEncryptionByDefault:
+            SSEAlgorithm: aws:kms
+            KMSMasterKeyID: !Ref EncryptionKey
+      LifecycleConfiguration:
+        Rules:
+          - Status: Enabled
+            ExpirationInDays: 3653
+      VersioningConfiguration:
+        Status: Enabled
+      LoggingConfiguration:
```
ikben: Is logging on the AccessLogBucket required? If these lines can be removed, then there is no need to set the BucketName anymore, which would be a better template

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -19,6 +19,31 @@ Resources:
             Status: Enabled
             AbortIncompleteMultipartUpload:
               DaysAfterInitiation: 1
+      VersioningConfiguration:
+        Status: Enabled
+      LoggingConfiguration:
+        DestinationBucketName: !Ref AccessLogsBucket
```
knassef: @ikben you are right, this is better and we do not need depends on.

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: Yes, it is a critical requirement on our side. I understand what you mean regarding the template and this is the least intervening way I could think of.

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: @ikben you are right, fixed.
- It will not work if AccessLogBucket does not exist
- also will not work if permissions are wrong
no need for DependsOn

rjlohan:
Thanks for this contribution, this is a nice improvement. 👍

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: I think we might be using two different version for "works". I was talking about getting no errors in CloudFormation. If you can't put the LoggingConfiguration without having the permissions already there the dependsOn is needed, I assume this is not the case.

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: It is probably the best way to enable it. I'm not convinced that pushing this to all users of the cloudformation-cli is the way to go. Not only because the template, but also because now you've created a loop and logs will cause more logs to be written. Or like how they describe it in the S3 documentation:

> When your source bucket and target bucket are the same bucket, additional logs are created for the logs that are written to the bucket. This behavior might not be ideal for your use case because it could result in a small increase in your storage billing. In addition, the extra logs about logs might make it harder to find the log that you're looking for.

 Would enabling CloudTrail DataEvents for either bucket fulfil your requirements? That's something  you can do  outside of this template.

Adding a "DestinationBucketName"  Parameter might also be way to allow different requirements for access logs, without putting lots of options in the main template

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: You are right. Would you prefer to have CFN failing incase the policy does not exist then? dependsOn will add this constraint, correct?

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: could you maybe explain more about your last suggestion regarding Adding a "DestinationBucketName" Parameter?

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: DependsOn will make sure that the AccessLogsPolicy is created before CloudFormation attempts to create the ArtifactBucket, so that the policy will always exist before the LoggingConfiguration is created / updated

rjlohan: > DependsOn will make sure that the AccessLogsPolicy is created before CloudFormation attempts to create the ArtifactBucket, so that the policy will always exist before the LoggingConfiguration is created / updated

Does that work though? The Policy needs to `!Ref` the Bucket ARN for a bunch of settings?

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: You could add a Parameter that can be changed in the CloudFormation console. The cloudformation-cli code might need be checked to make   sure it uses UsePreviousValue for all parameters when doing that though. The template could look like this

```
# [...]
Parameters:
  DestinationBucketName:
    Description: Set this if you want to write access logs to a bucket you manage
    Type: String
    Default: ""
Conditions:
  CreateLogBucket: !Not [!Equals [!Ref DestinationBucketName,  ""]]
Resources:
  ArtifactBucket:
    Type: AWS::S3::Bucket
    Properties:
      # [...]
      LoggingConfiguration:
        DestinationBucketName: !If [CreateLogBucket, !Ref AccessLogsBucket, !Ref DestinationBucketName]
        LogFilePrefix: ArtifactBucket

  AccessLogsBucket:
    Condition: CreateLogBucket
    Type: AWS::S3::Bucket
    # [..]
# [...]
```
(Not shown: removing the BucketName and LoggingConfiguration from the AccessLogsBucket too).

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: I just noticed that `AccessControl: LogDeliveryWrite` is in the Properties of `AccessLogsBucket`. I assume this is to make it possible to write to itself. If I understand this correctly than there are two options to make the template work:

- Keep the `AccessControl` and remove the whole `AccessLogsPolicy` as both do the same thing
- Remove the `AccessControl` and add the `DependsOn`. This will only be possible when not writing to itself (see other comment)

ikben:
@knassef I promise I didn't set out to nitpick every line of this. There's nothing wrong with the template in general, but I have a tendency to keep asking questions

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: I see. would I then need to refactor this part https://github.com/aws-cloudformation/cloudformation-cli/blob/fd8f1dc65633db9c910efac96dc9f466f557c1b0/src/rpdk/core/upload.py#L26
and pass the parameter there?

knassef:
@ikben ah no worries at all, we are having a reasonable conversation here, trying to understand pros and cons of different approaches. so all good :)

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: > Does that work though? The Policy needs to `!Ref` the Bucket ARN for a bunch of settings?

This is how I read the template

```
ArtifactBucket -------------Ref-----------> AccessLogsBucket
    \---Depend---> AccessLogsPolicy ---Sub-->/
```
That does not create a circle (AccessLogsBucket does not need to know about ArtifactBucket)

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: I propose to not override the default with the cloudformation cli, so I wouldn't refactor to pass a parameter (but  maybe it should be something you can pass to the cli, @rjlohan will have a more  informed opinion on that).

Unfortunately this line will throw errors if there is a parameter in the template and you do not specify it when you update. https://github.xcom/aws-cloudformation/cloudformation-cli/blob/fd8f1dc65633db9c910efac96dc9f466f557c1b0/src/rpdk/core/upload.py#L98

Adding this to that api call should solve that
```
parameters=[{"ParameterKey": "DestinationBucketName", "UsePreviousValue": True}]
```

jotompki: The approach being logging in the same bucket

knassef:
@jotompki yes, my team is using this approach with all S3 bucket as we have a hard security requirement to log all s3 buckets (with no exceptions). AFAIK CloudTrail also dumps logs into an s3 bucket so we will face the same issue with that approach as well. I have addressed the other comments and now `DestinationBucketName` is an optional parameter and logging will only be enabled if it was passed. Do you think we can come to an agreement on this?

knassef:
would appreciate any response regarding my last comment :)

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -19,6 +27,32 @@ Resources:
...
+        Status: Enabled
+      LoggingConfiguration:
+        DestinationBucketName: !If [CreateLogBucket, !Ref AccessLogsBucket, !Ref DestinationBucketName]
+        LogFilePrefix: ArtifactBucket
+
+  AccessLogsBucket:
+    Condition: CreateLogBucket
+    Type: AWS::S3::Bucket
+    Properties:
+      BucketName: !Ref DestinationBucketName
```
jotompki: so if we create the bucket, we set a bucket name of "" that logs to ""? That will throw some validation errors on the CloudFormation end. 

I'm all for adding an access logs bucket just don't think it is necessary for the logs bucket to log to itself.

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -3,6 +3,14 @@ Description: >
   This CloudFormation template provisions all the infrastructure that is
   required to upload artifacts to CloudFormation's managed experience.
 
+Parameters:
+  DestinationBucketName:
+    Description: Set this if you want to write access logs to a bucket you manage
+    Type: String
+    Default: ""
```
PatMyron:
```suggestion
  DestinationBucketName:
    Description: Set this if you want to write access logs to a bucket you manage
    Type: String
    Default: ""
    AllowedPattern: "^[a-z0-9.-]*$"
```

We can't be as strict as the [AWS::S3::Bucket.BucketName AllowedPattern](https://github.com/aws-cloudformation/cfn-python-lint/pull/1208) because of the empty default, but we can at least constrain with some [AllowedPattern](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/parameters-section-structure.html#parameters-section-structure-properties) to prevent invalid S3 bucket name characters

PatMyron (dismissed):
This fails to successfully create any CloudFormation stack regardless if the `DestinationBucketName` parameter is specified or not. Even after negating the `CreateLogBucket` condition, this template still fails to successfully create a CloudFormation stack if the `DestinationBucketName` parameter is not specified

@knassef please test not only that these CloudFormation stacks successfully deploy but also the desired behavior of both specifying `DestinationBucketName` and not specifying `DestinationBucketName`

This bucket should also enforce secure transport: https://github.com/aws-cloudformation/cloudformation-cli/pull/389

src/rpdk/core/data/managed-upload-infrastructure.yaml
ikben: Quick AllowedPattern note:`^[a-z0-9][a-z0-9.-]*[a-z0-9]$|^$` should catch both cases

knassef:
To unblock this, I have removed the self logging of the logs bucket to keep things simple for this PR as I see its not the general use case and I will try to think of another way to keep it optional without complicating the template.

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: please check https://github.com/aws-cloudformation/cloudformation-cli/pull/396#issuecomment-598676122

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -19,6 +27,32 @@ Resources:
             Status: Enabled
             AbortIncompleteMultipartUpload:
               DaysAfterInitiation: 1
+      VersioningConfiguration:
+        Status: Enabled
+      LoggingConfiguration:
+        DestinationBucketName: !If [CreateLogBucket, !Ref AccessLogsBucket, !Ref DestinationBucketName]
+        LogFilePrefix: ArtifactBucket
```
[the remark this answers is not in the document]
knassef: please check https://github.com/aws-cloudformation/cloudformation-cli/pull/396#issuecomment-598676122

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -3,6 +3,14 @@ Description: >
   This CloudFormation template provisions all the infrastructure that is
   required to upload artifacts to CloudFormation's managed experience.
 
+Parameters:
+  DestinationBucketName:
+    Description: Set this if you want to write access logs to a bucket you manage
+    Type: String
+    Default: ""
+Conditions:
+  CreateLogBucket: !Equals [!Ref DestinationBucketName,  ""]
```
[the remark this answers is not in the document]
knassef: please check https://github.com/aws-cloudformation/cloudformation-cli/pull/396#issuecomment-598676122

src/rpdk/core/data/managed-upload-infrastructure.yaml
knassef: please check https://github.com/aws-cloudformation/cloudformation-cli/pull/396#issuecomment-598676122

knassef:
Thanks you as well everyone! 🙏

PR #409, opened 2020-03-21
Documentation generation to ignore read-only properties and remove autoescaping Jinja

iann0036:
*Description of changes:*

Relating to #367 , I though we had done the change for read-only properties (https://github.com/aws-cloudformation/cloudformation-cli/pull/367#issuecomment-578320188) but apparently not.

This PR fixes that. cc @wbingli @rjlohan 

Also removing the Jinja autoescaping as the docs templates contain HTML links.


rjlohan:
Wouldn't it be useful to keep the `readOnlyProperties` listed somewhere for `!GetAtt` target information?

iann0036:
The current state looks like this: https://gist.github.com/iann0036/508de64a3e4fc8fb66b22c5ada7fa327

With the PR, it'd look like this: https://gist.github.com/iann0036/bed48c05eb44038b09f2b15118c8f312

Note that the `TPSCode` is now absent from the Properties section but present in the Return Values. This _does_ remove pattern/minLength/maxLength information, but that would also deviate from the current look and feel of the existing CloudFormation resource docs. Happy to keep that as-is if you feel it's better?

Also, the HTML escaping fix also can be seen.

rjlohan:
This is good 👍

Issue #147, opened 2018-12-12
Fast sanely when credentials are invalid


johnttompkins:
On commands that require interaction with AWS (package and submit), we should fast fail and report some error to the user if their credentials are invalid. Currently the user would have to look in the logs to see this result. A function that checks credentials could be added before these command execute their main portions, and this function could inform the user of no credentials or bad credentials.

PatMyron:
[`botocore.exceptions.ClientError: An error occurred (ExpiredToken) when calling the ListExports operation: The security token included in the request is expired`](https://github.com/aws-cloudformation/cloudformation-cli/pull/411)

PR #411, opened 2020-03-25
Support CloudFormation exports for dynamic values in overrides file

ninavolkmuth:
### Background

The CloudFormation CLI performs resource contract tests using input that is generated from the patterns you define in your resource's property definitions. However, some inputs can't be randomly generated. For example, the AWS::EC2::Instance SubnetId property requires an actual subnet ID for testing, not just a string that matches the appearance of a subnet ID. So in order to test this property, we need to include a file with actual values for the contract tests to use. The file would currently look like this:

```
{
 "CREATE": {
 "/SubnetId": "subnet-0bc6136e" 
 }
}
```

It would be useful if instead of hardcoding the values, we could reference CloudFormation exports. That way you could create the resource dependencies through a CloudFormation stack and reference them easily in the overrides file. 

### Changes

`cfn test` can now take in an overrides.json file that uses jinja. Any jinja variables will be assumed to be CloudFormation exports and are resolved accordingly. 

A sample overrides.json file will look like:
```
{
 "CREATE": {
 "/SubnetId": "{{SubnetExport}}"
 }
}
```

Where `SubnetExport` is the name of the export value. The stack that creates the export would look something like the following:

```
Parameters:
  LambdaService:
    Type: String
Resources:
  VPC:
    Type: "AWS::EC2::VPC"
    Properties:
      CidrBlock: "10.0.0.0/16"
  Subnet:
    Type: "AWS::EC2::Subnet"
    Properties:
      CidrBlock: "10.0.0.0/24"
      VpcId: !Ref VPC
Outputs:
  SubnetId:
    Value: !Ref Subnet
    Export:
      Name: SubnetExport
```

### Testing

* Unit tests I added cover the following cases: 
  * Happy simple case where the export exists on the first page
  * The export is not on the first page when calling list exports
  * Invalid export
  * Export is an integer
* Tested manually with my own resource

### DIFF
--- a/src/rpdk/core/test.py
+++ b/src/rpdk/core/test.py
@@ -57,47 +57,32 @@ def temporary_ini_file():
 def get_cloudformation_exports(region_name, endpoint_url):
     session = create_sdk_session(region_name)
     cfn_client = session.client("cloudformation", endpoint_url=endpoint_url)
-    next_token = ""
+    paginator = cfn_client.get_paginator("list_exports")
+    pages = paginator.paginate()
     exports = {}
-    while next_token is not None:
-        result = (
-            cfn_client.list_exports(NextToken=next_token)
-            if next_token != ""
-            else cfn_client.list_exports()
-        )
-        exports = {
-            **exports,
-            **{export["Name"]: export["Value"] for export in result["Exports"]},
-        }
-        next_token = None if "NextToken" not in result else result["NextToken"]
+    for page in pages:
+        exports.update({export["Name"]: export["Value"] for export in page["Exports"]})
     return exports
 
 
-def render_jinja(overrides_raw, region_name, endpoint_url):
+def render_jinja(overrides_string, region_name, endpoint_url):
     env = Environment(autoescape=True)
-    parsed_content = env.parse(json.dumps(overrides_raw))
+    parsed_content = env.parse(overrides_string)
     variables = meta.find_undeclared_variables(parsed_content)
     if variables:
         exports = get_cloudformation_exports(region_name, endpoint_url)
-        invalid_exports = []
-        var_map = {}
-        for variable in variables:
-            if variable in exports:
-                var_map[variable] = exports[variable]
-            else:
-                invalid_exports.append(variable)
-        if invalid_exports:
+        invalid_exports = variables - exports.keys()
+        if len(invalid_exports) > 0:
             invalid_exports_message = (
                 "Override file invalid: %s are not valid cloudformation exports."
                 + "No Overrides will be applied"
             )
             LOG.warning(invalid_exports_message, invalid_exports)
             return empty_override()
-        overrides_template = Template(json.dumps(overrides_raw))
-        to_return = json.loads(overrides_template.render(var_map))
+        overrides_template = Template(overrides_string)
+        to_return = json.loads(overrides_template.render(exports))
     else:
-        overrides_template = Template(json.dumps(overrides_raw))
-        to_return = json.loads(overrides_template.render())
+        to_return = json.loads(overrides_string)
     return to_return
 
 
@@ -108,7 +93,7 @@ def get_overrides(root, region_name, endpoint_url):
     path = root / "overrides.json"
     try:
         with path.open("r", encoding="utf-8") as f:
-            overrides_raw = render_jinja(json.load(f), region_name, endpoint_url)
+            overrides_raw = render_jinja(f.read(), region_name, endpoint_url)
     except FileNotFoundError:
         LOG.debug("Override file '%s' not found. No overrides will be applied", path)
         return empty_override()
--- a/src/rpdk/core/test.py
+++ b/src/rpdk/core/test.py
@@ -130,7 +130,9 @@ def test(args):
     project = Project()
     project.load()
 
-    overrides = get_overrides(project.root, args.region, args.endpoint_url)
+    overrides = get_overrides(
+        project.root, args.region, args.cloudformation_endpoint_url
+    )
 
     plugin = ContractPlugin(
         ResourceClient(
@@ -167,7 +169,9 @@ def setup_subparser(subparsers, parents):
         "--role-arn", help="Role used when performing handler operations."
     )
 
-    parser.add_argument("--endpoint-url", help="CloudFormation endpoint to use.")
+    parser.add_argument(
+        "--cloudformation-endpoint-url", help="CloudFormation endpoint to use."
+    )
 
     parser.add_argument("passed_to_pytest", nargs="*", help=SUPPRESS)
 
--- a/src/rpdk/core/test.py
+++ b/src/rpdk/core/test.py
@@ -10,11 +10,13 @@ from pathlib import Path
 from tempfile import NamedTemporaryFile
 
 import pytest
+from jinja2 import Environment, Template, meta
 from jsonschema import Draft6Validator
 from jsonschema.exceptions import ValidationError
 
 from rpdk.core.jsonutils.pointer import fragment_decode
 
+from .boto_helpers import create_sdk_session
 from .contract.contract_plugin import ContractPlugin
 from .contract.interface import Action
 from .contract.resource_client import ResourceClient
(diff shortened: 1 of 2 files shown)


src/rpdk/core/test.py
```diff
@@ -52,14 +54,61 @@ def temporary_ini_file():
         yield str(path)
 
 
-def get_overrides(root):
+def get_cloudformation_exports(region_name, endpoint_url):
+    session = create_sdk_session(region_name)
+    cfn_client = session.client("cloudformation", endpoint_url=endpoint_url)
+    next_token = ""
+    exports = {}
+    while next_token is not None:
```
jotompki: [Paginators](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/paginators.html) might be useful here and simplify the logic instead of using the next token approach.

https://boto3.amazonaws.com/v1/documentation/api/latest/reference/services/cloudformation.html#CloudFormation.Paginator.ListExports

src/rpdk/core/test.py
```diff
@@ -52,14 +54,61 @@ def temporary_ini_file():
...
+        result = (
+            cfn_client.list_exports(NextToken=next_token)
+            if next_token != ""
+            else cfn_client.list_exports()
+        )
+        exports = {
```
jotompki: exports.update({comphrehension}) could be a little cleaner

src/rpdk/core/test.py
```diff
@@ -52,14 +54,61 @@ def temporary_ini_file():
...
+
+def render_jinja(overrides_raw, region_name, endpoint_url):
+    env = Environment(autoescape=True)
+    parsed_content = env.parse(json.dumps(overrides_raw))
+    variables = meta.find_undeclared_variables(parsed_content)
+    if variables:
+        exports = get_cloudformation_exports(region_name, endpoint_url)
+        invalid_exports = []
+        var_map = {}
+        for variable in variables:
```
jotompki: could compare keys of exports as set to variables using issuperset to avoid loop and concatenating list. set difference would be invalid exports

src/rpdk/core/test.py
```diff
@@ -52,14 +54,61 @@ def temporary_ini_file():
...
+                "Override file invalid: %s are not valid cloudformation exports."
+                + "No Overrides will be applied"
+            )
+            LOG.warning(invalid_exports_message, invalid_exports)
+            return empty_override()
+        overrides_template = Template(json.dumps(overrides_raw))
+        to_return = json.loads(overrides_template.render(var_map))
+    else:
+        overrides_template = Template(json.dumps(overrides_raw))
+        to_return = json.loads(overrides_template.render())
```
jotompki: not sure how lightweight the render operation is, but shouldn't this just be overrides_raw? could save the render/load

tests/test_test.py
```diff
@@ -167,7 +167,52 @@ def test_get_overrides_good_path(base):
...
+    patch_sdk = patch("rpdk.core.test.create_sdk_session", autospec=True)
+
+    path = base / "overrides.json"
+    with path.open("w", encoding="utf-8") as f:
+        json.dump(overrides, f)
+    with patch_sdk as mock_sdk:
+        mock_sdk.return_value.client.side_effect = [mock_cfn_client, Mock()]
+        result = get_overrides(base, DEFAULT_REGION, None)
+
+    assert result == expected_overrides
```
jotompki: could we add a test for the logic if there are more variables than exports

src/rpdk/core/test.py
ninavolkmuth: Thanks for the suggestion, will use that instead

src/rpdk/core/test.py
ninavolkmuth: Will do. Thought that jinja would throw an error for extra values you pass into render but it doesn't seem like it does

tests/test_test.py
ninavolkmuth: Sure

src/rpdk/core/test.py
ninavolkmuth: Was thinking about if they wanted to just use jinja in general. But thinking about it again, I don't see a use cause where they wouldn't use variables anyway. I'll remove it

src/rpdk/core/test.py
```diff
@@ -133,6 +167,8 @@ def setup_subparser(subparsers, parents):
         "--role-arn", help="Role used when performing handler operations."
     )
 
+    parser.add_argument("--endpoint-url", help="CloudFormation endpoint to use.")
```
jotompki: There's also an endpoint argument for contract tests I would make both a little more explicit in the name so that people don't get confused

https://github.com/aws-cloudformation/cloudformation-cli/blob/33a7fa4ad70b8fe52b635ed19f173f363c054356/src/rpdk/core/test.py#L139-L162

src/rpdk/core/test.py
jotompki: At least for that we'd need a way to pass in the variables. Do you think that'd be feasible or worth doing? I don't really have an opinion

src/rpdk/core/test.py
ninavolkmuth: Hm ok, I was mimicking the standard for specifying service endpoints in aws cli so I thought that might be clear enough but I can make it more explicit

src/rpdk/core/test.py
ninavolkmuth: Until we have a strong use case, I don't think it's worth doing. It would require api changes to pass in custom parameters. If we don't want that, we'd have to let them put the inputs in another text file, which is what the overrides.json is meant for in the first place. I removed it

src/rpdk/core/test.py
```diff
@@ -52,14 +54,46 @@ def temporary_ini_file():
...
 
-def get_overrides(root):
+def get_cloudformation_exports(region_name, endpoint_url):
+    session = create_sdk_session(region_name)
+    cfn_client = session.client("cloudformation", endpoint_url=endpoint_url)
+    paginator = cfn_client.get_paginator("list_exports")
+    pages = paginator.paginate()
+    exports = {}
+    for page in pages:
+        exports.update({export["Name"]: export["Value"] for export in page["Exports"]})
```
aygold92: so we pull all the exports for an account every time we run cfn-test? there doesn't appear to be any limit on the number of exports for an account https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html

src/rpdk/core/test.py
```diff
@@ -52,14 +54,46 @@ def temporary_ini_file():
...
+    if variables:
+        exports = get_cloudformation_exports(region_name, endpoint_url)
+        invalid_exports = variables - exports.keys()
+        if len(invalid_exports) > 0:
+            invalid_exports_message = (
+                "Override file invalid: %s are not valid cloudformation exports."
+                + "No Overrides will be applied"
+            )
+            LOG.warning(invalid_exports_message, invalid_exports)
+            return empty_override()
```
aygold92: why don't we fail immediately when this happens?

ninavolkmuth: I was keeping with the current behavior of cfn test. If there are any issues with the overrides template in general, we will simply not apply any overrides and try to execute the tests. If we want to change it, it probably shouldn't be part of this pr.

src/rpdk/core/test.py
ninavolkmuth: Max number of exports should be 12000. Outputs per stack is 60 and number of stacks is 200. Though I'm assuming outputs behave like parameters and nested stack outputs all count towards the same limit. 

Improving the best case (we can't improve the worst case) will increase the time complexity as well.

src/rpdk/core/test.py
aygold92: seems weird to me but okay

src/rpdk/core/test.py
PatMyron: Seems okay since this function shouldn't get called unless opting into this feature

Should we catch exceptions like [`ExpiredToken`](https://github.com/aws-cloudformation/cloudformation-cli/issues/147#issuecomment-664663330)?

PR #417, opened 2020-04-13
Set default type when generating docs to object

ninavolkmuth:
*Description of changes:* `type` is not a required key when declaring a definition/property. When auto-generating the docs, it assumed that `type` would always be specified. This sets the type to object if it is not specified in the schema which is the default.

Tested with a unit test and a personal type. 

### DIFF
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -416,17 +416,19 @@ class Project:  # pylint: disable=too-many-instance-attributes
         ):
             prop["readonly"] = True
 
-        if prop["type"] in BASIC_TYPE_MAPPINGS:
-            mapped = BASIC_TYPE_MAPPINGS[prop["type"]]
+        prop_type = "object" if "type" not in prop else prop["type"]
+
+        if prop_type in BASIC_TYPE_MAPPINGS:
+            mapped = BASIC_TYPE_MAPPINGS[prop_type]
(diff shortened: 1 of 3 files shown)


src/rpdk/core/project.py
```diff
@@ -416,17 +416,19 @@ def _set_docs_properties(self, propname, prop, proppath):
         ):
             prop["readonly"] = True
 
-        if prop["type"] in BASIC_TYPE_MAPPINGS:
-            mapped = BASIC_TYPE_MAPPINGS[prop["type"]]
+        prop_type = "object" if "type" not in prop else prop["type"]
```
aygold92, line 419: `prop_type = prop.get("type", "object")`
https://www.tutorialspoint.com/python/dictionary_get.htm

tests/test_project.py
```diff
@@ -177,6 +177,29 @@ def test_generate_no_handlers(project):
     mock_plugin.generate.assert_called_once_with(project)
 
 
+def test_generate_no_type_defined(project, tmp_path_factory):
```
aygold92, line 180: this looks like the exact same test as below... can we use `@parametrized`?
http://doc.pytest.org/en/latest/example/parametrize.html

ninavolkmuth: Sure

src/rpdk/core/project.py
ninavolkmuth: Good suggestion, changed

Issue #418, opened 2020-04-17
Doc generator fails on objects with patternProperties


arnulfojr:
Hi!

According to the resource schema, specifying both properties and patternProperties is not allowed and validation fails... https://github.com/aws-cloudformation/aws-cloudformation-resource-schema#constraints

But when having an schema with patternProperties the doc generation fails.
I have this property that is referenced as a definition and the definition looks like:

```
// ignore the regex :)
    "SomeProp": {
      "type": "object",
      "additionalProperties": false,
      "patternProperties": {
        ".*": {
          "type": "string",
          "maxLength": 30000
        }
      }
    },
```

I peek in the code and it seems that the doc generation assumes that all objects have properties and no "patternProperties" exists.

https://github.com/aws-cloudformation/cloudformation-cli/blob/c2f4ff2128a52fa51c9d9bf6cc8c4922da17f831/src/rpdk/core/project.py#L431-L439

The stack trace is:
```
Traceback (most recent call last):
  File ".../lib/python3.6/site-packages/rpdk/core/cli.py", line 98, in main
    args.command(args)
  File ".../lib/python3.6/site-packages/rpdk/core/generate.py", line 16, in generate
    project.generate_docs()
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 353, in generate_docs
    for name, value in docs_schema["properties"].items()
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 353, in <dictcomp>
    for name, value in docs_schema["properties"].items()
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 426, in _set_docs_properties
    propname, prop["items"], proppath
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 437, in _set_docs_properties
    for name, value in prop["properties"].items()
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 437, in <dictcomp>
    for name, value in prop["properties"].items()
  File ".../lib/python3.6/site-packages/rpdk/core/project.py", line 437, in _set_docs_properties
    for name, value in prop["properties"].items()
KeyError: 'properties'
```

ammokhov:
looks like this issue is no longer there, should we close it?

PR #419, opened 2020-04-17
Render patternProperties

arnulfojr:
### Fixes:
https://github.com/aws-cloudformation/cloudformation-cli/issues/418

### Description of changes
Allow patternProperties to be documented and not failing the doc generation if no properties were found.
Escape propname in markdown (in case of patterProperty)

I created this PR to trigger a discussion for the issue linked, looking forward to read your comments!

> I was using Python 3.7 and seems that the pre-commit isort and black are contradicting each other


src/rpdk/core/project.py
```diff
@@ -432,10 +447,20 @@ def _set_docs_properties(self, propname, prop, proppath):
...
             docs_path = self.root / "docs"
 
-            prop["properties"] = {
-                name: self._set_docs_properties(name, value, proppath + (name,))
-                for name, value in prop["properties"].items()
-            }
+            if "properties" in prop:
+                prop["properties"] = {
+                    name: self._set_docs_properties(name, value, proppath + (name,))
+                    for name, value in prop["properties"].items()
```
jotompki, line 453:
```
try:
    prop["properties"] = {
                    name: self._set_docs_properties(name, value, proppath + (name,))
                    for name, value in prop.get("properties", prop["patternProperties"])
                }
except KeyError:
    raise InvalidSchemaPropertyError(
                    "Object property require a properties or patternProperties field"
                )
```
 could consolidate this logic nicely without all the branching.  i don't know if we need to even worry about a schema where neither properties or patternproperties is specified, but it guess it doesn't hurt including it

arnulfojr: The problem with the solution is that `prop["patternProperties"]` will be evaluated before dict fallback happens so the KeyError always happens.

This is what would be in JS
```js
prop.get("properties", () => prop["patternProperties"])
```

Possibile solutions:
```python
            try:
                object_properties = prop["properties"]
                  if "properties" in prop else prop["patternProperties"]
            except KeyError:
                raise InvalidSchemaPropertyError(
                    "Object property requires a properties or patternProperties field"
                )

            prop["properties"] = {
                name: self._set_docs_properties(name, value, proppath + (name,))
                for name, value in object_properties.items()
            }
```

```python
            try:
                prop["properties"] = {
                    name: self._set_docs_properties(name, value, proppath + (name,))
                    for name, value in prop.get(
                        "properties", prop.get("patternProperties")
                    ).items()
                }
            except AttributeError:
                raise InvalidSchemaPropertyError(
                    "Object property requires a properties or patternProperties field"
                )

```

or the above with default to empty list

arnulfojr: I pushed the first possible solution, let me know what you think :)

jotompki:
Looks good

src/rpdk/core/project.py
```diff
@@ -432,9 +449,20 @@ def _set_docs_properties(self, propname, prop, proppath):
             template = self.env.get_template("docs-subproperty.md")
             docs_path = self.root / "docs"
 
+            try:
+                object_properties = (
```
wbingli, line 453: The `object` is valid is without `properties` or `patternProperties`.  See [SNS schema](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-sns/blob/master/aws-sns-topicpolicy/aws-sns-topicpolicy.json#L9). 
Could you help to address this?

arnulfojr: I see! I didn't know that was possible, now I wonder how CDK handles this 🤔
It seems to be that they have it there as a `any` TS type for the L1 and added some types for the L2... Interesting

Anyways, I pushed new rev for the PR

Issue #426, opened 2020-04-27
"cfn test" gives "UnboundLocalError: local variable 'updated_model' referenced before assignment"


miparnisari:
https://github.com/aws-cloudformation/cloudformation-cli/blob/dbf6dc34f1ee52b9baf6d0a78b1e62e7f06f83d7/src/rpdk/core/contract/suite/handler_update.py#L26-L30

I think that snippet is the reason why while running the contract tests I get 

```
ERROR handler_update.py::contract_update_read_success - UnboundLocalError: local variable 'updated_model' referenced before assignment
ERROR handler_update.py::contract_update_list_success - UnboundLocalError: local variable 'updated_model' referenced before assignment
```

jotompki:
Is your update handler returning successfully? I have a hunch that this call is failing so the updated model isn't getting set:
https://github.com/aws-cloudformation/cloudformation-cli/blob/dbf6dc34f1ee52b9baf6d0a78b1e62e7f06f83d7/src/rpdk/core/contract/suite/handler_update.py#L23-L25

then we try using the model during cleanup:
https://github.com/aws-cloudformation/cloudformation-cli/blob/dbf6dc34f1ee52b9baf6d0a78b1e62e7f06f83d7/src/rpdk/core/contract/suite/handler_update.py#L28-L31

Ideally we should just have some cleanup model that we first set to the update request, then the model after creation

miparnisari:
> Is your update handler returning successfully?

No :D but still, I think the `finally` block should be changed to consider this possibility

jotompki:
Yeah that's just a bug, we handle it correctly for the create fixture by setting a model to the requested model then the returned model to handle :

https://github.com/aws-cloudformation/cloudformation-cli/blob/dbf6dc34f1ee52b9baf6d0a78b1e62e7f06f83d7/src/rpdk/core/contract/suite/handler_create.py#L23-L31

I'll send out a quick fix

PR #427, opened 2020-04-27
Fix update contract test fixture

jotompki:
*Issue #, if available:* #426 

*Description of changes:* This fixes the update fixture used in contract tests to be in line with how the create fixture operates:
https://github.com/aws-cloudformation/cloudformation-cli/blob/dbf6dc34f1ee52b9baf6d0a78b1e62e7f06f83d7/src/rpdk/core/contract/suite/handler_create.py#L23-L31

### DIFF
--- a/src/rpdk/core/contract/suite/handler_update.py
+++ b/src/rpdk/core/contract/suite/handler_update.py
@@ -13,12 +13,12 @@ from rpdk.core.contract.suite.handler_commons import (
 
 @pytest.fixture(scope="module")
 def updated_resource(resource_client):
-    create_request = resource_client.generate_create_example()
+    create_request = model = resource_client.generate_create_example()
     try:
         _status, response, _error = resource_client.call_and_assert(
             Action.CREATE, OperationStatus.SUCCESS, create_request
         )
-        created_model = response["resourceModel"]
+        created_model = model = response["resourceModel"]
         update_request = resource_client.generate_update_example(created_model)
         _status, response, _error = resource_client.call_and_assert(
             Action.UPDATE, OperationStatus.SUCCESS, update_request, created_model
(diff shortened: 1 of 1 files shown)


src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -13,22 +13,20 @@
...
         _status, response, _error = resource_client.call_and_assert(
             Action.CREATE, OperationStatus.SUCCESS, create_request
         )
-        created_model = response["resourceModel"]
+        created_model = model = response["resourceModel"]
         update_request = resource_client.generate_update_example(created_model)
         _status, response, _error = resource_client.call_and_assert(
             Action.UPDATE, OperationStatus.SUCCESS, update_request, created_model
         )
         updated_model = response["resourceModel"]
```
ninavolkmuth, line 26: Do you need to also assign the model here as well so it is deleting the updated model if the update call is successful?

jotompki: we could but primary identifier shouldnt change between create and update and that should be included in the successful create.

ninavolkmuth: Thought about it more and there are cases where the delete operation depends on other properties besides the primary identifier so it would be better to assign the model here as well

ninavolkmuth: Thought about it more and there are cases where the delete operation depends on other properties besides the primary identifier so it would be better to assign the model here as well

rjlohan: > Thought about it more and there are cases where the delete operation depends on other properties besides the primary identifier so it would be better to assign the model here as well

That would be in violation of the contract. Both Read and Delete handlers MUST operate on nothing more than the fields specified as primaryIdentifier in the schema. That's all they are guaranteed to receive.

aygold92: I don't think we ever specified that anywhere yet, but you are right we need to define a contract for handler input

Issue #228, opened 2019-05-17
Support disconnected development


rjlohan:
Currently, at least the `generate` command fails when operating disconnected. Partly because of JSON Schema fetch issues. Perhaps we can treat these errors as `WARN` level and allow a best-effort codegen to work when disconnected? Or, once successful a local cache can be used in favour of live fetch of remote assets on every build.

PatMyron:
> Or, once successful a local cache can be used in favour of live fetch of remote assets on every build.

[`requests_cache`](https://pypi.org/project/requests-cache/) works well if we're using `requests`

wbingli:
Quick search, it can be fixed use local file for schema reference, see https://github.com/Julian/jsonschema/issues/98

wbingli:
Fixed under #436

PR #436, opened 2020-05-12
Use local meta schema for reference to support disconnected network build

wbingli:
*Issue #, if available:* #228 

*Description of changes:*.  Currently `cfn generate` will do schema validation, and therefore it will require network to get the meta schema reference from internet `https://json-schema.org/draft-07/schema`.

This PR get the meta schema from `https://json-schema.org/draft-07/schema` and store with the resource schema, therefore, it will resolve it from local file instead from internet.

Also it provides a better meta schema locking as we source control the meta schema now.

### DIFF
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -1,11 +1,11 @@
 import json
 import logging
+import os
 import shutil
 from io import TextIOWrapper
 from pathlib import Path
 
 import pkg_resources
-import requests
 import yaml
 from jsonschema import Draft7Validator, RefResolver
 from jsonschema.exceptions import RefResolutionError, ValidationError
@@ -14,7 +14,6 @@ from .exceptions import InternalError, SpecValidationError
 from .jsonutils.inliner import RefInliner
 
 LOG = logging.getLogger(__name__)
-
 
 TIMEOUT_IN_SECONDS = 10
 STDIN_NAME = "<stdin>"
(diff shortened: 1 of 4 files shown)


PatMyron:
[`isort` failed btw](https://travis-ci.com/github/aws-cloudformation/cloudformation-cli/jobs/332287936#L451)

wbingli:
> [`isort` failed btw](https://travis-ci.com/github/aws-cloudformation/cloudformation-cli/jobs/332287936#L451)

Hmm,  lots of changes are not in this PR. Seems something changes in isort.

wbingli:
The `isort` issue is fixed in PR #437 .

src/rpdk/core/data_loaders.py
```diff
@@ -52,24 +51,27 @@ def copy_resource(package_name, resource_name, out_path):
...
+        meta_schema = Path(os.path.dirname(os.path.realpath(__file__))).joinpath(
+            "data/schema/meta-schema.json"
+        )
+        return json.load(meta_schema.open())
 
     resolver = RefResolver(
         base_uri=base_uri,
         referrer=schema,
-        handlers={"http": get_with_timeout, "https": get_with_timeout},
+        handlers={"http": get_from_local, "https": get_from_local},
```
ammokhov, line 67: why do you need protocols if it's reading from the disk now?

wbingli: As the reference in the schema is still pointing to the external HTTPS link.  So the handler is trying to override the protocol handler to retrieve from disk.

PR #443, opened 2020-05-20
Assert CRUDL Handlers complete in expected time

anshikg:
*Issue #, if available:*

*Description of changes:* Create, Update, Delete Handler should complete in 60 seconds. Read and List handler should complete in 30 seconds. Asserting this in contract test

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -1,6 +1,7 @@
 # pylint: disable=import-outside-toplevel
 import json
 import logging
+import time
 from time import sleep
 from uuid import uuid4
 
@@ -234,6 +235,13 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     def generate_token():
         return str(uuid4())
 
+    @staticmethod
+    def assert_time(start_time, end_time, action):
+        timeout_in_seconds = 60
+        if action in (Action.READ, Action.LIST):
+            timeout_in_seconds = 30
+        assert end_time - start_time <= timeout_in_seconds
+
     def _make_payload(self, action, request):
         return {
             "credentials": self._creds.copy(),
@@ -258,7 +266,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         if assert_status not in [OperationStatus.SUCCESS, OperationStatus.FAILED]:
             raise ValueError("Assert status {} not supported.".format(assert_status))
         request = self.make_request(current_model, previous_model, **kwargs)
+
+        start_time = time.time()
         status, response = self.call(action, request)
+        self.assert_time(start_time, time.time(), action)
         if assert_status == OperationStatus.SUCCESS:
             self.assert_success(status, response)
             error_code = None
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -1,5 +1,6 @@
 # fixture and parameter have the same name
 # pylint: disable=redefined-outer-name,protected-access
+import time
 from io import StringIO
 from unittest.mock import ANY, patch
 
@@ -543,3 +544,11 @@ def test_has_update_handler(resource_client):
     schema = {"handlers": {"update": {"permissions": ["permission"]}}}
     resource_client._update_schema(schema)
     assert resource_client.has_update_handler()
+
+
+def test_assert_create_time(resource_client):
+    resource_client.assert_time(time.time() - 30, time.time(), "CREATE")
+
+
+def test_assert_read_time(resource_client):
+    resource_client.assert_time(time.time() - 10, time.time(), "READ")


src/rpdk/core/contract/resource_client.py
```diff
@@ -234,6 +235,13 @@ def make_request(desired_resource_state, previous_resource_state, **kwargs):
     def generate_token():
         return str(uuid4())
 
+    @staticmethod
+    def assert_time(start_time, end_time, action):
+        timeout_in_seconds = 60
+        if action in (Action.READ, Action.LIST):
+            timeout_in_seconds = 30
```
aygold92, line 242: `timeout_in_seconds = 30 if (action in Action.Read, Action.List) else 60`

tests/contract/test_resource_client.py
```diff
@@ -543,3 +544,11 @@ def test_has_update_handler(resource_client):
     schema = {"handlers": {"update": {"permissions": ["permission"]}}}
     resource_client._update_schema(schema)
     assert resource_client.has_update_handler()
+
+
+def test_assert_create_time(resource_client):
```
aygold92, line 549: can we add negative tests?  You should use `@parametrized` so we test all CRUDL

src/rpdk/core/contract/resource_client.py
```diff
@@ -234,6 +235,13 @@ def make_request(desired_resource_state, previous_resource_state, **kwargs):
...
+        assert end_time - start_time <= timeout_in_seconds
```
aygold92, line 243: can we add some kind of meaningful message if the handler fails this assertion?

anshikg: do you mean Log before we assert?

tests/contract/test_resource_client.py
anshikg: 1) I am unsure how we would run a negative test for time constraint. In what condition can we expect a handler to respond in more than 30/60 seconds?
2) Correct me if I am wrong, as far as I know this is being tested for all CRUDL apis right now.

anshikg:
> Will this be asserted for every handler invocation throughout all the contract tests (as opposed to just at the end of each whole test)?

Yes, this is running for all test cases. It made sense for to test it across the board.

src/rpdk/core/contract/resource_client.py
```diff
@@ -234,6 +235,13 @@ def make_request(desired_resource_state, previous_resource_state, **kwargs):
     def generate_token():
         return str(uuid4())
 
+    @staticmethod
+    def assert_time(start_time, end_time, action):
+        timeout_in_seconds = 60
+        if action in (Action.READ, Action.LIST):
```
jotompki, line 241: i'm not sure how useful this assertion will be since the sam cli local cold start time is a lot longer than on the lambda service and varies depending on if the latest docker lambda image is downloaded. we already account for this in the [client](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L87). i could see this becoming frustrating for people trying to test if their machine is slow. maybe we can have this as an option to enable?

src/rpdk/core/contract/resource_client.py
aygold92: what will the customer see when this assertion fails?  I'm guessing it will just say "assertion error: line 243" with no useful info.  I'm pretty sure assert takes a message parameter, or if not maybe logging will work? As long as the customer can clearly see why their test failed

tests/contract/test_resource_client.py
aygold92: i meant for these unit tests

src/rpdk/core/contract/resource_client.py
aygold92: it should not be optional. is there any better way to account for the cold start time when checking the time?  How long is the cold start time?

aygold92:
> Yes, this is running for all test cases. It made sense for to test it across the board.

I mean, if there are multiple invocations within 1 test case, will it test each invocation individually?

tests/contract/test_resource_client.py
anshikg: My bad. Let me go ahead and add that in.

src/rpdk/core/contract/resource_client.py
anshikg: makese sense.

jotompki: pytest is actually pretty detailed and prints out entire stack trace and values

anshikg:
> > Yes, this is running for all test cases. It made sense for to test it across the board.
> 
> I mean, if there are multiple invocations within 1 test case, will it test each invocation individually?

If there are multiple calls for Create or Update, each call will be timed and checked against the timeout value.

src/rpdk/core/contract/resource_client.py
aygold92: yeah but will it tell them that their handler timed out/took too long? as opposed to just saying "70 > 60"

jotompki:
Superseded by #444

PR #444, opened 2020-05-20
Contract test

anshikg:
*Issue #, if available:*

*Description of changes:* Create, Update, Delete Handler should complete in 60 seconds. Read and List handler should complete in 30 seconds. Asserting this in contract test

I misconfigured my head repo and it does not allow me to push to the contract_test branch: https://github.com/aws-cloudformation/cloudformation-cli/pull/443/commits/4c780955ddb66e0191738987a572644d6e6e773f

I have created a new pull request. I have linked my origin PR above. 

### DIFF
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -546,19 +546,23 @@ def test_has_update_handler(resource_client):
     assert resource_client.has_update_handler()
 
 
-def test_assert_create_time(resource_client):
-    resource_client.assert_time(time.time() - 30, time.time(), "CREATE")
+@pytest.mark.parametrize("action", [Action.CREATE, Action.UPDATE, Action.DELETE])
+def test_assert_CUD_time(resource_client, action):
+    resource_client.assert_time(time.time() - 59, time.time(), action)
 
 
-def test_assert_read_time(resource_client):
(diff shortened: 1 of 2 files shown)


tests/contract/test_resource_client.py
```diff
@@ -543,3 +544,21 @@ def test_has_update_handler(resource_client):
     schema = {"handlers": {"update": {"permissions": ["permission"]}}}
     resource_client._update_schema(schema)
     assert resource_client.has_update_handler()
+
+
+def test_assert_create_time(resource_client):
+    resource_client.assert_time(time.time() - 30, time.time(), "CREATE")
```
aygold92, line 550: the point of this unit test is to accomplish two things:
* prove that the code you wrote actually does what it intended
* ensure that if the code does change, it does not have unintended consequences.  In other words, if this test starts to fail, either it was on purpose and they need to update the test, or the test failure will show they did something wrong.

I'm not sure these tests accomplish either one.  They don't test every handler, and the timeout value could change a lot without causing this test to fail.

some suggestions:
* you can use `@pytest.mark.parametrize` for this test to do it for CUD actions.
* I would use 59 seconds instead 30.  Do same for the read/list at 29 seconds.  
* parameterize the failure tests at 61/31 seconds.

PR #446, opened 2020-05-27
Desired state stabilization contract test

anshikg:
*Issue #, if available:*

*Description of changes:* Following are changes made in this PR:

1. Check if Primary Identifier is added after create finishes successfully
~~2. Check  Primary Identifier should not exist after unsuccessful create finishes~~
3. Check Primary Identifier after update has not changed.
4. Assert ResourceModel does not exist  in response after delete.
5. Test case where update fails with NotUpdatable error code when, we try to update an existing resource with readOnly/createOnly property
6. Test case where update fails with NotFound error code when the resource does not exists.
7. Check if Primary Identifier is added after in_progress is returned.

*Test*:
1. **pre-commit run --all-files**  to build the contract test changes
2. **cfn test** to run the contract test for the a resource.
3. The contract test were run for ses-configurationSet and dummy resource.
**SES-ConfigurationSet Test Output**
```
cachedir: .pytest_cache
rootdir: /Volumes/Unix/workspace/aws-cloudformation-resource-providers-ses/aws-ses-configurationset, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_usjnl2r1.ini

handler_create.py::contract_invalid_create PASSED                                                                                                                                                    [ 20%]

============================================================================================= warnings summary =============================================================================================
/Volumes/Unix/workspace/rpdk/cloudformation-cli/env/lib/python3.7/site-packages/_pytest/logging.py:499
  /Volumes/Unix/workspace/rpdk/cloudformation-cli/env/lib/python3.7/site-packages/_pytest/logging.py:499: PytestDeprecationWarning: --no-print-logs is deprecated and scheduled for removal in pytest 6.0.
  Please use --show-capture instead.
    _issue_warning_captured(NO_PRINT_LOGS, self._config.hook, stacklevel=2)

handler_create.py::contract_create_delete
handler_create.py::contract_invalid_create
handler_create.py::contract_create_duplicate
handler_delete.py::contract_delete_read
  /Volumes/Unix/workspace/rpdk/cloudformation-cli/env/lib/python3.7/site-packages/hypothesis/strategies/_internal/strategies.py:274: NonInteractiveExampleWarning: The `.example()` method is good for exploring strategies, but should only be used interactively.  We recommend using `@given` for tests - it performs better, saves and replays failures to avoid flakiness, and reports minimal examples. (strategy: fixed_dictionaries({'Name': from_regex('^[a-zA-Z0-9_-]{1,64}\\Z')}))
    NonInteractiveExampleWarning,

-- Docs: https://docs.pytest.org/en/latest/warnings.html
========================================================================= 10 passed, 5 deselected, 5 warnings in 488.21s (0:08:08) =========================================================================
```

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -16,7 +16,11 @@ from ..boto_helpers import (
     create_sdk_session,
     get_temporary_credentials,
 )
-from ..jsonutils.pointer import fragment_decode
+from ..jsonutils.pointer import (
+    fragment_decode,
+    fragment_decode_primary_identifier,
+    fragment_encode,
+)
 from ..jsonutils.utils import traverse
 
 LOG = logging.getLogger(__name__)
@@ -98,6 +102,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._schema = None
         self._strategy = None
         self._update_strategy = None
+        self._invalid_strategy = None
         self._overrides = overrides
         self._update_schema(schema)
 
@@ -109,8 +114,9 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._schema = schema
         self._strategy = None
         self._update_strategy = None
+        self._invalid_strategy = None
 
-        self._primary_identifier_paths = self._properties_to_paths("primaryIdentifier")
+        self.primary_identifier_paths = self._properties_to_paths("primaryIdentifier")
         self.read_only_paths = self._properties_to_paths("readOnlyProperties")
         self._write_only_paths = self._properties_to_paths("writeOnlyProperties")
         self._create_only_paths = self._properties_to_paths("createOnlyProperties")
@@ -121,22 +127,14 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for identifier in additional_identifiers
         ]
 
-    def get_primary_identifier(self):
-        return self._primary_identifier_paths
-
     def has_writable_identifier(self):
-        for path in self._primary_identifier_paths:
+        for path in self.primary_identifier_paths:
             if path not in self.read_only_paths:
                 return True
         for identifier_paths in self._additional_identifiers_paths:
             for path in identifier_paths:
                 if path not in self.read_only_paths:
                     return True
-        return False
-
-    def has_readable_identifier(self):
-        if len(self.read_only_paths) > 0:
-            return True
         return False
 
     def has_read_create_property(self):
@@ -166,8 +164,8 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     @property
     def invalid_strategy(self):
         # an empty strategy (i.e. false-y) is valid
-        if self._strategy is not None:
-            return self._strategy
+        if self._invalid_strategy is not None:
+            return self._invalid_strategy
 
         # imported here to avoid hypothesis being loaded before pytest is loaded
         from .resource_generator import ResourceGenerator
@@ -175,8 +173,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         # make a copy so the original schema is never modified
         schema = json.loads(json.dumps(self._schema))
 
-        self._strategy = ResourceGenerator(schema).generate_schema_strategy(schema)
-        return self._strategy
+        self._invalid_strategy = ResourceGenerator(schema).generate_schema_strategy(
+            schema
+        )
+        return self._invalid_strategy
 
     @property
     def update_strategy(self):
@@ -198,23 +198,6 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         )
         return self._update_strategy
 
-    @property
-    def update_invalid_strategy(self):
-        # an empty strategy (i.e. false-y) is valid
-        if self._update_strategy is not None:
-            return self._update_strategy
-
-        # imported here to avoid hypothesis being loaded before pytest is loaded
-        from .resource_generator import ResourceGenerator
-
-        # make a copy so the original schema is never modified
-        schema = json.loads(json.dumps(self._schema))
-
-        self._update_strategy = ResourceGenerator(schema).generate_schema_strategy(
-            schema
-        )
-        return self._update_strategy
-
     def generate_create_example(self):
         example = self.strategy.example()
         return override_properties(example, self._overrides.get("CREATE", {}))
@@ -230,7 +213,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
 
     def generate_invalid_update_example(self, create_model):
         overrides = self._overrides.get("UPDATE", self._overrides.get("CREATE", {}))
-        example = override_properties(self.update_invalid_strategy.example(), overrides)
+        example = override_properties(self.invalid_strategy.example(), overrides)
         return {**create_model, **example}
 
     @staticmethod
@@ -299,6 +282,38 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             "Handler %r timed out." % action
         )
 
+    @staticmethod
+    def get_decoded_identifier_list(primary_identifier_path):
+        tuple_identifier = []
+        primary_identifier_list = []
+        for identifier in primary_identifier_path:
+            tuple_identifier.append(fragment_encode(identifier))
+        for identifier in tuple_identifier:
+            primary_identifier_list.append(
+                fragment_decode_primary_identifier(identifier)
+            )
+        return primary_identifier_list
+
+    @staticmethod
+    def assert_primary_identifier(primary_identifier_path, resource_model):
+        primary_identifiers_list = ResourceClient.get_decoded_identifier_list(
+            primary_identifier_path
+        )
+        for primary_identifier in primary_identifiers_list:
+            assert traverse(resource_model, primary_identifier) is not None
+
+    @staticmethod
+    def assert_primary_identifier_not_updated(
+        primary_identifier_path, created_model, updated_model
+    ):
+        primary_identifiers_list = ResourceClient.get_decoded_identifier_list(
+            primary_identifier_path
+        )
+        for primary_identifier in primary_identifiers_list:
+            assert traverse(created_model, primary_identifier) == traverse(
+                updated_model, primary_identifier
+            )
+
     def _make_payload(self, action, request):
         return {
             "credentials": self._creds.copy(),
@@ -317,15 +332,22 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         LOG.debug("Received response\n%s", payload)
         return payload
 
+    # pylint: disable=too-many-arguments
     def call_and_assert(
-        self, action, assert_status, current_model, previous_model=None, **kwargs
+        self,
+        action,
+        assert_status,
+        current_model,
+        primary_identifier_path=None,
+        previous_model=None,
+        **kwargs
     ):
         if assert_status not in [OperationStatus.SUCCESS, OperationStatus.FAILED]:
             raise ValueError("Assert status {} not supported.".format(assert_status))
         request = self.make_request(current_model, previous_model, **kwargs)
 
         start_time = time.time()
-        status, response = self.call(action, request)
+        status, response = self.call(action, request, primary_identifier_path)
         self.assert_time(start_time, time.time(), action)
         if assert_status == OperationStatus.SUCCESS:
             self.assert_success(status, response)
@@ -334,7 +356,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             error_code = self.assert_failed(status, response)
         return status, response, error_code
 
-    def call(self, action, request):
+    def call(self, action, request, primary_identifier_path=None):
         payload = self._make_payload(action, request)
         response = self._call(payload)
 
@@ -346,6 +368,9 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
 
         while status == OperationStatus.IN_PROGRESS:
             callback_delay_seconds = self.assert_in_progress(status, response)
+            self.assert_primary_identifier(
+                primary_identifier_path, response.get("resourceModel")
+            )
             sleep(callback_delay_seconds)
 
             payload["callbackContext"] = response.get("callbackContext")
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -1,4 +1,5 @@
 # pylint: disable=import-outside-toplevel
+# pylint: disable=R0904
 import json
 import logging
 import time
@@ -120,6 +121,9 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for identifier in additional_identifiers
         ]
 
+    def get_primary_identifier(self):
+        return self._primary_identifier_paths
+
     def has_writable_identifier(self):
         for path in self._primary_identifier_paths:
             if path not in self.read_only_paths:
@@ -128,6 +132,18 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for path in identifier_paths:
                 if path not in self.read_only_paths:
                     return True
+        return False
+
+    def has_readable_identifier(self):
+        if len(self.read_only_paths) > 0:
+            return True
+        return False
+
+    def has_read_create_property(self):
+        if len(self.read_only_paths) > 0:
+            return True
+        if len(self._create_only_paths) > 0:
+            return True
         return False
 
     @property
@@ -143,6 +159,21 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         schema = json.loads(json.dumps(self._schema))
 
         prune_properties(schema, self.read_only_paths)
+
+        self._strategy = ResourceGenerator(schema).generate_schema_strategy(schema)
+        return self._strategy
+
+    @property
+    def invalid_strategy(self):
+        # an empty strategy (i.e. false-y) is valid
+        if self._strategy is not None:
+            return self._strategy
+
+        # imported here to avoid hypothesis being loaded before pytest is loaded
+        from .resource_generator import ResourceGenerator
+
+        # make a copy so the original schema is never modified
+        schema = json.loads(json.dumps(self._schema))
 
         self._strategy = ResourceGenerator(schema).generate_schema_strategy(schema)
         return self._strategy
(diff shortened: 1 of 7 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -130,6 +134,18 @@ def has_writable_identifier(self):
                     return True
         return False
 
+    def has_readable_identifier(self):
+        if len(self.read_only_paths) > 0:
```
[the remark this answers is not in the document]
jotompki: since read_only_paths is public id just use that instead of having this method to check if there are paths

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -46,5 +51,46 @@ def contract_update_list_success(updated_resource, resource_client):
     # should be able to use the created model
     # to read since physical resource id is immutable
     _create_request, _created_model, _update_request, updated_model = updated_resource
+    test_primary_identifier_not_updated(
```
jotompki, line 54: ideally we could just make this its own test instead of checking for it twice. the fixture is the same for both contract_update_list_success and contract_update_read_success so you might be better off making a new test

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -46,5 +51,46 @@ def contract_update_list_success(updated_resource, resource_client):
...
+        _created_model, updated_model, resource_client.get_primary_identifier()
+    )
     models = test_list_success(resource_client, updated_model)
     assert updated_model in models
+
+
+@pytest.mark.update
+def contract_update_invalid_property(resource_client):
+    if resource_client.has_read_create_property:
```
jotompki, line 63: not sure what this means

aygold92:
>     1. Check if Primary Identifier is added after create finishes successfully

Can we add a check that it is returned in the first progress event as well?

>     2. Check  Primary Identifier should not exist after unsuccessful create finishes

This is not always the case. What if the resource is created but then fails during stabilization?

>     5. Test case where update fails with NotUpdatable error code when, we try to update an existing resource with readOnly/createOnly property

The handler should never receive readOnly property changes, this will be verified by CFN up front


Also, could you add a section to this PR description on how you tested this code?  What commands did you run? What resources did you use?

anshikg:
> >> ```
> > >1. Check if Primary Identifier is added after create finishes successfully
> > >```
> >```
> >Can we add a check that it is returned in the first progress event as well?
> >```
>
> Updating this to check the identifier
>
> > > ```
> > >2. Check  Primary Identifier should not exist after unsuccessful create finishes
> > >```
> >```
> >This is not always the case. What if the resource is created but then fails during stabilization?
> >```
>
>  I can go ahead and remove this check. But, this may lead to resource leaking if the resource is not getting cleaned up.
>
> > >```
> > >5. Test case where update fails with NotUpdatable error code when, we try to update an existing resource with readOnly/createOnly property
> > >```
> >```
> >The handler should never receive readOnly property changes, this will be verified by CFN up front
> > ```
>
> I am checking for both readOnly and createOnly property. Is it worth just checking it? Or should I remove this?
>
>> ```
> > Also, could you add a section to this PR description on how you tested this code? What commands did you run? What resources did you use?
> > ```
>
> Done. Please let me know if you need more details.
>

src/rpdk/core/contract/resource_client.py
```diff
@@ -167,15 +198,41 @@ def update_strategy(self):
         )
         return self._update_strategy
 
+    @property
+    def update_invalid_strategy(self):
```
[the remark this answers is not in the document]
anshikg: makes sense. thanks for pointing it out.

src/rpdk/core/contract/resource_client.py
```diff
@@ -130,6 +134,18 @@ def has_writable_identifier(self):
                     return True
         return False
 
+    def has_readable_identifier(self):
+        if len(self.read_only_paths) > 0:
+            return True
+        return False
+
+    def has_read_create_property(self):
```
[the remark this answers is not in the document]
anshikg: This method checks if creatOnly/readOnly property exist in the schema. I need a better name for the method. I am open to suggestions.

src/rpdk/core/contract/resource_client.py
```diff
@@ -147,6 +163,21 @@ def strategy(self):
...
+
+        # imported here to avoid hypothesis being loaded before pytest is loaded
+        from .resource_generator import ResourceGenerator
+
+        # make a copy so the original schema is never modified
+        schema = json.loads(json.dumps(self._schema))
+
+        self._strategy = ResourceGenerator(schema).generate_schema_strategy(schema)
+        return self._strategy
+
```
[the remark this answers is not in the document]
anshikg: I will go ahead and update the _strategy to _invalid _strategy
Now, that I have removed, update_invalid_strategy method we will be checking for both createOnly and readOnly properties. Therefore, invalid invalid_strategy will be accurate. Although, I will go ahead and add some comments above the method to clarify.

src/rpdk/core/contract/suite/handler_update.py
anshikg: I think it makes sense to remove the **get_primary_identifier** method. This can be replaced by a public variable **primary_identifier_paths**

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -95,3 +112,27 @@ def test_delete_failure_not_found(resource_client, current_resource_model):
...
+
+def test_primary_identifier_exists(created_model, primary_identifier):
+    for identifiers in primary_identifier:
+        for identifier in identifiers:
+            if identifier != "properties":
+                assert created_model[identifier] is not None
+                assert created_model[identifier]
+
+
+def test_primary_identifier_should_not_exist(created_model, primary_identifier):
```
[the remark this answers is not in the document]
anshikg: I am planning to remove the method **get_primary_identifier**

src/rpdk/core/contract/suite/handler_update.py
anshikg: This test updates the resource with createOnly/readOnly property and expects it to fail.

aygold92:
> > > > 2. Check  Primary Identifier should not exist after unsuccessful create finishes
> > I can go ahead and remove this check. But, this may lead to resource leaking if the resource is not getting cleaned up.

Cleaning up a failed resource is not a responsibility of the handler.  If they actually created a resource and then failed, the handler did its job if it reports the physicalId, which allows the consumer (e.g. CloudFormation) to follow up and delete it.  

Resource leaking would happen if the handler created a resource but did not report it.  I'm not sure there's a reliable way to test this since we can't force a case where the resource got created and then later failed.  

One way I can think is to call list, then call list again after resource failure to ensure that the size of the list is the same. I'm not sure how this would fit into the contract test framework though.  Also, this would only work assuming no other concurrent interactions on the account, and I'm not sure if we're making that assumption?

> > > > 5. Test case where update fails with NotUpdatable error code when, we try to update an existing resource with readOnly/createOnly property
> > I am checking for both readOnly and createOnly property. Is it worth just checking it? Or should I remove this?

Yes, you should remove it.  The passed in properties won't contain any readOnly properties (should be validated by schema validation), except it will include the primaryIdentifier, and there would again be something wrong on the caller side if these were different

src/rpdk/core/contract/resource_client.py
```diff
@@ -242,6 +282,38 @@ def assert_time(start_time, end_time, action):
             "Handler %r timed out." % action
         )
 
+    @staticmethod
+    def get_decoded_identifier_list(primary_identifier_path):
+        tuple_identifier = []
+        primary_identifier_list = []
```
jotompki, line 288: use [list comprehension](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) here. What is this method doing? encoding then decoding? seems like no-op

src/rpdk/core/contract/resource_client.py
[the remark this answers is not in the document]
aygold92: https://docs.python.org/3/library/stdtypes.html#truth-value-testing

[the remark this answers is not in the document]
anshikg: In this case, I was checking if createOnly or readOnly property exists in the resource. I could run ``` contract_update_invalid_property ``` test else I would skip it. But, as @aygold92 suggested I will remove the readOnly condition as a user will not be allowed to enter the readOnly property. Therefore, this method will be removed.

src/rpdk/core/contract/resource_client.py
```diff
@@ -242,6 +278,33 @@ def assert_time(start_time, end_time, action):
...
+    def assert_primary_identifier_not_updated(
+        primary_identifier_path, created_model, updated_model
+    ):
+        primary_identifiers_list = ResourceClient.get_decoded_identifier_list(
+            primary_identifier_path
+        )
+        for primary_identifier in primary_identifiers_list:
+            assert traverse(created_model, primary_identifier) == traverse(
+                updated_model, primary_identifier
+            )
```
anshikg, line 306: This needs to updated.

src/rpdk/core/contract/resource_client.py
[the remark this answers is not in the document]
jotompki: is this still needed?

src/rpdk/core/contract/resource_client.py
```diff
@@ -147,6 +157,23 @@ def strategy(self):
...
+        # imported here to avoid hypothesis being loaded before pytest is loaded
+        from .resource_generator import ResourceGenerator
+
+        # make a copy so the original schema is never modified
+        schema = json.loads(json.dumps(self._schema))
+
+        self._invalid_strategy = ResourceGenerator(schema).generate_schema_strategy(
+            schema
+        )
+        return self._invalid_strategy
```
jotompki, line 175: this is really only invalid if a resource has read only properties? is there another way to create an invalid strategy that doesnt depend on this being the case in the schema?

src/rpdk/core/contract/resource_client.py
```diff
@@ -242,6 +278,33 @@ def assert_time(start_time, end_time, action):
             "Handler %r timed out." % action
         )
 
+    @staticmethod
+    def get_decoded_identifier_list(identifier_path):
+        list_identifier = [
+            identifier[len(identifier) - 1] for identifier in list(identifier_path)
```
jotompki, line 284: is the casting to list necessary here? is this really decoded or just the last element in the identifier ?

if this needs to stay, please use  `identifier[-1]` instead of `identifier[len(identifier) - 1]` to achieve consistency with the codebase. Also can just return the list immediately instead of assigning then returning

src/rpdk/core/contract/resource_client.py
```diff
@@ -242,6 +278,33 @@ def assert_time(start_time, end_time, action):
...
+            assert resource_model.get(primary_identifier) is not None
+
+    @staticmethod
+    def assert_primary_identifier_not_updated(
+        primary_identifier_path, created_model, updated_model
+    ):
+        primary_identifiers_list = ResourceClient.get_decoded_identifier_list(
+            primary_identifier_path
+        )
+        for primary_identifier in primary_identifiers_list:
```
jotompki, line 303: `assert all( condition for primary_identifier in primary_identifiers_list ) `

src/rpdk/core/contract/resource_client.py
```diff
@@ -277,7 +347,7 @@ def call_and_assert(
             error_code = self.assert_failed(status, response)
         return status, response, error_code
 
-    def call(self, action, request):
+    def call(self, action, request, primary_identifier_path=None):
```
jotompki, line 350: why is this needed? isnt the primary identifier an object member? no reason to pass it in if so

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -7,11 +7,29 @@
...
+        current_resource_model,
+        resource_client.primary_identifier_paths,
+    )
+    resource_client.assert_primary_identifier(
+        resource_client.primary_identifier_paths, current_resource_model
     )
     return response
 
 
+def test_create_fail(resource_client, resource_model):
```
rjlohan, line 21: Please add comments or use a more explicit name to explain the test flow here, I don't really follow this one.

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -46,5 +50,47 @@ def contract_update_list_success(updated_resource, resource_client):
...
     _create_request, _created_model, _update_request, updated_model = updated_resource
+    resource_client.assert_primary_identifier_not_updated(
+        resource_client.primary_identifier_paths, _created_model, updated_model
+    )
     models = test_list_success(resource_client, updated_model)
     assert updated_model in models
+
+
+@pytest.mark.update
+def contract_update_invalid_property(resource_client):
```
rjlohan, line 61: Please use more explicit naming in test. It's not just an invalid property, it's a createOnlyProperties check right?

anshikg: sure, updated the name to: contract_update_create_only_property

src/rpdk/core/contract/suite/handler_commons.py
anshikg: update the method name to test_create_fail_if_read_only_property_in_input

src/rpdk/core/contract/resource_client.py
anshikg: changed the method to look similar to assert_primary_identifier

src/rpdk/core/contract/resource_client.py
anshikg: In case of update, a resource updated with createOnly property will also be invalid.

src/rpdk/core/contract/resource_client.py
anshikg: updating it to return the value instead of storing it. Plus replaced ```identifier[len(identifier) - 1]``` with ```identifier[-1]```

src/rpdk/core/contract/resource_client.py
```diff
@@ -242,6 +271,30 @@ def assert_time(start_time, end_time, action):
...
+    def get_identifier_list(identifier_path):
+        return [identifier[-1] for identifier in list(identifier_path)]
+
+    @staticmethod
+    def assert_primary_identifier(primary_identifier_path, resource_model):
+        primary_identifiers_list = ResourceClient.get_identifier_list(
+            primary_identifier_path
+        )
+        for primary_identifier in primary_identifiers_list:
+            assert resource_model.get(primary_identifier) is not None
```
jotompki, line 284: try using `assert all(resource_model.get(primary_identifier) for primary_identifier in primary_identifiers_list)`

example here:
https://github.com/aws-cloudformation/cloudformation-cli/blob/1194ce3002fee23a4a2917eff6b98c0c65dd55ec/tests/test_test.py#L235-L237

also i think `assert None` throws an error so `assert thing is not None` can be simplified

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -7,11 +7,23 @@
 
 def test_create_success(resource_client, current_resource_model):
     _status, response, _error_code = resource_client.call_and_assert(
-        Action.CREATE, OperationStatus.SUCCESS, current_resource_model
+        Action.CREATE, OperationStatus.SUCCESS, current_resource_model,
```
jotompki, line 10: dangling comma

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -22,34 +34,36 @@ def test_create_failure_if_repeat_writeable_id(resource_client, current_resource
         # resource model means that the same resource is trying to be
         # created twice.
         _status, _response, error_code = resource_client.call_and_assert(
-            Action.CREATE, OperationStatus.FAILED, current_resource_model
+            Action.CREATE, OperationStatus.FAILED, current_resource_model,
```
jotompki, line 37: ditto here

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -22,34 +34,36 @@ def test_create_failure_if_repeat_writeable_id(resource_client, current_resource
...
         ), "creating the same resource should not be possible"
-
     else:
         LOG.debug("no identifiers are writeable; skipping duplicate-CREATE-failed test")
 
 
 def test_read_success(resource_client, current_resource_model):
     _status, response, _error_code = resource_client.call_and_assert(
-        Action.READ, OperationStatus.SUCCESS, current_resource_model
+        Action.READ, OperationStatus.SUCCESS, current_resource_model,
```
jotompki, line 48: same

src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -43,6 +44,12 @@ def contract_create_delete(resource_client):
         test_delete_success(resource_client, delete_model)
 
 
+@pytest.mark.create
+def contract_invalid_create(resource_client):
+    requested_model = resource_client.generate_invalid_create_example()
+    test_create_fail_if_read_only_property_in_input(resource_client, requested_model)
```
jotompki, line 50: i know we have this "handler commons" module but honestly if this test is only used once might as well inline it here for clarity. no reason to have someone jump to another file to get the gist

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -46,5 +50,47 @@ def contract_update_list_success(updated_resource, resource_client):
...
+        resource_client.primary_identifier_paths, _created_model, updated_model
+    )
     models = test_list_success(resource_client, updated_model)
     assert updated_model in models
+
+
+@pytest.mark.update
+def contract_update_create_only_property(resource_client):
+
+    if resource_client.create_only_paths:
```
jotompki, line 63: Use pytest.skip here to denote the test didnt run: https://github.com/aws-cloudformation/cloudformation-cli/blob/f12d5b8a5d75efb71f9d259ab045020ef79fcbb0/src/rpdk/core/contract/suite/handler_delete.py#L94

src/rpdk/core/contract/resource_client.py
```diff
@@ -260,6 +308,7 @@ def _call(self, payload):
         LOG.debug("Received response\n%s", payload)
         return payload
 
+    # pylint: disable=too-many-arguments
```
jotompki, line 311: nit is this still necessary no argument change

tests/contract/test_resource_client.py
```diff
@@ -566,3 +636,78 @@ def test_assert_CUD_time_fail(resource_client, action):
...
+            "properties": {
+                "a": {"type": "number", "const": 1},
+                "b": {"type": "number", "const": 2},
+                "c": {"type": "number", "const": 3},
+            },
+            "readOnlyProperties": ["/properties/b"],
+            "createOnlyProperties": ["/properties/c"],
+            "primaryIdentifier": ["/properties/c"],
+        }
+        print(resource_client.primary_identifier_paths)
```
jotompki, line 707: get rid of this

anshikg: can I remove this in the next CR?

LaikaN57:
It seems that`contract_update_create_only_property` performs poorly when you have an overrides file that is overriding an `additionalIdentifiers` property due to https://github.com/aws-cloudformation/cloudformation-cli/blame/0cceffd8d308d6a6c6b3b419191b3ec972f2dfe2/src/rpdk/core/contract/suite/handler_update.py#L63-L67

Here is what I am experiencing:
1. `updated_resource` fixture calls create handler with `coProp/addIdProp=A` (from overrides file) and **passes** with `roProp/priIdProp=1`.
2. `updated_resource` fixture calls update handler with `roProp/priIdProp=1` and **passes**.
3. other update tests run against `roProp/priIdProp=1` and **pass**.
3. contract_update_create_only_property test calls create handler with `coProp/addIdProp=A` (from overrides file) and **fails** with `error=AlreadyExists`.

I am not sure if this is a failure of the contact test, lacking a custom fixture for this contract test, or a failure of the overrides systems we currently have. I would like to get some feedback of how we should handle this case better.

PR #449, opened 2020-06-03
Eventual Consistency Contract test

anshikg:
*Issue #, if available:*

*Description of changes:*
1) List and Read handler should not return InProgress
2) List handler should return at least an array of primary Identifiers and nextToken, run through the pagination and ensure NextToken is being returned also there is an end to the list by checking nextToken = null
3)  List → Read(primaryIdentifier from the list) → should not throw “NotFound” error code
4)  List handler should return empty in case no resource is found

*Test*
SES ConfigurationSet Resource
```
cachedir: .pytest_cache
rootdir: /Volumes/Unix/workspace/aws-cloudformation-resource-providers-ses/aws-ses-configurationset, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_48berbsw.ini
collected 17 items / 5 deselected / 12 selected

handler_create.py::contract_create_delete PASSED                                                                                                                                                     [  8%]
handler_create.py::contract_invalid_create SKIPPED                                                                                                                                                   [ 16%]
handler_create.py::contract_create_duplicate PASSED                                                                                                                                                  [ 25%]
handler_create.py::contract_create_read_success PASSED                                                                                                                                               [ 33%]
handler_create.py::contract_create_list_success PASSED                                                                                                                                               [ 41%]
handler_delete.py::contract_delete_read PASSED                                                                                                                                                       [ 50%]
handler_delete.py::contract_delete_list PASSED                                                                                                                                                       [ 58%]
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                     [ 66%]
handler_delete.py::contract_delete_create PASSED                                                                                                                                                     [ 75%]
handler_list.py::contract_list_read_success PASSED                                                                                                                                                   [ 83%]
handler_list.py::contract_list_empty PASSED                                                                                                                                                          [ 91%]
handler_list.py::contract_list_empty PASSED                                                                                                                                                           [ 91%]
```


src/rpdk/core/contract/resource_client.py
```diff
@@ -287,9 +287,25 @@ def assert_primary_identifier_not_updated(
             == traverse(updated_model, fragment_list(primary_identifier, "properties"))[
                 0
             ]
-            for primary_identifier in list(primary_identifier_path)
+            for primary_identifier in primary_identifier_path
         )
 
+    @staticmethod
+    def does_primary_identifier_exist(
```
jotompki, line 294: isnt this the same as assert_primary_identifier other than the assert? the former ensures all primary identifiers are present and asserts on that. this one just returns true false. could maybe combine and use the `assert_primary_identifier` logic here and have `assert_primary_identifier` just assert on `does_primary_identifier_exist`

src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -73,5 +73,4 @@ def contract_create_read_success(created_resource, resource_client):
 @pytest.mark.list
 def contract_create_list_success(created_resource, resource_client):
     created_model, _request = created_resource
-    models = test_list_success(resource_client, created_model)
-    assert created_model in models
+    assert test_list_success(resource_client, created_model)
```
jotompki, line 76: we already enforce the primary identifier to be in the created_model. i think checking here again is redundant, as the created_model will already have a primary identifier. if thats the case, just checking that the model is in the list of models is sufficient.

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -62,8 +62,7 @@ def contract_delete_list(resource_client, deleted_resource):
     #       remove the model from the list, however.
 
     deleted_model, _request = deleted_resource
-    list_models = test_list_success(resource_client, deleted_model)
-    assert deleted_model not in list_models
+    assert not test_list_success(resource_client, deleted_model)
```
jotompki, line 65: same here

src/rpdk/core/contract/suite/handler_list.py
```diff
@@ -0,0 +1,49 @@
...
+    finally:
+        resource_client.call_and_assert(Action.DELETE, OperationStatus.SUCCESS, model)
+
+
+@pytest.mark.list
+@pytest.mark.read
+def contract_list_read_success(created_resource, resource_client):
+    created_model, _request = created_resource
+    assert test_list_success(resource_client, created_model)
+    test_read_success(resource_client, created_model)
```
jotompki, line 34: what contract is this fulfilling? we already test that you can read a created resource in the create suite.

src/rpdk/core/contract/suite/handler_list.py
```diff
@@ -0,0 +1,49 @@
...
+
+
+@pytest.mark.list
+def contract_list_empty(resource_client):
```
jotompki, line 38: i wouldnt do this. if someone is running these tests in their account it could potentially delete resources they dont want to delete

src/rpdk/core/contract/suite/handler_create.py
jotompki: ah forgot list only returns identifiers

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -70,6 +70,18 @@ def test_list_success(resource_client, current_resource_model):
     return resource_models
 
 
+def test_list_success(resource_client, current_resource_model):
+    resource_models = get_resource_model_list(resource_client, current_resource_model)
+    result = False
+    for resource_model in resource_models:
```
jotompki, line 76: can use all() here like you did for assert functions

src/rpdk/core/contract/suite/handler_list.py
anshikg: makes sense. It might be something we can add to optional test.

src/rpdk/core/contract/suite/handler_list.py
anshikg: This ensures, if the model exists in the list. The returned model when read using the ReadHandler doe snot return "NotFound". Point 3 in the description above.

src/rpdk/core/contract/suite/handler_commons.py
anshikg: I am re using this method for both success and failure. Therefore, I don't want to assert. I just wanted to return the value. This just decreased replicating code.

tests/contract/test_resource_client.py
```diff
@@ -710,3 +710,45 @@ def test_assert_primary_identifier_not_updated_fail(resource_client):
...
+    schema = {
+        "properties": {
+            "a": {"type": "number", "const": 1},
+            "b": {"type": "number", "const": 2},
+            "c": {"type": "number", "const": 3},
+        },
+        "readOnlyProperties": ["/properties/b"],
+        "createOnlyProperties": ["/properties/c"],
+        "primaryIdentifier": ["/properties/c"],
+    }
```
PatMyron, line 725: Can we extract these schemas instead of copy-pasting the same few schemas in all these tests?

src/rpdk/core/contract/suite/handler_commons.py
jotompki: not assert, just suggesting to use 

```
all(resource_client.does_primary_identifier_exist(
            resource_client.primary_identifier_paths,
            resource_model,
            current_resource_model,
        ) for model in resource models)
```

to keep codebase consistent. should be the same thing if youre trying to test if every model has primary identifiers. 

Why are we using bitwise operators here?

tests/contract/test_resource_client.py
anshikg: I can set a global variable with this schema and re-use

src/rpdk/core/contract/resource_client.py
jotompki: this method name doesnt seem to match what is occurring,. this simply checks if the updated and created model have the same values for primary identifier, not if they exist or not. if the two models both have a value of none for the pid, it would pass

src/rpdk/core/contract/suite/handler_commons.py
PatMyron: > Why are we using bitwise operators here?

+1, was also confused by these, would prefer `and`/`or`

src/rpdk/core/contract/resource_client.py
anshikg: renaming this method to ```assert_primary_identifier_equal``` and removing the method ```assert_primary_identifier_not_updated``` to avoid replication

src/rpdk/core/contract/suite/handler_commons.py
anshikg: ok. I think I misunderstood. Will add all

tests/contract/test_resource_client.py
PatMyron: @anshikg [one more at line 409 btw](https://github.com/aws-cloudformation/cloudformation-cli/blob/4f4512064e1a0dc10ec3c27465d458f373dde50c/tests/contract/test_resource_client.py#L409)

anshikg: Updated.

src/rpdk/core/contract/suite/handler_list.py
```diff
@@ -0,0 +1,47 @@
...
+        )
+        model = response["resourceModel"]
+        yield model, request
+    finally:
+        resource_client.call_and_assert(Action.DELETE, OperationStatus.SUCCESS, model)
+
+
+@pytest.mark.list
+@pytest.mark.read
+def contract_list_read_success(created_resource, resource_client):
```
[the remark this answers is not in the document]
anshikg: in that case moving this to the create handler contract test. This can just be part of the existing test "contract_create_list_success"

src/rpdk/core/contract/suite/handler_list.py
```diff
@@ -0,0 +1,20 @@
...
+
+import pytest
+
+# WARNING: contract tests should use fully qualified imports to avoid issues
+# when being loaded by pytest
+from rpdk.core.contract.interface import Action, OperationStatus
+
+
+@pytest.mark.list
+def contract_list_empty(resource_client):
```
jotompki, line 12: can this test fail at the moment? either there arent resources and we "assert True" or we skip it? might be good to just  remove completely

PR #450, opened 2020-06-03
Shape Conformance Contract test

anshikg:
*Issue #, if available:*
Tracking this issue here: https://github.com/aws-cloudformation/cloudformation-cli/pull/451
I messed up fixing the merge conflicts in this CR.
*Description of changes:*
1)  verify "In Progress event" is being returned. If so, verify the resource created and the model returned have the same set of values
2) Verify "WriteOnlyProperties" are not being returned.

*Test*
```
cachedir: .pytest_cache
rootdir: /Volumes/Unix/workspace/aws-cloudformation-resource-providers-ses/aws-ses-configurationset, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_eje7jekx.ini

handler_create.py::contract_invalid_create SKIPPED                                                                                                                                                   [ 20%]
```

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -114,7 +114,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
 
         self.primary_identifier_paths = self._properties_to_paths("primaryIdentifier")
         self.read_only_paths = self._properties_to_paths("readOnlyProperties")
-        self._write_only_paths = self._properties_to_paths("writeOnlyProperties")
+        self.write_only_paths = self._properties_to_paths("writeOnlyProperties")
         self.create_only_paths = self._properties_to_paths("createOnlyProperties")
 
         additional_identifiers = self._schema.get("additionalIdentifiers", [])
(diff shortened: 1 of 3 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -337,9 +349,16 @@ def call(self, action, request):
 
         while status == OperationStatus.IN_PROGRESS:
             callback_delay_seconds = self.assert_in_progress(status, response)
+            _status, read_response, _error = self.call_and_assert(
+                Action.READ, OperationStatus.SUCCESS, response["resourceModel"]
+            )
+            assert read_response["resourceModel"] == response["resourceModel"]
```
ninavolkmuth, line 355: I'm worried this may fail transiently if an update/create call is stabilizing. In between when the update/create is called and the read is called, a property could stabilize though it should happen very rarely. What is the reason to check for this?

anshikg: This was done to check if the handlers are returning the whole resource model as entered by the customer or just returning the resource model which has been updated. As mentioned in the contract test, "**A Create handler SHOULD return a model containing all properties set so far and nothing more during each IN_PROGRESS event.**"

aygold92: SHOULD != MUST, and therefore we cannot enforce this in contract test

It was written as SHOULD because this may be infeasible for many handlers

aygold92:
>     1. verify "In Progress event" is being returned. If so, verify the resource created and the model returned have the same set of values

I don't think we need to or should be doing this. It's possible that handlers need to go through a phased approach for some property value (for example traffic shifting). I don't believe the contract says anything about this, but if it does we need to update the contract

src/rpdk/core/contract/resource_client.py
anshikg: makes sense. I will go ahead and remove this.

PR #451, opened 2020-06-05
Shape Conformance Contract test

anshikg:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/pull/450/

*Description of changes:*
~~1)  verify "In Progress event" is being returned. If so, verify the resource created and the model returned have the same set of values~~
2) Verify "WriteOnlyProperties" are not being returned.

*Test*
```
cachedir: .pytest_cache
rootdir: /Volumes/Unix/workspace/aws-cloudformation-resource-providers-ses/aws-ses-configurationset, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_eje7jekx.ini

handler_create.py::contract_invalid_create SKIPPED                                                                                                                                                   [ 20%]
```

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -133,6 +133,15 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                     return True
         return False
 
+    def assert_write_only_property_does_not_exist(self, resource_model):
+        if self.write_only_paths:
+            assert not any(
+                traverse(
+                    resource_model, fragment_list(write_only_property, "properties")
+                )
+                for write_only_property in self.write_only_paths
+            )
+
     @property
     def strategy(self):
         # an empty strategy (i.e. false-y) is valid
@@ -290,18 +299,6 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for primary_identifier in primary_identifier_path
         )
 
-    @staticmethod
-    def assert_write_only_property_does_not_exist(
-        resource_model, write_only_properties
-    ):
-        if write_only_properties:
-            assert not all(
-                traverse(
-                    resource_model, fragment_list(write_only_property, "properties")
-                )
-                for write_only_property in write_only_properties
-            )
-
     def _make_payload(self, action, request):
         return {
             "credentials": self._creds.copy(),
@@ -353,9 +350,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             self.assert_primary_identifier(
                 self.primary_identifier_paths, response.get("resourceModel")
             )
-            self.assert_write_only_property_does_not_exist(
-                response["resourceModel"], self.write_only_paths
-            )
+            self.assert_write_only_property_does_not_exist(response["resourceModel"])
             sleep(callback_delay_seconds)
 
             payload["callbackContext"] = response.get("callbackContext")
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -138,7 +138,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             assert not any(
                 traverse(
                     resource_model, fragment_list(write_only_property, "properties")
-                )
+                )[0]
                 for write_only_property in self.write_only_paths
             )
 
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -114,7 +114,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
 
         self.primary_identifier_paths = self._properties_to_paths("primaryIdentifier")
         self.read_only_paths = self._properties_to_paths("readOnlyProperties")
-        self._write_only_paths = self._properties_to_paths("writeOnlyProperties")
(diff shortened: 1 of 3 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -290,6 +290,18 @@ def is_primary_identifier_equal(
             for primary_identifier in primary_identifier_path
         )
 
+    @staticmethod
```
jotompki, line 293: make this an object method and use self.write_only_properties instead in the inside. it looks like everywhere it is used it uses that

src/rpdk/core/contract/resource_client.py
```diff
@@ -290,6 +290,18 @@ def is_primary_identifier_equal(
...
+    def assert_write_only_property_does_not_exist(
+        resource_model, write_only_properties
+    ):
+        if write_only_properties:
+            assert not all(
```
jotompki, line 298: the way all() works is all evaluations have to satisfy. that means one writeOnlyProperty exists but the rest don't this would evaluate to false and pass the assertion. I think you want to use any() here and if any properties are found fail.

tests/contract/test_resource_client.py
```diff
@@ -670,3 +670,62 @@ def test_is_primary_identifier_equal_fail(resource_client):
...
+            "properties": {
+                "a": {"type": "number", "const": 1},
+                "b": {"type": "number", "const": 2},
+                "c": {"type": "number", "const": 3},
+                "d": {"type": "number", "const": 4},
+            },
+            "readOnlyProperties": ["/properties/b"],
+            "createOnlyProperties": ["/properties/c"],
+            "primaryIdentifier": ["/properties/c"],
+            "writeOnlyProperties": ["/properties/d"],
```
jotompki, line 699: add a test with multiple writeOnlyProperties

anshikg: do we need a separate test for this? Or just adding it this test will work?

jotompki: i would just parameterize this test with two schemas

src/rpdk/core/contract/resource_client.py
```diff
@@ -133,6 +133,15 @@ def has_writable_identifier(self):
                     return True
         return False
 
+    def assert_write_only_property_does_not_exist(self, resource_model):
+        if self.write_only_paths:
+            assert not any(
+                traverse(
```
jotompki, line 139: if this attempts to fetch a key that does exist does it continue or throw an exception?

anshikg: This test was not set up right. I have fixed it now to pass None value to write_only property

jotompki: to clarify if a writeOnlyProperty doesn't exist what do we return? it looks like traverse will throw a KeyError:

https://github.com/aws-cloudformation/cloudformation-cli/blob/71e17edee7193fc4bbf17940ed4d4905ca81df8a/src/rpdk/core/jsonutils/utils.py#L82-L85

anshikg: I am checking ```if self.write_only_paths:``` at line 137. We should not get the Key Error.

tests/contract/test_resource_client.py
```diff
@@ -670,3 +685,28 @@ def test_is_primary_identifier_equal_fail(resource_client):
...
+        "b": {"type": "number", "const": 2},
+        "c": {"type": "number", "const": 3},
+    }
+    resource_client._update_schema(schema)
+    resource_client.assert_write_only_property_does_not_exist(schema)
+
+
+@pytest.mark.parametrize("schema", [SCHEMA, SCHEMA_WITH_MULTIPLE_WRITE_PROPERTIES])
+def test_assert_write_only_property_does_not_exist_success(resource_client, schema):
+    created_resource = {"a": None, "b": 2, "c": 3, "d": None}
```
jotompki, line 702: what happens if a and b arent specified at all? it should handle it in this case

tests/contract/test_resource_client.py
```diff
@@ -670,3 +685,28 @@ def test_is_primary_identifier_equal_fail(resource_client):
         {"a": 1, "b": 2, "c": 3},
         {"a": 1, "b": 2, "c": 4},
     )
+
+
+def test_assert_write_only_property_does_not_exist(resource_client):
```
jotompki, line 690: what is this testing? a schema with no write only properties? and why are we passing the schema in as a resource model

anshikg: checking for KeyError Exception now.

tests/contract/test_resource_client.py
anshikg: checking for that now.

src/rpdk/core/contract/resource_client.py
```diff
@@ -214,6 +212,15 @@ def generate_invalid_update_example(self, create_model):
         example = override_properties(self.invalid_strategy.example(), overrides)
         return {**create_model, **example}
 
+    @staticmethod
```
anshikg, line 215: set this method to static to avoid: ``` src/rpdk/core/contract/resource_client.py:143:4: R0201: Method could be a function (no-self-use) ```

jotompki: can we just move to jsonutils module?

Issue #242, opened 2019-05-29
cfn-cli codegen oneOf union types is not handled


dchakrav-github:
### Updated Issue Description 
The issue here is that we initially disallowed having a property be declared as multiple types.  For example `{"type": ["string", "array"]}` was deemed to be invalid. This includes use of *combiners* (`oneOf`, `anyOf`, `allOf`) to declare multiple types as well.

We then loosened that restriction on the `type` property in #454, but we did not move the restriction for `$ref`, causing an inconsistency there

---
### Original Issue Description
When modeling AWS::SQS::Queue, we either have a FIFOQueue or StandardQueue for the type. This can be modeled via oneOf these 2 types. The property sets and property based constraints are different in these 2 sets. E.g. QueueName for FIFOQueue has to end with .fifo$ vs standard one doesn't have this constraint. oneOf semantics of JSONSchema is not handled.

[aws-sqs-queue.json.txt](https://github.com/aws-cloudformation/aws-cloudformation-rpdk/files/3234391/aws-sqs-queue.json.txt)

rpdk.core.jsonutils.utils.ConstraintError: Object at path '#' declared multiple values for '$ref': found '('definitions', 'StandardQueue')' and '('definitions', 'FifoQueue')'

[rpdk.log](https://github.com/aws-cloudformation/aws-cloudformation-rpdk/files/3234374/rpdk.log)

ammokhov:
tracking #333

PR #454, opened 2020-06-12
added generic  type for multiple type support

ammokhov:
*Issue #, if available:*

*Description of changes:*

added a generic type for multi type support. 

Example:

```json
"properties": {
    "property1": {
        "type": ["string", "object"]
    }
}
```
generated `ResourceModel.java`:
```java
@JsonAutoDetect(fieldVisibility = Visibility.ANY, getterVisibility = Visibility.NONE, setterVisibility = Visibility.NONE)
public class ResourceModel {
    @JsonProperty("property1")
    private Object property1;
    ...
```

test cases:
* 1  [CREATE] with escaped string `property1` -> [UPDATE] with modified escaped string `property1`
* 2 [CREATE] with escaped string `property1` -> [UPDATE] with json object `property1` 
* 3 [CREATE] with json object `property1` -> [UPDATE] with modified json `property1`
* 4 [CREATE] with list input for `property1` -> Failed with error `Model validation failed (#/property1: #: no subschema matched out of the total 2 subschemas) #/property1: expected type: JSONObject, found: JSONArray (#/property1) #/property1: expected type: String, found: JSONArray (#/property1)` 

### DIFF
--- a/src/rpdk/core/jsonutils/resolver.py
+++ b/src/rpdk/core/jsonutils/resolver.py
@@ -15,6 +15,7 @@ class ContainerType(Enum):
     DICT = auto()
     LIST = auto()
     SET = auto()
+    MULTIPLE = auto()
 
 
 class ResolvedType:
@@ -27,14 +28,6 @@ class ResolvedType:
 
     def __eq__(self, other):
         return self.container == other.container and self.type == other.type
-
-
-def convert_list_to_object(prop_type):
-    if isinstance(
-        prop_type, list
-    ):  # generate a generic type Object which will be casted on the client side
-        return MULTIPLE
-    return prop_type
 
 
 class ModelResolver:
@@ -113,7 +106,11 @@ class ModelResolver:
             return self._get_array_lang_type(property_schema)
         if schema_type == "object":
             return self._get_object_lang_type(property_schema)
-        return self._get_primitive_lang_type(convert_list_to_object(schema_type))
+        if isinstance(
+            schema_type, list
+        ):  # generate a generic type Object which will be casted on the client side
+            return self._get_multiple_lang_type()
+        return self._get_primitive_lang_type(schema_type)
 
     @staticmethod
     def _get_array_container_type(property_schema):
@@ -124,6 +121,10 @@ class ModelResolver:
         if insertion_order or not unique_items:
             return ContainerType.LIST
         return ContainerType.SET
+
+    @staticmethod
+    def _get_multiple_lang_type():
+        return ResolvedType(ContainerType.MULTIPLE, MULTIPLE)
 
     @staticmethod
     def _get_primitive_lang_type(schema_type):
--- a/src/rpdk/core/jsonutils/resolver.py
+++ b/src/rpdk/core/jsonutils/resolver.py
@@ -6,6 +6,7 @@ from .flattener import JsonSchemaFlattener
 from .utils import BASE, fragment_encode
 
 UNDEFINED = "undefined"
+MULTIPLE = "multiple"
 
 
 class ContainerType(Enum):
@@ -26,6 +27,14 @@ class ResolvedType:
 
     def __eq__(self, other):
         return self.container == other.container and self.type == other.type
+
+
+def convert_list_to_object(prop_type):
+    if isinstance(
+        prop_type, list
+    ):  # generate a generic type Object which will be casted on the client side
+        return MULTIPLE
+    return prop_type
 
 
 class ModelResolver:
@@ -104,7 +113,7 @@ class ModelResolver:
             return self._get_array_lang_type(property_schema)
         if schema_type == "object":
             return self._get_object_lang_type(property_schema)
-        return self._get_primitive_lang_type(schema_type)
+        return self._get_primitive_lang_type(convert_list_to_object(schema_type))
 
     @staticmethod
     def _get_array_container_type(property_schema):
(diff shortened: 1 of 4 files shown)


rjlohan:
I think we actually want to support a fix for this issue: https://github.com/aws-cloudformation/cloudformation-cli/issues/242

`oneOf` is the correct schema structure for a union type.

src/rpdk/core/jsonutils/resolver.py
```diff
@@ -104,7 +113,7 @@ def _schema_to_lang_type(self, property_schema):
             return self._get_array_lang_type(property_schema)
         if schema_type == "object":
             return self._get_object_lang_type(property_schema)
-        return self._get_primitive_lang_type(schema_type)
+        return self._get_primitive_lang_type(convert_list_to_object(schema_type))
```
wbingli, line 116: It's more like hacking into the primitive type, but actually we should have a different type. e.g. 

```
if isinstance(schema_type, list)
       return self._get_multiple_lang_type(property_schema)

def _get_multiple_lang_type(self, property_schema):
 return ResolvedType(ContainerType.MULTIPLE, property_schema)

```

ammokhov: done

src/rpdk/core/project.py
```diff
@@ -434,7 +438,10 @@ def _set_docs_properties(
 
         prop_type = prop.get("type", "object")
 
-        if prop_type in BASIC_TYPE_MAPPINGS:
+        if isinstance(prop_type, list):
+            mapped = MULTIPLE_TYPE_MAPPINGS[MULTIPLE]
+            prop["jsontype"] = prop["yamltype"] = prop["longformtype"] = mapped
```
wbingli, line 443: Here is to generate the doc.  We actually could show all the types in the doc, it's different as to generate the Java model.  If we only generate 'Object' for now,  not need the extra `MULTIPLE_TYPE_MAPPINGS[MULTIPLE]`

src/rpdk/core/jsonutils/resolver.py
```diff
@@ -116,6 +122,10 @@ def _get_array_container_type(property_schema):
             return ContainerType.LIST
         return ContainerType.SET
 
+    @staticmethod
+    def _get_multiple_lang_type():
+        return ResolvedType(ContainerType.MULTIPLE, MULTIPLE)
```
wbingli, line 127: It's better to pass the multiple types information, `return ResolvedType(ContainerType.MULTIPLE, schema_type)`

src/rpdk/core/jsonutils/resolver.py
```diff
@@ -109,7 +109,7 @@ def _schema_to_lang_type(self, property_schema):
         if isinstance(
             schema_type, list
         ):  # generate a generic type Object which will be casted on the client side
-            return self._get_multiple_lang_type()
+            return self._get_multiple_lang_type(MULTIPLE)
```
wbingli, line 112: Should it be `self._get_multiple_lang_type(schema_type)`?

aygold92:
what is the use case for this? As far as I can tell, an indvidual property should never have two different possible types. This would be horrible API design and also does not make sense to map to a language model (which is why you need to cast it here)

benkehoe:
There are plenty of places that you might want multiple types. A common one is to allow a nominally list-of-strings type to be given as string directly when there's only one element. The AWS IAM policy language supports this, and is embedded in a number of CloudFormation resources.

However, I agree with @rjlohan that instead of extending the JSON Schema spec, `oneOf` is the right way to support this in the schema.

aygold92:
> There are plenty of places that you might want multiple types. A common one is to allow a nominally list-of-strings type to be given as string directly when there's only one element. The AWS IAM policy language supports this, and is embedded in a number of CloudFormation resources.

true, if we want to fully model an IAM policy, we would have to either allow this or allow multiple types declared using oneOf.  It probably would make the most to allow either/or.  This PR needs to edit this validation and merge the types into an array here https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/jsonutils/utils.py#L154
It wouldn't make sense to allow multiple types to be declared in one way but not using oneOf

aygold92:
> This PR needs to edit this validation and merge the types into an array here https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/jsonutils/utils.py#L154
> It wouldn't make sense to allow multiple types to be declared in one way but not using oneOf

track that in https://github.com/aws-cloudformation/cloudformation-cli/issues/242

srujithpoondla03:
> > There are plenty of places that you might want multiple types. A common one is to allow a nominally list-of-strings type to be given as string directly when there's only one element. The AWS IAM policy language supports this, and is embedded in a number of CloudFormation resources.
> 
> true, if we want to fully model an IAM policy, we would have to either allow this or allow multiple types declared using oneOf. It probably would make the most to allow either/or. This PR needs to edit this validation and merge the types into an array here https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/jsonutils/utils.py#L154
> It wouldn't make sense to allow multiple types to be declared in one way but not using oneOf

As Json schema already supports the functionality of using oneOf to specify multiple types, I think we should utilize the same and that makes our life easier in terms of validating the schema and template in later stages. If we support this new generic type it opens multiple options in authoring the schema where schema authors could confuse with the original oneOf usage.

Issue #445, opened 2020-05-22
[contract tests] create tests do not pass updated resource model


ammokhov:
```
handler_create.py::contract_create_delete
handler_create.py::contract_create_duplicate
handler_create.py::contract_create_read_success
handler_create.py::contract_create_list_success
```
following tests do not pass updated resource model on handler reinvocation, which causes issues for the resources that have primary identifier as `readOnly` property

anshikg:
can you please provide with the stack trace or the error message for each of the resources. Also, could you mention the resource you are working on.

ammokhov:
this is relevant to any resource where primary id is read only and create handler requires stabilization. Once resource is created the updated resource model is returned by the handler but not received one reinvocation, hence the test just runs forever (cant get passed first step since it receives model without primary id) and creates a bunch of duplicate resources.

ammokhov:
sample trace of four consecutive invocations of the handler by the test `handler_create.py::contract_create_delete`

```
[CREATE] invoking handler…
ResourceModel(primaryId=null)
[CREATE] handler invokedEND

[CREATE] invoking handler…
ResourceModel(primaryId=null)
[CREATE] handler invokedEND

[CREATE] invoking handler…
ResourceModel(primaryId=null)
[CREATE] handler invokedEND

[CREATE] invoking handler…
ResourceModel(primaryId=null)
[CREATE] handler invokedEND
```

anshikg:
Hey @ammokhov could you please provide a stack trace when you run the contract test. If you are trying to create a resource whose primary Identifier is readOnly I would expect the input to the identifier to be null. It would be helpful to get the input and output to the contract test to see what is exactly happening. You could run a specific contract test using the following command:
```
cfn test -- -k contract_create_delete
```
The above should give you a detailed stack on your terminal.

PR #467, opened 2020-06-23
Update payload once InProgress is returned

anshikg:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/445

*Description of changes:* Resources which have in progress being returned with some update to the resource model, fail the contract test because the payload for the consecutive request was not getting updated.

*Test:*
* Ran: cfn test -- -k contract_delete_delete for AWS::KMS::Key resource


src/rpdk/core/contract/resource_client.py
```diff
@@ -370,7 +370,10 @@ def call(self, action, request):
             self.assert_write_only_property_does_not_exist(response["resourceModel"])
             sleep(callback_delay_seconds)
 
-            payload["callbackContext"] = response.get("callbackContext")
+            request["desiredResourceState"] = response.get("resourceModel")
+            payload = self._make_payload(
+                action, request, response.get("callbackContext")
```
jotompki, line 375: can we just make this:
```
payload["callbackContext"] = response.get("callbackContext")
request["desiredResourceState"] = response.get("resourceModel")
```
? what benefit do we get from making the payload again? i guess changing the request token?
```

anshikg: > can we just make this:
> 
> ```
> payload["callbackContext"] = response.get("callbackContext")
> request["desiredResourceState"] = response.get("resourceModel")
> ```
> 
> ? what benefit do we get from making the payload again? i guess changing the request token?
> 
> ```
> 
> ```

you are right, it makes sense to just update

PR #476, opened 2020-07-06
Fix Update Tests Ordering and Fixture Usage

LaikaN57:
*Issue #, if available:*
Original issue PR: https://github.com/aws-cloudformation/cloudformation-cli/pull/468

*Description of changes:*
Use the `update_resource` test fixture in the `contract_update_create_only_property` test. This way we won't run into the issue of attempting to create the same resource twice due to overrides. However this same approach cannot be done to `contract_update_non_existent_resource` since it is not expecting the resources to exist at all. In this case, we simply change the test execution order to run this test before the test fixture is run. I have added a large note on the function as well.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_update.py
+++ b/src/rpdk/core/contract/suite/handler_update.py
@@ -30,6 +30,21 @@ def updated_resource(resource_client):
         resource_client.call_and_assert(Action.DELETE, OperationStatus.SUCCESS, model)
 
 
+# This test must be run before any tests that use the updated_resource fixture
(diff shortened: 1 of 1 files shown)


LaikaN57:
@jotompki Here is an updated attempt at https://github.com/aws-cloudformation/cloudformation-cli/pull/468

anshikg:
I think this issue should be resolved now: https://github.com/aws-cloudformation/cloudformation-cli/pull/499

LaikaN57:
Closing as this looks to be fixed in #499. I will open a new issue and link if I see continued similar issues.

Issue #361, opened 2019-12-02
Contract tests: catch primary identifier change during update


tobywf:
I don't think primary identifiers/physical resource IDs can be changed during UPDATE, so we should test for this. Maybe additional validation can be added to the resource schema validation to ensure the primary identifier is marked create only.

rjlohan:
Definitely `primaryIdentifier` should require the same setting in `createOnlyProperties` as it can *not* be changed on `Update`.

`readonlyProperties` is an additional semantic setting which would be present if the `primaryIdentifier` is *not* user-defined.

PR #482, opened 2020-07-08
added primary id validation

ammokhov:
*Issue #, if available:* #361 

*Description of changes:*

we do not allow mutable updates on primary identifier, instead it could be updated immutably, which causes the replacement.
However, primary identifier could be specified in two ways: 
* `readOnly` - auto generated by handler or api
* `createOnly` - specified by the user and cant be further mutably updated

Since, we would want to enforce the schema modeling it makes sense to add this to schema validation to provide definition correctness. 


rjlohan:
Should this validation be added to https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/tree/master/src as well/instead?

jotompki:
> Should this validation be added to https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/tree/master/src as well/instead?

Agree with this we should just schematize this

src/rpdk/core/data_loaders.py
```diff
@@ -106,6 +113,16 @@ def load_resource_spec(resource_spec_file):  # noqa: C901
         LOG.debug("Resource spec validation failed", exc_info=True)
         raise SpecValidationError(str(e)) from e
 
+    in_readonly = _is_in(resource_spec, "readOnlyProperties")
+    in_createonly = _is_in(resource_spec, "createOnlyProperties")
+
+    primary_id = resource_spec["primaryIdentifier"]
+    if not in_readonly(primary_id) and not in_createonly(primary_id):
```
jotompki, line 120: can simplify a bit to make condition `primaryIdentifier - (createOnlyProperties | readOnlyProperties)`  where they are all sets

ammokhov:
> 
> Should this validation be added to https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/tree/master/src as well/instead? 

> Agree with this we should just schematize this

as discussed offline, we cant really schematize the condition, so this change should go in both repos. 
However, if we start raising this exception it might break the registration, so alternatively we can log a warning to highlight the issue

src/rpdk/core/data_loaders.py
ammokhov: used closure to reuse the subset operation but we might use a disjoint of `createOnly | readOnly`

ammokhov:
> why do we want to change this to warning?

we want to inform a customer that id should be either createonly or readonly but if we start throwing an exception it will block existing resources from registration. warning still informs but make it backward compatible

src/rpdk/core/data_loaders.py
PatMyron: [not sure this is handling composite primaryIdentifiers correctly](https://sage.amazon.com/questions/1025861)

tests/data/schema/valid/invalid_primary_id.json
```diff
@@ -0,0 +1,13 @@
+{
+    "typeName": "AWS::Valid::TypeName",
```
PatMyron, line 2: Was this schema to test https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/issues/98?
`typeName` should be renamed and should switch directories if so

PR #485, opened 2020-07-09
Fixed contract test check on writeOnlyProperties redaction

rjlohan:
*Issue #, if available:* N/A

*Description of changes:* Contract test was asserting that writeOnlyProperties are not returned on an IN_PROGRESS result which is problematic, as many resources need writeOnlyProperties to exist on the resource through re-invocation or stabilization checks. The redaction assert is only needed on the final result from a SUCCESS or FAILED response.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -367,7 +367,6 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             self.assert_primary_identifier(
                 self.primary_identifier_paths, response.get("resourceModel")
             )
-            self.assert_write_only_property_does_not_exist(response["resourceModel"])
             sleep(callback_delay_seconds)
 
             request["desiredResourceState"] = response.get("resourceModel")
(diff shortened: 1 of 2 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -377,6 +376,10 @@ def call(self, action, request):
             response = self._call(payload)
             status = OperationStatus[response["status"]]
 
+        # ensure writeOnlyProperties are not returned on final responses
+        if "resourceModel" in response.keys():
```
anshikg, line 380: Do we need this check? Would there be a time when the response will not have "resourceModel"?

tests/contract/test_resource_client.py
```diff
@@ -476,6 +476,23 @@ def test_call_async(resource_client, action):
     assert response == {"status": OperationStatus.SUCCESS.value}
 
 
+@pytest.mark.parametrize("action", [Action.CREATE, Action.UPDATE, Action.DELETE])
```
anshikg, line 479: Is it worth adding a test for In_Progress with writeOnlyProperties in the response?

src/rpdk/core/contract/resource_client.py
rjlohan: Some of the tests don't set it up, and it's possible that a model is not returned, e.g; on a `Delete` action being successful?

tests/contract/test_resource_client.py
rjlohan: Good idea, will do.

Issue #481, opened 2020-07-07
Contract test timeout assert fails when the tests are run locally


anshikg:
Local Contract test fail to assert the handler timeout. This happens even though the handler timeout meets the handler contract. There needs to be a better mechanism to overcome the timeout other than just overriding the existing code by getting a local copy of cloudformation-cli.
Following line asserts when running the contract test: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L295

PR #486, opened 2020-07-10
Add Enforce timeout tag to increase timeout for handler for local tes…

anshikg:
…ting

*Issue #, if available:*https://github.com/aws-cloudformation/cloudformation-cli/issues/481

*Description of changes:* Adding an additional tag to allow customers to add timeout in order to increase the timeout asserts for handlers when running contract test locally. By default the timeout for CDU handlers is 60 seconds and for RL handler is 30 seconds. 

for the following command: cfn test --enforce-timeout 120 -- -k contract_create_delete
the timeout for RL handler is 120 seconds and CDU handler is 240 seconds.

*Test:*
* built the code locally: pre-commit run --all-files
* ran contract test for : aws-ses-configurationset resource 
```
 cfn test --enforce-timeout 120 -- -k contract_create_delete 
```

```
(env) 88e9fe53273b:aws-ses-configurationset anshikg$ cfn test --enforce-timeout 120 -- -k contract_create_delete
============================================================================================================================= test session starts ==============================================================================================================================
platform darwin -- Python 3.7.2, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /Volumes/Unix/workspace/rpdk/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Volumes/Unix/workspace/rpdk/aws-cloudformation-resource-providers-ses/aws-ses-configurationset/.hypothesis/examples')
rootdir: /Volumes/Unix/workspace/rpdk/aws-cloudformation-resource-providers-ses/aws-ses-configurationset, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_quccp96l.ini
plugins: hypothesis-5.19.0, cov-2.10.0, localserver-0.5.0, random-order-1.0.4
collected 15 items / 14 deselected / 1 selected

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                                                         [100%]

=========================================================================================================== 1 passed, 14 deselected, 2 warnings in 90.56s (0:01:30) ============================================================================================================
```

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -78,6 +78,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         overrides,
         inputs=None,
         role_arn=None,
+        timeout_in_seconds=None,
     ):  # pylint: disable=too-many-arguments
         self._schema = schema
         self._session = create_sdk_session(region)
@@ -109,6 +110,9 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._overrides = overrides
         self._update_schema(schema)
         self._inputs = inputs
+        self._timeout_in_seconds = (
+            30 if timeout_in_seconds is None else int(timeout_in_seconds)
+        )
 
     def _properties_to_paths(self, key):
         return {fragment_decode(prop, prefix="") for prop in self._schema.get(key, [])}
@@ -290,9 +294,12 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     def generate_token():
         return str(uuid4())
 
-    @staticmethod
-    def assert_time(start_time, end_time, action):
-        timeout_in_seconds = 30 if action in (Action.READ, Action.LIST) else 60
+    def assert_time(self, start_time, end_time, action):
+        timeout_in_seconds = (
+            self._timeout_in_seconds
+            if action in (Action.READ, Action.LIST)
+            else self._timeout_in_seconds * 2
+        )
         assert end_time - start_time <= timeout_in_seconds, (
             "Handler %r timed out." % action
         )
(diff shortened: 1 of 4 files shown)


src/rpdk/core/test.py
```diff
@@ -201,6 +201,7 @@ def invoke_test(args, project, overrides, inputs):
             overrides,
             inputs,
             args.role_arn,
+            args.enforce_timeout,
```
rjlohan, line 204: Does this flag require an int value as argument? Can it default to 60seconds if you supply only the flag.

Also, can we include an update to docs/README with this change?

anshikg: This takes in a string and converts it to int. Right now if no flag is provided it defaults to 60/30 seconds depending on the handler. I can add the feature to default to 60/30 with no flag value. 

I will add readme for this.

rjlohan: Oh I thought this PR was to make 'no timeout enforced' the default - because it fails all the time on SAM Local?

anshikg: it just allows customers to increase the timeout. If no --enforce-timeout arg is mentioned it assert on 60 seconds for CUD and 30 seconds fo RL handlers.

src/rpdk/core/contract/resource_client.py
```diff
@@ -109,6 +110,9 @@ def __init__(
         self._overrides = overrides
         self._update_schema(schema)
         self._inputs = inputs
+        self._timeout_in_seconds = (
+            30 if timeout_in_seconds is None else int(timeout_in_seconds)
```
ammokhov, line 114: maybe instead of `timeout_in_seconds=None` we should have `timeout_in_seconds=30`?
then code could be just `self._timeout_in_seconds = int(timeout_in_seconds)`

anshikg: I wanted to set the timeout_in_seconds parameter to None so that we can keep this at the very end.

Issue #494, opened 2020-07-16
Add Contract test for Read


anshikg:
Read handler for a resource which does not exists should fail with "NotFound" error code.

PR #497, opened 2020-07-17
Add read without create test, add error code to update handler

anshikg:
*Issue #, if available:*https://github.com/aws-cloudformation/cloudformation-cli/issues/494

*Description of changes:* 
* Adding a test case to check if read fails when no resource was created.
* **contract_update_create_only_property**  fails with Not_Found error code in resources where createOnly property is also the primary Identifier. Therefore asserting for either error codes.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_read.py
+++ b/src/rpdk/core/contract/suite/handler_read.py
@@ -1,0 +1,14 @@
+# fixture and parameter have the same name
+# pylint: disable=redefined-outer-name
+
+import pytest
+
+# WARNING: contract tests should use fully qualified imports to avoid issues
+# when being loaded by pytest
+from rpdk.core.contract.suite.handler_commons import test_read_failure_not_found
+
+
+@pytest.mark.read
+def contract_read_without_create(resource_client):
+    model = resource_client.generate_create_example()
+    test_read_failure_not_found(resource_client, model)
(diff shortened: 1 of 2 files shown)


src/rpdk/core/contract/suite/handler_read.py
```diff
@@ -0,0 +1,14 @@
...
+
+import pytest
+
+# WARNING: contract tests should use fully qualified imports to avoid issues
+# when being loaded by pytest
+from rpdk.core.contract.suite.handler_commons import test_read_failure_not_found
+
+
+@pytest.mark.read
+def contract_read_without_create(resource_client):
```
ammokhov, line 12: you mean if resource been deleted? we cant read if it was not created

ammokhov:
is it a duplicate of this #496 ?

anshikg:
> is it a duplicate of this #496 ?

yes, the CR was not set up right. I closed it.

src/rpdk/core/contract/suite/handler_read.py
anshikg: Lets say we try to query an invalid primary identifier for a resource. It should fail with "Not Found"

ammokhov: is that a possible scenario?

PR #498, opened 2020-07-17
asserts that primary id has not got changed throughout update

ammokhov:
*Issue #, if available:*

*Description of changes:*

this is also a part of this issue #361
We need to assert that primary id did not get changed withing handler code


anshikg:
> I think we already make this check in update: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update.py#L39

Discussed offline with Anton, it seems we want to assert the primary identifier has not changed after the update is complete.

Issue #488, opened 2020-07-13
contract_update_create_only_property's call to Create Handler has AlreadyExists Error


cortez7:
When I run contract tests, on the 2nd to last one, handler_update.py::contract_update_create_only_property, I get an AlreadyExists failure on the initial creation at the start of the test. Sometimes I get a timeout error in one of the Read or List Handlers, but even when I don't get a timeout error, I still get an AlreadyExists error on contract_update_create_only_property.

[Contract Test Results](https://paste.amazon.com/show/cortpage/1594667429)

When I run this test by itself, it passes.

I enter a sample LoadBalancerName (which is the primary id) in the overrides.json. I tried removing this override, but the contract tests just generate a name of “0” and the same thing happens. I tried entering null to force a random name generation, but when null is passed to the Delete handler, it doesn’t know what resource to delete.

The contract tests didn’t delete the resource beforehand so the AlreadyExists failure is expected. Looking at rpdk.log, the order of calls leading up to the error is Delete, Create, Update, Read, List, Create with AlreadyExists error.

ammokhov:
regarding RL timout - should this solve it (#486)?

cortez7:
Yes, I think that will.

anshikg:
Waiting on @cortez7 for stack trace for the failing contract test

cortez7:
I re-ran tests with a local copy of https://github.com/aws-cloudformation/cloudformation-cli. I happened to not get errors due to timeouts this time, but the last 2 tests still fail.
[Contract Test Results](https://paste.amazon.com/show/cortpage/1594772142)

PR #499, opened 2020-07-18
Fix the fixture scope for update handler

anshikg:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/488

*Description of changes:*
fixtures for update handler test need to run for functions which need the create and update common model/request

* Test:*
* Ran test for elasticloadbalancing::loadbalancer
```
==================================================================================================================== test session starts =====================================================================================================================
platform darwin -- Python 3.8.4, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /Users/anshikg/workspace/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-elasticloadbalancing/aws-elasticloadbalancing-loadbalancer/.hypothesis/examples')
rootdir: /Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-elasticloadbalancing/aws-elasticloadbalancing-loadbalancer, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_ugnqj5e2.ini
collected 16 items                                                                                                                                                                                                                                           

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                                       [  6%]
handler_create.py::contract_invalid_create PASSED                                                                                                                                                                                                      [ 12%]
handler_create.py::contract_create_duplicate PASSED                                                                                                                                                                                                    [ 18%]
handler_create.py::contract_create_read_success PASSED                                                                                                                                                                                                 [ 25%]
handler_create.py::contract_create_list_success PASSED                                                                                                                                                                                                 [ 31%]
handler_delete.py::contract_delete_read PASSED                                                                                                                                                                                                         [ 37%]
handler_delete.py::contract_delete_list PASSED                                                                                                                                                                                                         [ 43%]
handler_delete.py::contract_delete_update PASSED                                                                                                                                                                                                       [ 50%]
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                                                                       [ 56%]
handler_delete.py::contract_delete_create PASSED                                                                                                                                                                                                       [ 62%]
handler_misc.py::contract_check_asserts_work PASSED                                                                                                                                                                                                    [ 68%]
handler_read.py::contract_read_without_create PASSED                                                                                                                                                                                                   [ 75%]
handler_update.py::contract_update_read_success PASSED                                                                                                                                                                                                 [ 81%]
handler_update.py::contract_update_list_success PASSED                                                                                                                                                                                                 [ 87%]
handler_update.py::contract_update_create_only_property PASSED                                                                                                                                                                                         [ 93%]
handler_update.py::contract_update_non_existent_resource PASSED                                                                                                                                                                                        [100%]


```

### DIFF
--- a/src/rpdk/core/contract/suite/handler_update.py
+++ b/src/rpdk/core/contract/suite/handler_update.py
@@ -5,14 +5,14 @@ import pytest
 
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
-from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
+from rpdk.core.contract.interface import Action, OperationStatus
 from rpdk.core.contract.suite.handler_commons import (
     test_model_in_list,
     test_read_success,
 )
 
 
-@pytest.fixture(scope="function")
+@pytest.fixture(scope="module")
 def updated_resource(resource_client):
     create_request = model = resource_client.generate_create_example()
     try:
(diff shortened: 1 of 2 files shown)


src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -12,7 +12,7 @@
 )
 
 
-@pytest.fixture(scope="module")
+@pytest.fixture(scope="function")
```
jotompki, line 15: this will severely slow down tests. what's the justification? the update tests are written in a way where this fixture can be created once before tests run and the same fixture is used in all tests

cathlh: Cortez and I were running into an error for contract_update_create_only_property(), where it's automatically failing on the [create call](https://github.com/aws-cloudformation/cloudformation-cli/blob/0d71a6520ba364c94301f6b8659c203ed4b4d5a3/src/rpdk/core/contract/suite/handler_update.py#L63) because the resource from the previous test function is never cleaned up since the fixture is using a modular scope

anshikg: If you are worried about slowing  the test down, we could rip these 2 test: contract_update_create_only_property and  contract_update_non_existent_resource into another class which does not have a fixture.

anshikg: As far as I understand now, the fixture is being created twice instead of once, as only 2 tests require it. But there is a risk of the number of tests increasing in this class causing the tests to slow down

LaikaN57: I think that the update tests are a bit special in that they are testing multiple lifecycle event happening on a resource (because we inherently have to do creates first). There is not really a way to test _only_ the update functionality but rather the interplay of create-update.

PR #500, opened 2020-07-20
added asserts into a library

ammokhov:
*Issue #, if available:*

*Description of changes:*
adding a module where we could concentrate all the checks or asserts necessary for tests cases and decorate them.

It has a parent method `decorate` which makes decorator out of any simple function. The convention is to have variables that could be either returned by the target function (which is `response` if not returned then `None`) or accepted by the target function (e.g. `resource_client`). `decorate` rebinds `positional` or `keyword_or_positional` args from the target function and rebinds it to the assumed decorator. 

Then it could be used to decorate any method, which response needs to be checked upon `write_only_properties` or something else

### DIFF
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -1,16 +1,17 @@
 from functools import wraps
 from inspect import Parameter, signature
 
+
 def _rebind(decorator, func, *args, **kwargs):
     """Helper function to construct decorated arguments
 
     This works only with positional and likely positional arguments
-    strongly keyword arguments are in **kwargs. It constructs kwargs' 
+    strongly keyword arguments are in **kwargs. It constructs kwargs'
     from positional values
     """
     parameters = signature(func).parameters.values()
     decorated_parameters = set(signature(decorator).parameters.keys())
-    
+
     positional_kwargs = dict(
         zip(
             [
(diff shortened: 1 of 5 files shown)


src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -0,0 +1,90 @@
...
+
+@decorate
+def resource_model_equal_updated(response, current_resource_model, update_resource_model):
+    assert response["resourceModel"] == {
+            **current_resource_model,
+            **update_resource_model,
+        }
+
+@decorate
+def primary_identifier(resource_client, response):
```
rjlohan, line 68: Can we rename this to something more expressive, like `response_contains_primary_identifier` ?

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -0,0 +1,90 @@
...
+            decorated_sig = signature(decorator)
+            bound_arguments = decorated_sig.bind(**kvargs)
+
+            decorator(*bound_arguments.args, **bound_arguments.kwargs) # calling a decorated funciont to execute check
+            return response
+        return function
+    return new_decorator
+
+@decorate
+def write_only_properties(resource_client, response):
```
rjlohan, line 51: Can we rename this to something more expressive, like `response_does_not_contain_write_only_properties`?

Issue #501, opened 2020-07-20
Add "region", "partition" and "accountId" value to contract test 


anshikg:
Contract test request model does not have region/partition/accountId value. These properties need to be added to the request model so that the right values can be picked up when running contract test locally.

PR #502, opened 2020-07-20
Add region, partition and accountId to contract test request to handler

anshikg:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/501

*Description of changes:* Contract test does not provide a mechanism to add Region/Partition/AccountId to the contract test request. This change will allow users to add the above mentioned values as parameters.

*Test:* Ran the following command successfully:
```
(env) anshikg@3c22fbb25c23 aws-elasticloadbalancing-loadbalancer % cfn test --region "us-west-2" --partition "aws" --account-id "12345678" -- -k contract_create_delete
==================================================================================================================== test session starts =====================================================================================================================
platform darwin -- Python 3.8.4, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /Users/anshikg/workspace/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-elasticloadbalancing/aws-elasticloadbalancing-loadbalancer/.hypothesis/examples')
rootdir: /Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-elasticloadbalancing/aws-elasticloadbalancing-loadbalancer, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_rnt4wt90.ini
collected 16 items / 15 deselected / 1 selected                                                                                                                                                                                                              

handler_create.py::contract_create_delete PASSED          
```


### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -14,6 +14,7 @@ from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStat
 from ..boto_helpers import (
     LOWER_CAMEL_CRED_KEYS,
     create_sdk_session,
+    get_account,
     get_temporary_credentials,
 )
 from ..jsonutils.pointer import fragment_decode, fragment_list
@@ -85,6 +86,9 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._creds = get_temporary_credentials(
             self._session, LOWER_CAMEL_CRED_KEYS, role_arn
         )
+        self.region = region
+        self.account = get_account(self._session)
+        self.partition = self._get_partition()
         self._function_name = function_name
         if endpoint.startswith("http://"):
             self._client = self._session.client(
@@ -111,6 +115,13 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._update_schema(schema)
         self._inputs = inputs
         self._timeout_in_seconds = int(timeout_in_seconds)
+
+    def _get_partition(self):
+        if self.region.startswith("cn"):
+            return "aws-cn"
+        if self.region.startswith("us-gov"):
+            return "aws-gov"
+        return "aws"
 
     def _properties_to_paths(self, key):
         return {fragment_decode(prop, prefix="") for prop in self._schema.get(key, [])}
@@ -277,12 +288,22 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         return error_code
 
     @staticmethod
-    def make_request(desired_resource_state, previous_resource_state, **kwargs):
+    def make_request(
+        desired_resource_state,
+        previous_resource_state,
+        region,
+        account,
+        partition,
+        **kwargs
+    ):
         return {
             "desiredResourceState": desired_resource_state,
             "previousResourceState": previous_resource_state,
             "logicalResourceIdentifier": None,
             **kwargs,
+            "region": region,
+            "awsPartition": partition,
+            "awsAccountId": account,
         }
 
     @staticmethod
@@ -341,7 +362,14 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     ):
         if assert_status not in [OperationStatus.SUCCESS, OperationStatus.FAILED]:
             raise ValueError("Assert status {} not supported.".format(assert_status))
-        request = self.make_request(current_model, previous_model, **kwargs)
+        request = self.make_request(
+            current_model,
+            previous_model,
+            self.region,
+            self.account,
+            self.partition,
+            **kwargs
+        )
 
         status, response = self.call(action, request)
         if assert_status == OperationStatus.SUCCESS:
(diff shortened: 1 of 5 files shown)


src/rpdk/core/invoke.py
```diff
@@ -23,7 +23,13 @@ def invoke(args):
...
 
     client = ResourceClient(
-        args.function_name, args.endpoint, args.region, project.schema, {}
+        args.function_name,
+        args.endpoint,
+        args.region,
+        args.partition,
+        args.account_id,
+        project.schema,
+        {},
```
ammokhov, line 32: what's that for? why empty `dict` is injected?

src/rpdk/core/invoke.py
```diff
@@ -23,7 +23,13 @@ def invoke(args):
     project.load()
 
     client = ResourceClient(
-        args.function_name, args.endpoint, args.region, project.schema, {}
+        args.function_name,
+        args.endpoint,
+        args.region,
+        args.partition,
+        args.account_id,
+        project.schema,
```
ammokhov, line 31: are these all positional or keyword or both?

src/rpdk/core/contract/resource_client.py
```diff
@@ -277,8 +282,18 @@ def assert_failed(status, response):
...
+        partition,
+        account_id,
+        **kwargs
+    ):
         return {
+            "region": region,
+            "awsPartition": partition,
+            "awsAccountId": account_id,
             "desiredResourceState": desired_resource_state,
             "previousResourceState": previous_resource_state,
```
ammokhov, line 298: I'd prefer to preserve the order of the inputs

src/rpdk/core/invoke.py
anshikg: empty dict is for overrides. This is not a change which was made in this PR.

src/rpdk/core/contract/resource_client.py
anshikg: I don't think order matters here. But, I can go ahead and change it.

src/rpdk/core/invoke.py
anshikg: positional

README.md
```diff
@@ -42,6 +42,8 @@ cfn test
 cfn test -- -k contract_delete_update #to run a single test
 cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD handler timeout to 120 seconds.
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
+cfn test --region "us-west-2" --partition "aws" --account-id "12345678" #allows users to set a valid region, account-id and partition.
```
jotompki, line 45: you shouldn't need to pass in the account id or partition. All of this should be available from the session we are using:

https://docs.aws.amazon.com/cli/latest/reference/sts/get-caller-identity.html

src/rpdk/core/invoke.py
ammokhov: usually positional are marked as `/`

src/rpdk/core/invoke.py
ammokhov: injecting an empty dict does not seem right.. should we set a default to `None` in the target function?

README.md
rjlohan: So I had to [use the partition/region](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/pull/4/files#diff-3fcebf9413b88490d8d6d7c17cb290cdR134) in my `AWS::CloudFormation::ResourceVersion` type, because I need to predict an Arn to comply with sync response requirement on Create. Can I extract this from the context I have available in the handler?

src/rpdk/core/contract/resource_client.py
rjlohan: There might be some expectation in the code (I hope there's *not*) that uses positional values from the map instead of key lookups.

src/rpdk/core/invoke.py
anshikg: I am relunctant to make an unrelated code change in this CR.

README.md
ninavolkmuth: Ideally we shouldn't require any additional permissions to run the contract tests other than the permissions needed to create/update/delete their resources. We could add default behavior to call the api, but we should allow people to pass in the account id as well

tests/test_boto_helpers.py
```diff
@@ -197,3 +198,12 @@ def test_get_temporary_credentials_assume_role():
...
     assert tuple(creds.values()) == (access_key, secret_key, token)
+
+
+def test_get_account():
+    session = create_autospec(spec=Session, spec_set=True)
+    client = session.client.return_value
+    get_account(session)
+
+    session.client.assert_called_once_with("sts")
+    client.get_caller_identity.assert_called_once()
```
jotompki, line 209: could we add an assert to check the return value of the function as well?

Issue #507, opened 2020-07-22
contract test with read only primary identifier fail because a null identifier is passed to the service


rjlohan:
The test starts by producing a model with the `resource_client.generate_create_example` implementation. However when the resource identifier is generated by the service, it is part of the `readOnlyProperties` in the schema and cannot be generated this way. We should allow for an identifier generation strategy (or overrides, or inputs) to support this test.

ammokhov:
what is the point of this test? to check if read handler returns not found exception? I think it's sufficient to run `contract_delete_read` which in fact has a good primary id (taken from previous actions). It's pretty hard to generate primary id using `generate_create_example` considering how strict pattern should be (it could include partition, region, and even more constratins)

anticobalt:
Also having this issue with `contract_update_non_existent_resource` when the resource has one read-only primary identifier, and with `contract_delete_create` when the resource has two identifiers (one read-only, one create-only).

anshikg:
@anticobalt could you elaborate more on what is happening in this test: contract_delete_create ?

anticobalt:
@anshikg The test seems to be doing a create-delete-create-delete sequence. Both create requests succeed. The first delete also succeeds, as both the read-only and create-only identifiers are included in the delete request. The second delete fails, because the read-only identifier is null. Delete handler cannot return expected `NotFound` error because it can't look up the resource without the identifier.

The read-only identifier is the only read-only property of the resource I'm testing on.

I believe [this](https://github.com/aws-cloudformation/cloudformation-cli/issues/329) issue/merge is related.

anshikg:
@anticobalt I see your point about the contract_delete_create I don't believe it is related to this issue. I will create a new issue for this contract test and track it there.

ammokhov:
closing this issues as it's been addressed in #515

PR #515, opened 2020-07-29
added a writable id skip for read without create

ammokhov:
*Issue #, if available:* #507 

*Description of changes:*

**contract_read_without_create**

Test anticipates `NotFound` error as it invokes `ReadHandler` without prior creation of the resource. This test starts with `resource_client.generate_create_example` but it's pretty hard to provide a good generation strategy for primary identifier to mock a good value, as primary identifier pattern is not always good enough to generate a good substitute (e.g. `arn` - partition or region constraints). This test might be very inconsistent if we rely on generation strategy, as different api might throw different error - `NotFound`, `InvalidRequest`, `CfnGeneralException`, etc. 

Eventually, this test checks if `ReadHandler` throws `NotFound` exception in case the resource has already been deleted. This could be achieved by `contract_delete_read` test, which in fact operates with a good identifier. 

~~This change skips the test for service generated identifiers~~

**UPDATE**

Changed generation strategy for `contract_read_without_create` to allow invalid input. Requirement: inputs_*n*_invalid.json should contain *invalid* primary identifier but correctly formatted so that target api will accept it and produces `NotFound` error

additional changes: made `decorate` meta function to accept argument {after} this allows to decorate checks that should be run before the target function is executed.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_create.py
+++ b/src/rpdk/core/contract/suite/handler_create.py
@@ -7,7 +7,10 @@ import pytest
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
 from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
-from rpdk.core.contract.suite.contract_asserts import failed_event
+from rpdk.core.contract.suite.contract_asserts import (
+    failed_event,
+    skip_not_writable_identifier,
+)
 from rpdk.core.contract.suite.handler_commons import (
     test_create_failure_if_repeat_writeable_id,
     test_create_success,
@@ -63,6 +66,7 @@ def _create_with_invalid_model(resource_client):
 
 
 @pytest.mark.create
+@skip_not_writable_identifier
 def contract_create_duplicate(created_resource, resource_client):
     _created_model, request = created_resource
     test_create_failure_if_repeat_writeable_id(resource_client, request)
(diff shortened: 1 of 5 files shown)


anshikg:
> If the test is redundant can we just remove it

The test is not redundant. We will run into resources which do a soft delete. This will not allows us to check for the "NotFound" Error code. Read without create bypasses that issue and just tries to read something which was never created.

Anton, could you update the detailed section to reflect our offline conversation.

src/rpdk/core/contract/suite/handler_read.py
```diff
@@ -10,5 +10,7 @@
 
 @pytest.mark.read
 def contract_read_without_create(resource_client):
-    model = resource_client.generate_create_example()
+    model = (
+        resource_client.generate_invalid_create_example()
```
jotompki, line 14: don't understand the comment here. is this related to the writable id skip?

src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -63,6 +66,7 @@ def _create_with_invalid_model(resource_client):
 
 
 @pytest.mark.create
+@skip_not_writable_identifier
```
jotompki, line 69: this marker is only used once? are there more potential uses down the line? as an aside i think we should inline the `test_create_failure_if_repeat_writeable_id` since it's only being called in this method as well for readability

ammokhov: I think it could be used more than once but I am not sure

src/rpdk/core/contract/suite/handler_read.py
ammokhov: not really it's no longer id skip. it's more like to use invalid primary id to assert read handler throws not found

Issue #510, opened 2020-07-23
Contract Test Decoding Issue


xaoyuli:
I was writing the contract tests for AWS::Redshift::ClusterSubnetGroup https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-redshift
My handlers passed the inte test, but I failed on almost all of the contract test. And the error at all about decoding

I wrote the override file
```
{
  "CREATE": {
    "/SubnetIds": ["XXXXXXXX"],
    "/Description": "cfn test"
  }
}
```
The command I run
```
cfn test --enforce-timeout 180 -- -k contract_create_delete
```
I got the result below
```
==================================================================== FAILURES ====================================================================
_____________________________________________________________ contract_create_delete _____________________________________________________________

resource_client = <rpdk.core.contract.resource_client.ResourceClient object at 0x1100b89d0>

    @pytest.mark.create
    @pytest.mark.delete
    def contract_create_delete(resource_client):
        requested_model = delete_model = resource_client.generate_create_example()
        try:
            response = test_create_success(resource_client, requested_model)
            # check response here
            delete_model = response["resourceModel"]
        finally:
>           test_delete_success(resource_client, delete_model)

/usr/local/lib/python3.7/site-packages/rpdk/core/contract/suite/handler_create.py:43: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/usr/local/lib/python3.7/site-packages/rpdk/core/contract/suite/handler_commons.py:108: in test_delete_success
    Action.DELETE, OperationStatus.SUCCESS, current_resource_model
/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/rpdk/core/contract/resource_client.py:346: in call_and_assert
    status, response = self.call(action, request)
/usr/local/Cellar/python/3.7.7/Frameworks/Python.framework/Versions/3.7/lib/python3.7/site-packages/rpdk/core/contract/resource_client.py:357: in call
    response = self._call(payload)
[...]

PR #491, opened 2020-07-16
Update pytest log printing option

LaikaN57:
*Issue #, if available:*
The following warning is generated when running `cfn test`:
```
================================================================= warnings summary ==================================================================
.venv/lib/python3.7/site-packages/_pytest/logging.py:499
  .venv/lib/python3.7/site-packages/_pytest/logging.py:499: PytestDeprecationWarning: --no-print-logs is deprecated and scheduled for removal in pytest 6.0.
  Please use --show-capture instead.
    _issue_warning_captured(NO_PRINT_LOGS, self._config.hook, stacklevel=2)
```

*Description of changes:*
Migrate to new pytest option to resolve warning.

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.

PR #518, opened 2020-07-30
increasing default lambda memory size

ammokhov:
*Issue #, if available:* #510 

*Description of changes:*

increasing the default memory size

bumping pytest to 6.x as annotations were not recognized as callable  https://github.com/pytest-dev/pytest/issues/7473

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,7 @@ ipdb>=0.12
 # testing tools
 pylint>=2.3.1
 coverage>=4.5.4
-pytest>=5.3.5
+pytest>=6.0.0
 pytest-cov>=2.7.1
 pytest-random-order>=1.0.4
 hypothesis>=4.32.3
--- a/src/rpdk/core/templates/template.yml
+++ b/src/rpdk/core/templates/template.yml
@@ -13,6 +13,7 @@ Description: AWS SAM template for the {{ resource_type }} resource type
 Globals:
   Function:
     Timeout: 180  # docker start-up times can be long for SAM CLI
+    MemorySize: 256
 
 Resources:
 {% if functions %}


requirements.txt
```diff
@@ -5,7 +5,7 @@ ipdb>=0.12
 # testing tools
 pylint>=2.3.1
 coverage>=4.5.4
-pytest>=5.3.5
+pytest>=6.0.0
```
PatMyron, line 8: [`--no-print-logs` needs to be removed for `pytest` 6.0](https://github.com/aws-cloudformation/cloudformation-cli/pull/491)

LaikaN57:
@ammokhov We have to manually apply this config to existing CFN RPs correct? If so, is there a better way that CFN RP developers could get notified about these types of changes?

cc: @jotompki

ammokhov:
> @ammokhov We have to manually apply this config to existing CFN RPs correct? If so, is there a better way that CFN RP developers could get notified about these types of changes?
> 
> cc: @jotompki

we dont, it's a common package and if other prs do not have merge conflicts then you can just restart travis job

LaikaN57:
@ammokhov Sorry, I did not understand your response, let me re-ask my question... There was a similar template change (rpdk protocol upgrade, if i remember correctly) where the new template was not applied to existing CFN RPs that end-users made with cfn-cli (ie. `cfn generate` had no effect on existing CFN RP files.)

I am asking if this template change will similarly need to be manually applied to end-user CFN RPs files or if the `cfn generate` command will overwrite the existing files that this template defines.

ammokhov:
> @ammokhov Sorry, I did not understand your response, let me re-ask my question... There was a similar template change (rpdk protocol upgrade, if i remember correctly) where the new template was not applied to existing CFN RP flies that end-users made with cfn-cli (ie. `cfn generate` had no effect on existing CFN RP files.)
> 
> I am asking if this template change will similarly need to be manually applied to end-user CFN RPs files or if the `cfn generate` command will overwrite the existing files that this template defines. To change protocol version from my above example, I needed to manually edit the `.rpdk-config` file.

Sorry, ignore previous reply I misunderstood your question. 
you are right `cfn generate` will not modify existing `template.yml` which should be manually changed, as this file is generate once on the project creation. This is a static file and not a subject to change based off schema modifications like `ResourceModel` or resource specs

PR #520, opened 2020-07-31
Fix Contract test for resources with one createOnly Property

anshikg:
*Issue #, if available:*

*Description of changes:* Contract test which have just one create only property as identifier fail with NotFound error code instead of NotUpdatable

*Test*
* Ran the test against cloudwatch alarm resource

### DIFF
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -2,9 +2,9 @@ from functools import wraps
 from inspect import Parameter, signature
 from unittest.test.testmock.support import is_instance
 
-from rpdk.core.contract.interface import HandlerErrorCode
-
 import pytest
+
+from rpdk.core.contract.interface import HandlerErrorCode
 
 
 def _rebind(decorator, func, *args, **kwargs):
@@ -126,9 +126,8 @@ def failed_event(error_code, msg=""):
             response_error = func(*args, **kwargs)
             if response_error is not None:
                 if is_instance(error_code, HandlerErrorCode):
-                    assert response_error == error_code, msg
-                else:
-                    assert response_error in error_code
+                    error_code_tuple = (error_code,)
+                assert response_error in error_code_tuple, msg
             return response_error
 
         return wrapper
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -1,5 +1,8 @@
 from functools import wraps
 from inspect import Parameter, signature
+from unittest.test.testmock.support import is_instance
+
+from rpdk.core.contract.interface import HandlerErrorCode
 
 import pytest
 
@@ -120,10 +123,13 @@ def failed_event(error_code, msg=""):
     def decorator_wrapper(func: object):
         @wraps(func)
         def wrapper(*args, **kwargs):
-            response = func(*args, **kwargs)
-            if response is not None:
-                assert response == error_code, msg
-            return response
+            response_error = func(*args, **kwargs)
+            if response_error is not None:
+                if is_instance(error_code, HandlerErrorCode):
+                    assert response_error == error_code, msg
+                else:
+                    assert response_error in error_code
+            return response_error
 
         return wrapper
 
--- a/src/rpdk/core/contract/suite/handler_update_invalid.py
+++ b/src/rpdk/core/contract/suite/handler_update_invalid.py
@@ -26,8 +26,9 @@ def contract_update_create_only_property(resource_client):
                 Action.UPDATE, OperationStatus.FAILED, update_request, created_model
             )
             assert response["message"]
-            assert (
-                _error == HandlerErrorCode.NotUpdatable
+            assert _error in (
+                HandlerErrorCode.NotUpdatable,
+                HandlerErrorCode.NotFound,
             ), "updating readOnly or createOnly properties should not be possible"
         finally:
             resource_client.call_and_assert(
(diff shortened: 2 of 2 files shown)


LaikaN57:
@anshikg `NotFound` does not seem like an appropriate response code for this as the test fixture should have successfully created a resource which can be found and acted on by the update handler. Can you explain a little more about the scenario you are seeing?

LaikaN57:
@anshikg I feel like `NotFound` is not quite appropriate here. It should almost be the code's responsibility to check for this.

For example, if you have:
```
MyRes:
  Type: MyOrg::MySvc::MyRes
  Properties:
    ID: 1 # create only / primary id
```
and then you try to update to:
```
MyRes:
  Type: MyOrg::MySvc::MyRes
  Properties:
    ID: 2 # create only / primary id
```

The update handler code should be checking that no create only properties have been updated. Furthermore, it should send back the failed message using the previous state model's primary id, not the desired state model's (would CFN get confused by seeing an id it does not know about?). Actually, I think that CFN should be doing the create only property or primary id change check before it ever makes it to the CFN RP handler.

Maybe I am fundamentally misunderstanding something here, but it is my understanding that physical id == primary id and shall never be modified once assigned.

ammokhov:
> @anshikg I feel like `NotFound` is not quite appropriate here. It should almost be the code's responsibility to check for this.
> 
> For example, if you have:
> 
> ```
> MyRes:
>   Type: MyOrg::MySvc::MyRes
>   Properties:
>     ID: 1 # create only / primary id
> ```
> 
> and then you try to update to:
> 
> ```
> MyRes:
>   Type: MyOrg::MySvc::MyRes
>   Properties:
>     ID: 2 # create only / primary id
> ```
> 
> The update handler code should be checking that no create only properties have been updated. Furthermore, it should send back the failed message using the previous state model's primary id, not the desired state model's (would CFN get confused by seeing an id it does not know about?). Actually, I think that CFN should be doing the create only property or primary id change check before it ever makes it to the CFN RP handler.
> 
> Maybe I am fundamentally misunderstanding something here, but it is my understanding that physical id == primary id and shall never be modified once assigned.

This is true. `NotFound` does not seem to be a correct assertion. This test has good intentions but in reality `ReadHandler` will never be reached as the instructions are to DELETE and CREATE. Maybe we should rethink the need of this test

src/rpdk/core/contract/suite/handler_update_invalid.py
```diff
@@ -26,8 +26,9 @@ def contract_update_create_only_property(resource_client):
                 Action.UPDATE, OperationStatus.FAILED, update_request, created_model
             )
             assert response["message"]
-            assert (
-                _error == HandlerErrorCode.NotUpdatable
+            assert _error in (
+                HandlerErrorCode.NotUpdatable,
+                HandlerErrorCode.NotFound,
```
ammokhov, line 31: can u use `@failed_event`?

anshikg:
> @anshikg I feel like `NotFound` is not quite appropriate here. It should almost be the code's responsibility to check for this.
> 
> For example, if you have:
> 
> ```
> MyRes:
>   Type: MyOrg::MySvc::MyRes
>   Properties:
>     ID: 1 # create only / primary id
> ```
> 
> and then you try to update to:
> 
> ```
> MyRes:
>   Type: MyOrg::MySvc::MyRes
>   Properties:
>     ID: 2 # create only / primary id
> ```
> 
> The update handler code should be checking that no create only properties have been updated. Furthermore, it should send back the failed message using the previous state model's primary id, not the desired state model's (would CFN get confused by seeing an id it does not know about?). Actually, I think that CFN should be doing the create only property or primary id change check before it ever makes it to the CFN RP handler.
> 
> Maybe I am fundamentally misunderstanding something here, but it is my understanding that physical id == primary id and shall never be modified once assigned.

Let me try and clarify what I was trying to do. I am going to try and use the above example. Lets say we have a resource with primary identifier "1" (also a create only property). Now, we call update handler for this resource and pass primaryIdentifier "2". A resource with primary identifier "2" does not exist. In this case, because the createOnly property and primary identifier are the same value the update handler does not know what the users intent was. The handler can not know whether the user was actually trying to update the resource with primary identifier "2" (in which case NotFound is the right error code) or if the user was trying to update the resource with primary identifier "1"(probably made a typo). This is why "NotFound" is thrown for resource with primary identifier as createOnly property. The Update handler checks if a resource with primary identifier "2" exists. If none exist it throws and exception. Does this clarify the concerns?

src/rpdk/core/contract/suite/handler_update_invalid.py
anshikg: sure, I will update that

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -120,10 +123,13 @@ def failed_event(error_code, msg=""):
...
-            response = func(*args, **kwargs)
-            if response is not None:
-                assert response == error_code, msg
-            return response
+            response_error = func(*args, **kwargs)
+            if response_error is not None:
+                if is_instance(error_code, HandlerErrorCode):
+                    assert response_error == error_code, msg
+                else:
+                    assert response_error in error_code
```
ammokhov, line 131: this is great. can u do it this way?
```python
list_errors = error_code
if isinstance(list_errors, HandlerErrorCode):
    list_errors = [list_errors]

assert response_error in error_code, msg
```

LaikaN57:
@anshikg Sorry for the delay. Yes this makes sense now from the CFN RP perspective.

My concern is just that the example "update" we are doing should be illegal from a CFN service perspective in the first place and our CFN RP handler should never be called to reach this scenario. I would rather see a fix/check in the CFN service itself. CFN should be checking that the physical ID remains the same, but maybe the service does not have this ability.

I can agree that `NotFound` is an okay code for this PR and repo. This just makes code extra safe which is probably fine. I just don't want it to mask a higher issue in the CFN service itself.

Issue #517, opened 2020-07-29
Resource not deleted after contract_invalid_create contract test.


ugudip:
The contract_invalid_create contract test checks for any invalid parameters in the request. This test doesn't delete the resource after [creating one.](https://github.com/aws-cloudformation/cloudformation-cli/blob/aeef95c4abc9df8f6cffa9045b8f627bbb1056d6/src/rpdk/core/contract/suite/handler_create.py#L59) 

Use case: My resource has to create few properties soon after creating the actual resource in the same create handler. If one of those properties is invalid, only that property doesn't get created. In this case, even though the progress event is FAILED, the resource still exists(partial create). This results in the failure of the next set of contract tests(contract_create_duplicate - ERROR, contract_create_read_success - ERROR, contract_create_list_success - ERROR). However, those tests pass when invoked individually.

It would be great if the delete handler is called at the end of contract_invalid_create contract test, similar to other contract tests.
<img width="511" alt="Screen Shot 2020-07-29 at 4 27 03 PM" src="https://user-images.githubusercontent.com/65369219/88863729-59dc0300-d1b8-11ea-92d8-5df6e3304b35.png">

PR #521, opened 2020-07-31
Delete partial resource created by invalid create test

anshikg:
*Issue #, if available:*https://github.com/aws-cloudformation/cloudformation-cli/issues/517

*Description of changes:* Adding a delete call to delete partially created resources.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_create.py
+++ b/src/rpdk/core/contract/suite/handler_create.py
@@ -54,12 +54,15 @@ def contract_invalid_create(resource_client):
 
 @failed_event(error_code=HandlerErrorCode.InvalidRequest)
 def _create_with_invalid_model(resource_client):
-    requested_model = resource_client.generate_invalid_create_example()
-    _status, response, _error_code = resource_client.call_and_assert(
-        Action.CREATE, OperationStatus.FAILED, requested_model
-    )
-    assert response["message"]
-    return _error_code
+    try:
+        requested_model = resource_client.generate_invalid_create_example()
+        _status, response, _error_code = resource_client.call_and_assert(
+            Action.CREATE, OperationStatus.FAILED, requested_model
+        )
+        assert response["message"]
+        return _error_code
+    finally:
+        resource_client.call(Action.DELETE, requested_model)
 
 
 @pytest.mark.create


LaikaN57:
Or maybe for consistency we should be asserting as well like in the test fixture for all of the other tests? https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_create.py#L32

ammokhov:
> The return before finally was a bit confusing for new python devs but interesting functionality. Also I guess if the delete operation fails (due to delete code bugs) then test will provided and ERROR status instead of FAILURE which seems ok but sort of makes writing a DELETE handler a requirement of writing and testing a CREATE handler. Overall it looks good though.


Good point, not sure if `DeleteHandler` must be a requirement for the test but We should definitely try to delete created resource to avoid unexpected billing. 

> Or maybe for consistency we should be asserting as well like in the test fixture for all of the other tests? https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_create.py#L32




I think we could put `finally` as additional annotation and add that to all the necessary tests. In addition, I agree that we should be consistent and use [call_and_assert](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_create.py#L32)

src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -54,12 +54,15 @@ def contract_invalid_create(resource_client):
...
-    return _error_code
+    try:
+        requested_model = resource_client.generate_invalid_create_example()
+        _status, response, _error_code = resource_client.call_and_assert(
+            Action.CREATE, OperationStatus.FAILED, requested_model
+        )
+        assert response["message"]
+        return _error_code
+    finally:
+        resource_client.call(Action.DELETE, requested_model)
```
ammokhov, line 65:
```python
resource_client.call_and_assert(Action.DELETE, OperationStatus.SUCCESS, model)
```

anshikg: this is on purpose. Not all resources will create partial resource. If, the resource is not created we don't want to fail the test.

LaikaN57: I agree with @anshikg we should actually correct the test fixture.

LaikaN57: I have general thoughts around this and test error vs failure. I will create an issue with my thoughts.

ammokhov: > this is on purpose. Not all resources will create partial resource. If, the resource is not created we don't want to fail the test.

makes sense, I forgot that the testing framework does not skip `NotFoundException`

PR #522, opened 2020-08-02
added inputs generation on [cfn init]

ammokhov:
*Issue #, if available:*

*Description of changes:*

added `inputs`/{`inputs_1_create.json`, `inputs_1_update.json`, `inputs_1_invalid.json`} on `cfn init`

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -33,6 +33,7 @@ repos:
     - flake8-pep3101>=1.2.1
     # language_version: python3.6
   - id: pretty-format-json
+    exclude: inputs.json
     args:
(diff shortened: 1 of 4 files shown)


ammokhov:
> Is this helpful if we are only doing it on init with the example schema? Since this doesn't get updated on generate I don't see the use. Will just be a json blob that won't be used since it's properties in the example schema

that's true we cant update it, but at least we generate an example, thereby enforcing the correct development process. At least a developer will see what are the required files and what is the sample for the generated example.

I would argue that we need to do the same for the sam tests as lot's of devs dont bother and test via registering through cfn and end up having 200 versions locally.

Issue #503, opened 2020-07-21
[contract tests] Soft Delete/delayed delete source fail delete contract test


ammokhov:
This tests confirms that after delete the resource cannot appear in the subsequent list request. This is true for most of the resource but there are some exceptions. This tests will be failing for resources that cannot be immediately deleted and delete handler assumes a certain state as `deleted` interpret-able. One example would be `AWS::KMS::Key` which can only be scheduled for deletion, hence the state might be only `PendingDeletion`

> Schedules the deletion of a customer master key (CMK). You may provide a waiting period, specified in days, before deletion occurs. If you do not provide a waiting period, the default period of 30 days is used. When this operation is successful, the key state of the CMK changes to PendingDeletion.

anshikg:
It would be helpful if you could list the test cases which are failing for you. Also, a stack trace for the ones which fail with the delete issue.

shunjd:
We have similar issues when testing deletion for the StateMachine in Step Functions. There are two types of StateMachine, EXPRESS and STANDARD, the deletion behaviors are different. For EXPRESS, the deletion is eventual consistent, so `contract_delete_read` may fail because `read` will still return the resource if it is being called subsequently within a short interval. For STANDARD, we have a soft deletion mechanism, because in order to delete a standard statemachine, Step Functions need to first put it into DELETING status and then start to clean up all the executions (this is not in realtime and will be scheduled internally).

ammokhov:
> We have similar issues when testing deletion for the StateMachine in Step Functions. There are two types of StateMachine, EXPRESS and STANDARD, the deletion behaviors are different. For EXPRESS, the deletion is eventual consistent, so `contract_delete_read` may fail because `read` will still return the resource if it is being called subsequently within a short interval. For STANDARD, we have a soft deletion mechanism, because in order to delete a standard statemachine, Step Functions need to first put it into DELETING status and then start to clean up all the executions (this is not in realtime and will be scheduled internally).

one of the possible and quick solutions would be to have a status check in read/list handler if status is `DELETING` a `NotFound` should be thrown

anshikg:
As discussed in the Pull Request. These resources are not in alignment with the handler Contracts. Although, some resources do have the functionality of soft/delayed delete but, as of now this is not supported by CFN.

PR #523, opened 2020-08-03
Enable Soft delete for resources

anshikg:
*Issue #, if available:*https://github.com/aws-cloudformation/cloudformation-cli/issues/503

*Description of changes:* Some resources are not deleted right away when delete is called on the resource. This CR allows users to ensure their contract test pass if their resource supports soft delete.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -80,6 +80,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         inputs=None,
         role_arn=None,
         timeout_in_seconds="30",
+        soft_delete="false",
     ):  # pylint: disable=too-many-arguments
         self._schema = schema
         self._session = create_sdk_session(region)
@@ -115,6 +116,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         self._update_schema(schema)
         self._inputs = inputs
         self._timeout_in_seconds = int(timeout_in_seconds)
+        self.soft_delete = bool(soft_delete == "true")
 
     def _get_partition(self):
         if self.region.startswith("cn"):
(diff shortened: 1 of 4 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -115,6 +116,7 @@ def __init__(
         self._update_schema(schema)
         self._inputs = inputs
         self._timeout_in_seconds = int(timeout_in_seconds)
+        self.soft_delete = bool(soft_delete == "true")
```
ammokhov, line 119: what's the diff between `bool(soft_delete == "true")` and `soft_delete == "true"`?

anshikg: I don think there is a difference. Will remove "bool"

src/rpdk/core/test.py
```diff
@@ -267,3 +268,8 @@ def _sam_arguments(parser):
             "The region used for temporary credentials " f"(Default: {DEFAULT_REGION})"
         ),
     )
+    parser.add_argument(
+        "--soft-delete",
+        default="false",
```
jotompki, line 273: you can just make this a boolean so you don't have to pass in a string.

use `store_true`: https://docs.python.org/3/library/argparse.html

LaikaN57:
The other option is that the handlers (probably read) should have enough information such that it will return NotFound on soft-deleted resources.

rjlohan:
> The other option is that the handlers (probably read) should have enough information such that it will return NotFound on soft-deleted resources.

Yes this is how I imagined compliance could be achieved. Example is the `AWS::CloudFormation::Stack` resource - which hangs around in the account in a `DELETE_COMPLETE` state, but is still partially describable (as an audit record). These can be dropped from the `List` and `Read` handlers to achieve compliance at this level.

LaikaN57:
But yeah, fundamentally, I do not see how adding a bypass option to the cfn-cli is correct. These contract tests should be testing what the CFN RP service is expecting. If the CFN RP service does not actually care if you deleted a resource or not, then there should be no contract test for it.

anshikg:
Closing this. After the discussion on this pull request. It does not seem this change is in alignment with the handler contract

PR #530, opened 2020-08-05
Remove contract test which is implemented by Lambda Wrapper

anshikg:
*Issue #, if available:*

*Description of changes:* Removing contract test which is asserted by Lambda Wrapper.
Lambda Wrapper by default checks for readOnly properties in the input for the handlers. This functionality does need to be implemented in the contract test.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_create.py
+++ b/src/rpdk/core/contract/suite/handler_create.py
@@ -6,11 +6,8 @@ import pytest
 
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
-from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
-from rpdk.core.contract.suite.contract_asserts import (
-    failed_event,
-    skip_not_writable_identifier,
-)
+from rpdk.core.contract.interface import Action, OperationStatus
+from rpdk.core.contract.suite.contract_asserts import skip_not_writable_identifier
 from rpdk.core.contract.suite.handler_commons import (
     test_create_failure_if_repeat_writeable_id,
     test_create_success,
@@ -45,27 +42,6 @@ def contract_create_delete(resource_client):
         delete_model = response["resourceModel"]
     finally:
         test_delete_success(resource_client, delete_model)
-
-
-@pytest.mark.create
-def contract_invalid_create(resource_client):
-    if resource_client.read_only_paths:
-        _create_with_invalid_model(resource_client)
-    else:
-        pytest.skip("No readOnly Properties. Skipping test.")
-
-
-@failed_event(error_code=HandlerErrorCode.InvalidRequest)
-def _create_with_invalid_model(resource_client):
-    try:
-        requested_model = resource_client.generate_invalid_create_example()
-        _status, response, _error_code = resource_client.call_and_assert(
-            Action.CREATE, OperationStatus.FAILED, requested_model
-        )
-        assert response["message"]
-        return _error_code
-    finally:
-        resource_client.call(Action.DELETE, requested_model)
 
 
 @pytest.mark.create


rjlohan:
Shouldn't the contract test continue to enforce this to support other language wrappers? It's great that we don't need to do anything to comply, but I think the check should still be there.

ammokhov:
> Shouldn't the contract test continue to enforce this to support other language wrappers? It's great that we don't need to do anything to comply, but I think the check should still be there.

fair point

jotompki:
> Shouldn't the contract test continue to enforce this to support other language wrappers? It's great that we don't need to do anything to comply, but I think the check should still be there.

Yeah I think that having the test in place is good. We attempt to remove writeOnlyProperties in the lambda wrapper but also have a test in contract tests to verify this as well. If someone wanted to go out and try to implement their own language plugin or make modifications it's good to have these in place.

anshikg:
> >> Shouldn't the contract test continue to enforce this to support other language wrappers? It's great that we don't need to do anything to comply, but I think the check should still be there.
> >
> >Yeah I think that having the test in place is good. We attempt to remove writeOnlyProperties in the lambda wrapper but also have a test in contract tests to verify this as well. If someone wanted to go out and try to implement their own language plugin or make modifications it's good to have these in place.
>
> I see your point. But, keeping this test means we are asking teams using java plugin to add redundant read only property checks which are only used for contract test. Shouldn't we just add this check in other language specific plugins?

LaikaN57:
By "Lambda Wrapper" are we referring to the the language specific plugins (ie. cloudformation-cli-python-plugin, cloudformation-cli-go-plugin, cloudformation-cli-java-plugin, etc.)? Or is this some other codebase or AWS internal thing?

rjlohan:
> By "Lambda Wrapper" are we referring to the the language specific plugins (ie. cloudformation-cli-python-plugin, cloudformation-cli-go-plugin, cloudformation-cli-java-plugin, etc.)? Or is this some other codebase or AWS internal thing?

Yes, means these plugins. There are 2 parts to the plugin repos - a codegen piece and a runtime wrapper piece. Currently we run on Lambda/SAM Local (for test) and we got in the habit of calling the runtime part 'Lambda Wrappers' though don't assume that will be the forever runtime. 👍

LaikaN57:
@rjlohan ok this was my knowledge base as well. Thanks for the confirmation.

LaikaN57:
Hmm... So the way I see it, we have a very merkey split of responsibility between the CFN RP service, the Lambda Wrapper, and the actual end-user CFN RP.

We need to ensure that:
- the CFN RP service is well documented
- all of the GA Lambda Wrappers conform to some spec if they are going to be doing validation checks
- contract tests are written to ensure that the end-user's CFN RP code complies with the CFN RP service, but after the Lambda Wrapper has validated/modified the request/response

From my understanding the contract tests should be testing a contract between the CFN RP service and the end-user CFN RP. But I did not consider that we have this swappable middle-man (the Lambda Wrapper) that **may** behave differently from implementation to implementation.

It may be worth writing a design doc to specify all of these parts.

jotompki:
I think 

> Hmm... So the way I see it, we have a very merkey split of responsibility between the CFN RP service, the Lambda Wrapper, and the actual end-user CFN RP.
> 
> We need to ensure that:
> 
>     * the CFN RP service is well documented
> 
>     * all of the GA Lambda Wrappers conform to some spec if they are going to be doing validation checks
> 
>     * contract tests are written to ensure that the end-user's CFN RP code complies with the CFN RP service, but after the Lambda Wrapper has validated/modified the request/response
> 
> 
> From my understanding the contract tests should be testing a contract between the CFN RP service and the end-user CFN RP. But I did not consider that we have this swappable middle-man (the Lambda Wrapper) that **may** behave differently from implementation to implementation.
> 
> It may be worth writing a design doc to specify all of these parts.

Would be cool to make some dummy resource providers in each language and run against the contract tests or something similar that test the functionality of each language wrapper to ensure parity. Could make the tests off of the java plugin's functionality then use the tests to ensure other plugins behave similarly.

jotompki:
> > > > Shouldn't the contract test continue to enforce this to support other language wrappers? It's great that we don't need to do anything to comply, but I think the check should still be there.
> > > 
> > > 
> > > Yeah I think that having the test in place is good. We attempt to remove writeOnlyProperties in the lambda wrapper but also have a test in contract tests to verify this as well. If someone wanted to go out and try to implement their own language plugin or make modifications it's good to have these in place.
> > 
> > 
> > I see your point. But, keeping this test means we are asking teams using java plugin to add redundant read only property checks which are only used for contract test. Shouldn't we just add this check in other language specific plugins?

I think the real solution here is to make contract tests use the production entrypoint so we don't have to port functionality to the test entrypoint. Since handlers are now synchronous with the new protocol, this should be pretty straightforward. This came up again in https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/296

anshikg:
Closing this pull request, the above suggestions are to keep the invalid create contract test to check this functionality across all the language specific plugins because, not all plugins will have this functionality implemented.

PR #531, opened 2020-08-06
Assert input to the handler is equal to the output

anshikg:
*Issue #, if available:*

*Description of changes:*
This CR checks if the input to the create/update handler is equal to the output. 
* WriteOnly Properties are removed from the input as the output should not have writeOnlyProperty. 
* ReadOnlyProperties are removed from the output as the input should not have those. 
* CreateOnly Properties are removed from input and output because, these values can be generated by the handler or entered from the customer. 

*Test*
* Ran these tests against CloudWatch Alarm resource:
```
platform darwin -- Python 3.8.5, pytest-5.4.3, py-1.9.0, pluggy-0.13.1 -- /Users/anshikg/workspace/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-cloudwatch/aws-cloudwatch-alarm/.hypothesis/examples')
rootdir: /Users/anshikg/workspace/rpdk/aws-cloudformation-resource-providers-cloudwatch/aws-cloudwatch-alarm, inifile: /private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_nibs_l52.ini
collected 16 items                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [  6%]
handler_create.py::contract_invalid_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                               [ 12%]
handler_create.py::contract_create_duplicate PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             [ 18%]
handler_create.py::contract_create_read_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 25%]
handler_create.py::contract_create_list_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 31%]
handler_delete.py::contract_delete_read PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  [ 37%]
handler_delete.py::contract_delete_list PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  [ 43%]
handler_delete.py::contract_delete_update PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 50%]
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 56%]
handler_delete.py::contract_delete_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 62%]
handler_misc.py::contract_check_asserts_work PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             [ 68%]
handler_read.py::contract_read_without_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                            [ 75%]
handler_update.py::contract_update_read_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 81%]
handler_update.py::contract_update_list_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 87%]
handler_update_invalid.py::contract_update_create_only_property PASSED                                                                                                                                                                                                                                                                                                                                                                                                                          [ 93%]
handler_update_invalid.py::contract_update_non_existent_resource PASSED                                                                                                                                                                                                                                                                                                                                                                                                                         [100%]


```

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -41,6 +41,45 @@ def prune_properties(document, paths):
     return document
 
 
+def prune_null_properties_in_path(model, paths):
+    """Prune given properties from a document.
+
+    This assumes properties will always have an object (dict) as a parent.
+    The function modifies the document in-place, but also returns the document
+    for convenience. (The return value may be ignored.)
+    """
+    document = {"properties": model}
+    for path in paths:
+        try:
+            prop, resolved_path, parent = traverse(document, path)
+        except LookupError:
+            pass  # not found means nothing to delete
+        else:
+            if prop is None or prop == "":
+                key = resolved_path[-1]
+                del parent[key]
+    return document
+
+
+def prune_properties_from_model_in_path(model, pruned_model, paths):
+    """Prune given properties from a document.
+
+    This assumes properties will always have an object (dict) as a parent.
+    The function modifies the document in-place, but also returns the document
+    for convenience. (The return value may be ignored.)
+    """
+    document = {"properties": model}
+    pruned_document = {"properties": pruned_model}
+    for path in paths:
+        try:
+            _prop, _resolved_path, _parent = traverse(pruned_document, path)
+        except LookupError:
+            _prop, resolved_path, parent = traverse(document, path)
+            key = resolved_path[-1]
+            del parent[key]
+    return document
+
+
 def prune_properties_from_model(model, paths):
     """Prune given properties from a resource model.
 
@@ -299,7 +338,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         return {
             "desiredResourceState": desired_resource_state,
             "previousResourceState": previous_resource_state,
-            "logicalResourceIdentifier": None,
+            "logicalResourceIdentifier": "test",
             **kwargs,
             "region": region,
             "awsPartition": partition,
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -41,26 +41,6 @@ def prune_properties(document, paths):
     return document
 
 
-def prune_null_properties_in_path(model, paths):
-    """Prune given properties from a document.
-
-    This assumes properties will always have an object (dict) as a parent.
-    The function modifies the document in-place, but also returns the document
-    for convenience. (The return value may be ignored.)
-    """
-    document = {"properties": model.copy()}
-    for path in paths:
-        try:
-            prop, resolved_path, parent = traverse(document, path)
-        except LookupError:
-            pass  # not found means nothing to delete
-        else:
-            if not prop:
-                key = resolved_path[-1]
-                del parent[key]
-    return document["properties"]
-
-
 def prune_properties_from_model_in_path(output_model, input_model, paths):
     """Prune given properties from a document.
 
@@ -70,18 +50,13 @@ def prune_properties_from_model_in_path(output_model, input_model, paths):
     """
     output_document = {"properties": output_model.copy()}
     input_document = {"properties": input_model.copy()}
-    LOG.debug("This is output %s, input %s", output_document, input_document)
     for path in paths:
         try:
-            LOG.debug("This is path %s", path)
             _prop, _resolved_path, _parent = traverse(input_document, path)
         except LookupError:
-            LOG.debug("in exception")
             _prop, resolved_path, parent = traverse(output_document, path)
-            LOG.debug("Not exception")
             key = resolved_path[-1]
             del parent[key]
-    LOG.debug("final output: %s", output_document["properties"])
     return output_document["properties"]
 
 
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -41,23 +41,33 @@ def prune_properties(document, paths):
     return document
 
 
-def prune_properties_from_model_in_path(output_model, input_model, paths):
+def prune_properties_if_not_exist_in_path(output_model, input_model, paths):
     """Prune given properties from a document.
 
     This assumes properties will always have an object (dict) as a parent.
-    The function modifies the document in-place, but also returns the document
-    for convenience. (The return value may be ignored.)
+    The function returns the document after pruning the path which exists
+    in the paths tuple but not in the input_model
     """
     output_document = {"properties": output_model.copy()}
     input_document = {"properties": input_model.copy()}
     for path in paths:
         try:
-            _prop, _resolved_path, _parent = traverse(input_document, path)
+            if not path_exists(input_document, path):
+                _prop, resolved_path, parent = traverse(output_document, path)
+                key = resolved_path[-1]
+                del parent[key]
         except LookupError:
-            _prop, resolved_path, parent = traverse(output_document, path)
-            key = resolved_path[-1]
-            del parent[key]
+            pass
     return output_document["properties"]
+
+
+def path_exists(document, path):
+    try:
+        _prop, _resolved_path, _parent = traverse(document, path)
+    except LookupError:
+        return False
+    else:
+        return True
 
 
 def prune_properties_from_model(model, paths):
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -70,13 +70,18 @@ def prune_properties_from_model_in_path(output_model, input_model, paths):
     """
     output_document = {"properties": output_model.copy()}
     input_document = {"properties": input_model.copy()}
+    LOG.debug("This is output %s, input %s", output_document, input_document)
     for path in paths:
         try:
+            LOG.debug("This is path %s", path)
             _prop, _resolved_path, _parent = traverse(input_document, path)
         except LookupError:
+            LOG.debug("in exception")
             _prop, resolved_path, parent = traverse(output_document, path)
+            LOG.debug("Not exception")
             key = resolved_path[-1]
             del parent[key]
+    LOG.debug("final output: %s", output_document["properties"])
     return output_document["properties"]
 
 
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -48,36 +48,36 @@ def prune_null_properties_in_path(model, paths):
     The function modifies the document in-place, but also returns the document
     for convenience. (The return value may be ignored.)
     """
-    document = {"properties": model}
+    document = {"properties": model.copy()}
     for path in paths:
         try:
             prop, resolved_path, parent = traverse(document, path)
         except LookupError:
             pass  # not found means nothing to delete
         else:
-            if prop is None or prop == "":
+            if not prop:
                 key = resolved_path[-1]
                 del parent[key]
-    return document
+    return document["properties"]
 
 
-def prune_properties_from_model_in_path(model, pruned_model, paths):
+def prune_properties_from_model_in_path(output_model, input_model, paths):
     """Prune given properties from a document.
 
     This assumes properties will always have an object (dict) as a parent.
     The function modifies the document in-place, but also returns the document
     for convenience. (The return value may be ignored.)
     """
-    document = {"properties": model}
-    pruned_document = {"properties": pruned_model}
+    output_document = {"properties": output_model.copy()}
+    input_document = {"properties": input_model.copy()}
     for path in paths:
         try:
-            _prop, _resolved_path, _parent = traverse(pruned_document, path)
+            _prop, _resolved_path, _parent = traverse(input_document, path)
         except LookupError:
-            _prop, resolved_path, parent = traverse(document, path)
+            _prop, resolved_path, parent = traverse(output_document, path)
             key = resolved_path[-1]
             del parent[key]
-    return document
+    return output_document["properties"]
 
 
 def prune_properties_from_model(model, paths):
--- a/src/rpdk/core/contract/suite/handler_commons.py
+++ b/src/rpdk/core/contract/suite/handler_commons.py
@@ -1,7 +1,11 @@
 import logging
 
 from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
-from rpdk.core.contract.resource_client import prune_properties_from_model
+from rpdk.core.contract.resource_client import (
+    prune_null_properties_in_path,
+    prune_properties_from_model,
+    prune_properties_from_model_in_path,
+)
 from rpdk.core.contract.suite.contract_asserts import (
     failed_event,
     response_contains_primary_identifier,
@@ -128,10 +132,13 @@ def test_delete_failure_not_found(resource_client, current_resource_model):
 
 
 def test_input_equals_output(resource_client, input_model, output_model):
-    prune_properties_from_model(input_model, resource_client.create_only_paths)
+    prune_null_properties_in_path(input_model, resource_client.create_only_paths)
     prune_properties_from_model(input_model, resource_client.write_only_paths)
 
-    prune_properties_from_model(output_model, resource_client.read_only_paths)
-    prune_properties_from_model(output_model, resource_client.create_only_paths)
+    output_model_copy = output_model.copy()
+    prune_properties_from_model(output_model_copy, resource_client.read_only_paths)
+    prune_properties_from_model_in_path(
+        output_model_copy, input_model, resource_client.create_only_paths
+    )
 
-    assert input_model == output_model
+    assert input_model == output_model_copy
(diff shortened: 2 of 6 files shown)


src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -124,3 +125,13 @@ def test_delete_failure_not_found(resource_client, current_resource_model):
...
     )
     return error_code
+
+
+def test_input_equals_output(resource_client, input_model, output_model):
+    prune_properties_from_model(input_model, resource_client.create_only_paths)
+    prune_properties_from_model(input_model, resource_client.write_only_paths)
+
+    prune_properties_from_model(output_model, resource_client.read_only_paths)
+    prune_properties_from_model(output_model, resource_client.create_only_paths)
```
PatMyron, line 135: Why would [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) be pruned from both input and output models?

anshikg: because, create only property can be entered by the customer or generated by the handler. The test will intermittently depending on how this was enetered

PatMyron: > or generated by the handler

This should be covered by pruning [`readOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) already. [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) really shouldn't be [`readOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) anyway, but I suspect some [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) are being incorrectly marked as [`readOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) as well for [`GetAtt` support](https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/68#issuecomment-665243220)

PatMyron: @anshikg are those conditionally [`readOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) limited to [`primaryIdentifier`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics)? Would filtering [`primaryIdentifier`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) instead of [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) make sense?

anshikg: That is an interesting point. I have only seen createOnlyProperties which are primaryIdentifiers being generated. But I am not sure if this is a norm. @jotompki @rjlohan is that the case?

ammokhov: > This should be covered by pruning readOnlyProperties already. createOnlyProperties really shouldn't be readOnlyProperties anyway, but I suspect some createOnlyProperties are being incorrectly marked as readOnlyProperties as well for GetAtt support

there might be some primary identifiers that are not required and in this case create handler will have a method to generate them (as to my knowledge we do not generate any other types of create onlies). However, we should treat such primary id as provided by the customer (as we do it for them)

anshikg: sure. But, we wont have those ids in the input. Which will lead to input != output

ammokhov: No they should be just treated as normal properties and not as read only properties.

anshikg: As dicussed offline:
> I am not treating createOnly property as readOnly. The reason I pruned createOnly property is because, createOnly  property can be entered by the customer or by the handlers depending on how it was entered the contract test will pass or fail. Therefore, I took it out.

rjlohan: *If* `createOnlyProperties` are specified in the input, they should be equal on the output. Otherwise, we can ignore them as you've done.

anshikg: makes sense. I will add the check

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,45 @@ def prune_properties(document, paths):
...
+    document = {"properties": model}
+    for path in paths:
+        try:
+            prop, resolved_path, parent = traverse(document, path)
+        except LookupError:
+            pass  # not found means nothing to delete
+        else:
+            if prop is None or prop == "":
+                key = resolved_path[-1]
+                del parent[key]
```
ammokhov, line 60: I might be missing something but should this always be a value if no exception happened?
also it's removing keys from `parent` but we return a `document`.

ammokhov: also I think that modifying values in place isn't a good habit even though we been practicing it. it implies that we are intentionally causing side effects.

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,45 @@ def prune_properties(document, paths):
...
+    """
+    document = {"properties": model}
+    pruned_document = {"properties": pruned_model}
+    for path in paths:
+        try:
+            _prop, _resolved_path, _parent = traverse(pruned_document, path)
+        except LookupError:
+            _prop, resolved_path, parent = traverse(document, path)
+            key = resolved_path[-1]
+            del parent[key]
```
ammokhov, line 79: would you mind to provide more details of why it uses two models? model is a superset of pruned model, isnt it? if so, then I am confused why dont we just use model here

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,50 @@ def prune_properties(document, paths):
...
+
+
+def prune_properties_from_model_in_path(output_model, input_model, paths):
+    """Prune given properties from a document.
+
+    This assumes properties will always have an object (dict) as a parent.
+    The function modifies the document in-place, but also returns the document
+    for convenience. (The return value may be ignored.)
+    """
+    output_document = {"properties": output_model.copy()}
```
wbingli, line 71: If this method is in-place modify, why we are doing a `.copy()`?

tests/contract/test_resource_client.py
```diff
@@ -161,6 +163,54 @@ def test_prune_properties_from_model():
...
+        "spam": "eggs",
+        "one": "two",
+        "array": ["first", "second"],
+    }
+    model = prune_properties_from_model_in_path(
+        model,
+        previous_model,
+        [("properties", "foo"), ("properties", "spam"), ("properties", "array", "1")],
+    )
+    assert model == {
```
wbingli, line 207: Could we also assert previous model?

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -24,12 +25,13 @@
...
 @pytest.fixture(scope="module")
 def deleted_resource(resource_client):
-    request = model = resource_client.generate_create_example()
+    request = input_model = model = resource_client.generate_create_example()
     try:
         _status, response, _error = resource_client.call_and_assert(
             Action.CREATE, OperationStatus.SUCCESS, request
         )
         model = response["resourceModel"]
+        test_input_equals_output(resource_client, input_model, model)
```
wbingli, line 34: This is testing the create input and output, which it's covered in the create contract test.  It would be noise for the delete contract test. E.g. the delete contract should not fail because create input/output fails.

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,50 @@ def prune_properties(document, paths):
...
+    for path in paths:
+        try:
+            LOG.debug("This is path %s", path)
+            _prop, _resolved_path, _parent = traverse(input_document, path)
+        except LookupError:
+            LOG.debug("in exception")
+            _prop, resolved_path, parent = traverse(output_document, path)
+            LOG.debug("Not exception")
+            key = resolved_path[-1]
+            del parent[key]
```
wbingli, line 83: I still don't follow the logic. Why we need to `traverse` the output in the LookupError, and it only delete the output_document path?

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,50 @@ def prune_properties(document, paths):
...
+        except LookupError:
+            pass  # not found means nothing to delete
+        else:
+            if not prop:
+                key = resolved_path[-1]
+                del parent[key]
+    return document["properties"]
+
+
+def prune_properties_from_model_in_path(output_model, input_model, paths):
```
wbingli, line 64: Instead of doing an in-place update (which I don't think this method is doing so), it should directly return a pruned new model. During the contract test, those model will be used for following test, in-place update will impact the model which will be used for the rest of test.

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,50 @@ def prune_properties(document, paths):
...
+    """Prune given properties from a document.
+
+    This assumes properties will always have an object (dict) as a parent.
+    The function modifies the document in-place, but also returns the document
+    for convenience. (The return value may be ignored.)
+    """
+    output_document = {"properties": output_model.copy()}
+    input_document = {"properties": input_model.copy()}
+    LOG.debug("This is output %s, input %s", output_document, input_document)
```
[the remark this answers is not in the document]
anshikg: you are right I missed this from debugging

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,50 @@ def prune_properties(document, paths):
...
+    for path in paths:
+        try:
+            LOG.debug("This is path %s", path)
+            _prop, _resolved_path, _parent = traverse(input_document, path)
+        except LookupError:
+            LOG.debug("in exception")
```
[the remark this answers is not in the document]
anshikg: This was just for debugging. I am not sure if need to know every time we get and exception in this case.

src/rpdk/core/contract/resource_client.py
anshikg: may be I am missing something. This method is using a copy of the model. The passed in parameter (original model) should not be changed.

src/rpdk/core/contract/resource_client.py
anshikg: I am trying to move away from in place modify to avoid unintended changes. Also to make sure the methos is self contained.

src/rpdk/core/contract/resource_client.py
anshikg: This method is used to remove createOnly properties from the output. For example we call create handler with input (the input does not contain createOnly properties). The create handler gives us an output model (the create handler generated the createOnly property) Now, when we compare ```created_input == create_output``` the two models are going to be different. To avoid this issue, ```prune_null_properties_in_path``` prunes all createOnly properties from the input_model which are null/empty. Then, the ```prune_properties_from_model_in_path``` method removes all the createOnly properties from the output_model if they don't exist in the input_model (Lookup exception is encountered) This is done because the input_model would have been pruned in the previous method and because there can be multiple createOnly properties we will need to iterated through the list.

src/rpdk/core/contract/suite/handler_delete.py
anshikg: I wanted to ensure we had consistent behavior check across all contract test otherwise, the downstream teams usually ask why some creates are passing vs others.

tests/contract/test_resource_client.py
anshikg: makes sense. I will update this

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,25 @@ def prune_properties(document, paths):
     return document
 
 
+def prune_properties_from_model_in_path(output_model, input_model, paths):
```
wbingli, line 44: good to have a name like `prune_properties_if_not_exist_in_path`

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,25 @@ def prune_properties(document, paths):
...
+    """Prune given properties from a document.
+
+    This assumes properties will always have an object (dict) as a parent.
+    The function modifies the document in-place, but also returns the document
+    for convenience. (The return value may be ignored.)
```
wbingli, line 49: Better to update the comments. The document in-place still confuse. As the document here means the internal implementation, but the method doc should only talks about the method input and output.  Otherwise, it leads to consider it modify the input/output model in place.

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,25 @@ def prune_properties(document, paths):
...
+    """
+    output_document = {"properties": output_model.copy()}
+    input_document = {"properties": input_model.copy()}
+    for path in paths:
+        try:
+            _prop, _resolved_path, _parent = traverse(input_document, path)
```
wbingli, line 55: To make this better understanding, could we wrap the path check into a new method? e.g. 

```
def path_exists(input_doc, path) 
 try:
            _prop, _resolved_path, _parent = traverse(input_document, path)
 except LookupError:
    return False
return True

''' Use it in the prune
for path in paths
  if not path_exists(input_document, path)
    ''' Here we can also make a method to say delete_path
     _prop, resolved_path, parent = traverse(output_document, path)
     key = resolved_path[-1]
     del parent[key]
```

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,25 @@ def prune_properties(document, paths):
...
+        except LookupError:
+            _prop, resolved_path, parent = traverse(output_document, path)
```
wbingli, line 57: what if the path does not exist in the output_document?  It could happen if Create doesn't return the createOnlyProperties. But it will fail in an ugly way.   We should catch it and give proper error message.

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -124,3 +128,18 @@ def test_delete_failure_not_found(resource_client, current_resource_model):
         Action.DELETE, OperationStatus.FAILED, current_resource_model
     )
     return error_code
+
+
+def test_input_equals_output(resource_client, input_model, output_model):
```
wbingli, line 133: why we remove the prune_null method?

anshikg: As discussed offline, if a user enters null/"" for a property we should check that against the output. Therfore, we don't need the prune_null method

src/rpdk/core/contract/resource_client.py
anshikg: makes sense. I will add that check

src/rpdk/core/contract/resource_client.py
anshikg: sure, will do

src/rpdk/core/contract/suite/handler_commons.py
wbingli: OK, I totally missed that. But I think null and empty still valuable so keep them is right way.

src/rpdk/core/contract/resource_client.py
```diff
@@ -41,6 +41,35 @@ def prune_properties(document, paths):
...
+    output_document = {"properties": output_model.copy()}
+    input_document = {"properties": input_model.copy()}
+    for path in paths:
+        try:
+            if not path_exists(input_document, path):
+                _prop, resolved_path, parent = traverse(output_document, path)
+                key = resolved_path[-1]
+                del parent[key]
+        except LookupError:
+            pass
```
wbingli, line 60: Good to pass here, we can verify the input and output in the test_input_equals_output.

src/rpdk/core/contract/suite/handler_commons.py
PatMyron: Now that I think about it more, conditionally [`readOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) can apply to any type of property that has a default value, and we might have to ignore any property returned that was not in the input

(example: [`AWS::DynamoDB::Table.BillingMode`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-dynamodb-table.html#cfn-dynamodb-table-billingmode))

Issue #466, opened 2020-06-23
CloudFormationManagedUploadInfrastructure stack Company resource restrictions


AnthonyPoschen:
while trying to run `cfn submit` creating the CloudFormationManagedUploadInfrastructure stack fails as my companies enforced role boundaries require a specific PermissionBoundary and ManagedPolicy be present. If i manually create the stack it attempts to align it which would not be allowed in my circumstance.

the breakage for me occurs on the following resource where i need permissionBoundary set and a managed policy added.
https://github.com/aws-cloudformation/cloudformation-cli/blob/85dba80439e044f65e294af65d36fa0934151bad/src/rpdk/core/data/managed-upload-infrastructure.yaml#L99 

I have a local version of this repository working with the changes i needed and i am wondering if this kind of issue is something worth addressing, i am not sure what mechanism would be most appropriate to allow someone in my situation to control that stack and/or keep a modified version of the stack without forking this repo.

atheiman:
I need this as well, is there any recommended work around? I don't see a path forward to use the CFN CLI in any kind of regulated account other than to fork this project and change the cloudformation template? Any kind of tagging requirements in an SCP, or configuration requirements for S3 buckets prevent the use of the managed-upload-infrastructure-template.yml template in its current form.

I dont see a way to support every user's needs in the template, so I'd propose an argument to provide a custom cloudformation template and a note in documentation about how to customize the existing template.

```shell
cfn submit ... --custom-managed-upload-infrastructure-template ./my-custom-template.yml
```

PatMyron:
similar to https://github.com/aws-cloudformation/cloudformation-cli/issues/623

kylien65:
Anyone has a work around for this?  I tried adding the S3 and KMS actions to my local managed-upload-infrastructure.yaml template but that did not resolve the issue.  Getting API: iam:CreateRole User: arn:aws:sts::xxxxxxxxxxxx:assumed-role/DeveloperPowerUser/yyyyyyyyy is not authorized to perform: iam:CreateRole on resource: arn:aws:iam::xxxxxxxxxxxx:role/CloudFormationManagedUplo-LogAndMetricsDeliveryRol-1MMXUJ1M24H5Q with an explicit deny

atheiman:
[...]

PR #532, opened 2020-08-07
adding S3 and KMS IAM actions to LogAndMetricsDeliveryRole to see contract test results after cfn submit

PatMyron:
[KMS IAM actions](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_awskeymanagementservice.html#awskeymanagementservice-actions-as-permissions)
[S3 IAM actions](https://docs.aws.amazon.com/IAM/latest/UserGuide/list_amazons3.html#amazons3-actions-as-permissions)

["In order to see how the contract tests will run during registration, you can use cfn submit. You will need to add S3 and KMS permissions to your log and metrics delivery role. You can add the following policy as an inline policy to your logging role"](https://w.amazon.com/bin/view/AWS21/Design/Uluru/ContractTests/#HTestingcontracttestswithcfnsubmit)

### DIFF
--- a/src/rpdk/core/data/managed-upload-infrastructure.yaml
+++ b/src/rpdk/core/data/managed-upload-infrastructure.yaml
@@ -117,6 +117,11 @@ Resources:
           Statement:
           - Effect: Allow
             Action:
+            - kms:Encrypt
+            - kms:Decrypt
+            - kms:ReEncrypt*
+            - kms:GenerateDataKey*
+            - kms:DescribeKey
             - logs:CreateLogGroup
             - logs:CreateLogStream
             - logs:DescribeLogGroups
@@ -124,7 +129,10 @@ Resources:
             - logs:PutLogEvents
             - cloudwatch:ListMetrics
             - cloudwatch:PutMetricData
-            Resource: "*"
+            - s3:PutObject
+            Resource:
+            - !GetAtt EncryptionKey.Arn
+            - !GetAtt ArtifactBucket.Arn
 
 Outputs:
   CloudFormationManagedUploadBucketName:


src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -117,13 +117,19 @@ Resources:
...
+            - kms:GenerateDataKey*
+            - kms:DescribeKey
             - logs:CreateLogGroup
             - logs:CreateLogStream
             - logs:DescribeLogGroups
             - logs:DescribeLogStreams
             - logs:PutLogEvents
             - cloudwatch:ListMetrics
             - cloudwatch:PutMetricData
+            - s3:PutObject
```
LaikaN57, line 132: This seems particularly dangerous to leave wide open. Can we scope it to just the bucket above?

src/rpdk/core/data/managed-upload-infrastructure.yaml
```diff
@@ -117,13 +117,19 @@ Resources:
           Statement:
           - Effect: Allow
             Action:
+            - kms:Encrypt
+            - kms:Decrypt
```
LaikaN57, line 121: This seems particularly dangerous to leave wide open. Can we scope it to just the KMS key above?

wbingli: Good comment. We should have separate policy for KMS and S3, and scope down to the Key and bucket created above.

src/rpdk/core/data/managed-upload-infrastructure.yaml
PatMyron: https://github.com/aws-cloudformation/cloudformation-cli/issues/466

Issue #544, opened 2020-08-17
Contract tests not using role arn if specified for all client calls


ninavolkmuth:
Contract tests should use temporary credentials from the role arn for all client calls if a role is specified for the contract tests. This should be done when listing the cloudformation exports as well as when getting the account id.

Line references:
https://github.com/aws-cloudformation/cloudformation-cli/blob/033606a83cbd49387272f3ec2c8515459c8724b0/src/rpdk/core/boto_helpers.py#L82
https://github.com/aws-cloudformation/cloudformation-cli/blob/033606a83cbd49387272f3ec2c8515459c8724b0/src/rpdk/core/test.py#L62

advaj:
Im using CLI v0.1.11 and I have just run into this issue again. 

`Traceback (most recent call last):
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/rpdk/core/cli.py", line 98, in main
    args.command(args)
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/rpdk/core/test.py", line 193, in test
    args.role_arn,
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/rpdk/core/test.py", line 151, in get_inputs
    f.read(), region_name, endpoint_url, role_arn
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/rpdk/core/test.py", line 79, in render_jinja
    exports = get_cloudformation_exports(region_name, endpoint_url, role_arn)
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/rpdk/core/test.py", line 69, in get_cloudformation_exports
    for page in pages:
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/botocore/paginate.py", line 255, in __iter__
    response = self._make_request(current_kwargs)
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/botocore/paginate.py", line 332, in _make_request
    return self._method(**current_kwargs)
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/botocore/client.py", line 316, in _api_call
    return self._make_api_call(operation_name, kwargs)
  File "/root/.pyenv/versions/3.7.6/lib/python3.7/site-packages/botocore/client.py", line 626, in _make_api_call
    raise error_class(parsed_response, operation_name)
[...]

PR #548, opened 2020-08-21
Use role credentials when constructing clients for contract tests and use regional sts endpoint when getting account id

ninavolkmuth:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/544

*Description of changes:* This PR changes the contract tests so that they will use the role credentials if a role is passed in when constructing clients. This also introduces using the regional sts endpoint when constructing the sts client when we get the account id

My testing setup was to have one account's credentials exported locally but then pass in another account's role-arn when running cfn test. Tested with and without the role-arn to ensure the account id was correct and the overrides were picked up as expected (not found when using the local credentials and found when using the role-arn) from the stack exports.   

### DIFF
--- a/src/rpdk/core/boto_helpers.py
+++ b/src/rpdk/core/boto_helpers.py
@@ -46,14 +46,16 @@ def get_temporary_credentials(session, key_names=BOTO_CRED_KEYS, role_arn=None):
             response = sts_client.assume_role(
                 RoleArn=role_arn, RoleSessionName=session_name
             )
-        except ClientError as e:
+        except ClientError:
             # pylint: disable=W1201
             LOG.debug(
                 "Getting session token resulted in unknown ClientError. "
                 + "Could not assume specified role '%s'",
                 role_arn,
             )
-            raise DownstreamError("Could not assume specified role") from e
+            raise DownstreamError() from Exception(
+                "Could not assume specified role '{}'".format(role_arn)
+            )
         temp = response["Credentials"]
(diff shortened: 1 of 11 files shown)


src/rpdk/core/boto_helpers.py
```diff
@@ -78,7 +83,21 @@ def get_service_endpoint(service, region):
...
+    if temporary_credentials:
+        sts_client = session.client(
+            "sts",
+            endpoint_url=get_service_endpoint("sts", session.region_name),
+            region_name=session.region_name,
+            aws_access_key_id=temporary_credentials["accessKeyId"],
+            aws_secret_access_key=temporary_credentials["secretAccessKey"],
+            aws_session_token=temporary_credentials["sessionToken"],
+        )
+    else:
```
wbingli, line 96: Why we need to create a sts_client from default credential prodiver?    The `get_temporary_credentials` should cover all the cases.

ninavolkmuth: You are right; get_temporary_credentials should cover all the cases. I was incorrectly thinking that the temporary credentials would only be used when a role is specified (which is not the case). Will change

ninavolkmuth: You are right; get_temporary_credentials should cover all the cases. I was incorrectly thinking that the temporary credentials would only be used when a role is specified (which is not the case). Will change

Issue #478, opened 2020-07-07
Write Only property fails to assert on nested properties


anshikg:
In case of nested write only properties: https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-elasticloadbalancingv2/blob/master/aws-elasticloadbalancingv2-listener.json#L325

The contract test fails to assert: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_commons.py#L15

The test fails with the following stack trace:  
```
resource_model = {'AlpnPolicy': [], 'Certificates': [], 'DefaultActions': [{'ForwardConfig': {'TargetGroups': [{'TargetGroupArn': 'arn:...'arn:aws:elasticloadbalancing:us-east-1:362761549384:listener/net/MCM-29093805/37e4ea061f7d3f55/ee53b3bd90797dba', ...}
write_only_property = ('properties', 'DefaultActions', 'AuthenticateOidcConfig', 'ClientSecret')

    @staticmethod
    def key_error_safe_traverse(resource_model, write_only_property):
        try:
            return traverse(
>               resource_model, fragment_list(write_only_property, "properties")
            )[0]

../cloudformation-cli/cloudformation-cli/src/rpdk/core/contract/resource_client.py:235: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

document = [{'ForwardConfig': {'TargetGroups': [{'TargetGroupArn': 'arn:aws:elasticloadbalancing:us-east-1:362761549384:targetgro...tGroupArn': 'arn:aws:elasticloadbalancing:us-east-1:362761549384:targetgroup/CFN/c78de428819d902c', 'Type': 'forward'}]
path_parts = ['DefaultActions', 'AuthenticateOidcConfig', 'ClientSecret']

    def traverse(document, path_parts):
        """Traverse the document according to the reference.
    
        Since the document is presumed to be the reference's base, the base is
        discarded. There is no validation that the reference is valid.
    
        :raises ValueError, LookupError: the reference is invalid for this document
    
        >>> traverse({"foo": {"bar": [42]}}, tuple())
        ({'foo': {'bar': [42]}}, (), None)
        >>> traverse({"foo": {"bar": [42]}}, ["foo"])
[...]

PR #551, opened 2020-08-25
Support embeded list properties

anshikg:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/478

*Description of changes:*
Currently, if any property type like writeOnly/readOnly/createOnly/readOnly property have an embedded list we will get a KeyError in traverse method. For example, lets say this is a writeOnlyProperty: ```/properties/DefaultActions/AuthenticateOidcConfig/ClientSecret``` where ```DefaultActions``` is a list of objects. In contract test we want to ensure no writeOnly properties are being leaked to make sure, we traverse all writeOnly properties in the schema and check if the object does not exist. Right now, if a list exists in the schema we can look at a specific index of the list.Using the above example, contract test can support this property as an index is mentioned ```/properties/DefaultActions/0/AuthenticateOidcConfig/ClientSecret``` but, what happens when we want all the values under the list ```/properties/DefaultActions/*/AuthenticateOidcConfig/ClientSecret``` we will get a keyError. So, in order to get all the values under a list I added the following change which will recurse through the path provided as a parameter and if a list exists it will recurse through all the required indices (single index is asked and if * is present all the indices)

*Test*
* unit test

### DIFF
--- a/src/rpdk/core/jsonutils/utils.py
+++ b/src/rpdk/core/jsonutils/utils.py
@@ -103,17 +103,9 @@ def traverse(document, path_parts):
     return document, tuple(path), parent
 
 
-# pylint: disable=R0913,C0301
-# flake8: noqa=C901,B950
-def traverse_list_safe(
-    document,
-    path_parts,
-    parent=None,
-    path=None,
-    paths=None,
-    documents=None,
-    parents=None,
-):
+# pylint: disable=C0301
+# flake8: noqa=B950
+def traverse_list_safe(document, path_parts):
     """Traverse the document according to the reference.
     This method traverses all the elements of the lists
     and returns their corresponding document, path, parent
@@ -124,7 +116,7 @@ def traverse_list_safe(
     :raises ValueError, LookupError: the reference is invalid for this document
 
     >>> traverse_list_safe({"foo": {"bar": [42]}}, tuple())
-    ([{'foo': {'bar': [42]}}], [None], [None])
+    ([{'foo': {'bar': [42]}}], [[]], [None])
     >>> traverse_list_safe({"foo": {"bar": [42]}}, ["foo"])
     ([{'bar': [42]}], [['foo']], [{'foo': {'bar': [42]}}])
     >>> traverse_list_safe({"foo": {"bar": [42]}}, ("foo", "bar"))
@@ -132,7 +124,7 @@ def traverse_list_safe(
     >>> traverse_list_safe({"foo": {"bar": [42]}}, ("foo", "bar", "0"))
     ([42], [['foo', 'bar', 0]], [[42]])
     >>> traverse_list_safe({"foo": {"bar": [42, 21, 32]}}, ("foo", "bar", "*"))
-    ([42, 21, 32], [['foo', 'bar', 0], ['foo', 'bar', 1], ['foo', 'bar', 2]], [[42, 21, 32], [42, 21, 32], [42, 21, 32]])
+    ([[42], [21], [32]], [['foo', 'bar', 0], ['foo', 'bar', 1], ['foo', 'bar', 2]], [[42], [21], [32]])
     >>> traverse_list_safe({}, ["foo"])
     Traceback (most recent call last):
     ...
@@ -146,52 +138,47 @@ def traverse_list_safe(
     ...
     IndexError: list index out of range
     """
-    if not path_parts:
-        if not paths:
-            paths = []
-        paths.append(path)
+    parent = None
+    path = []
+    documents = []
+    parents = []
+    paths = []
+    for part in path_parts:
+        if isinstance(document, Sequence):
+            if part == "*":
+                return interate_over_list(document, path, path_parts)
+            part = int(part)
+        parent = document
+        try:
+            document = document[part]
+            path.append(part)
+        except TypeError:
+            pass
+    documents.append(document)
+    paths.append(path)
+    parents.append(parent)
+    return documents, paths, parents
 
-        if not documents:
-            documents = []
-        documents.append(document)
 
-        if not parents:
-            parents = []
-        parents.append(parent)
-
-        return documents, paths, parents
-
-    temp_path = path_parts[0]
-    if isinstance(document, Sequence):
-        if temp_path == "*":
-            index = 0
-            path_copy = path.copy()
-            while True:
-                temp_path = index
-                path = path_copy.copy()
-                try:
-                    path.append(temp_path)
-                    documents, paths, parents = traverse_list_safe(
-                        document[temp_path],
-                        path_parts[1:],
-                        document,
-                        path,
-                        paths,
-                        documents,
-                        parents,
-                    )
-                    index += 1
-                except LookupError:
-                    break
-            return documents, paths, parents
-        temp_path = int(temp_path)
-
-    if not path:
-        path = []
-    path.append(temp_path)
-    return traverse_list_safe(
-        document[temp_path], path_parts[1:], document, path, paths, documents, parents
-    )
+def interate_over_list(document, path, path_parts):
+    documents = []
+    parents = []
+    paths = []
+    index = 0
+    while True:
+        path_copy = path.copy()
+        path_copy.append(index)
+        try:
+            _document, _path, _parent = traverse_list_safe(
+                document[index], path_parts[1:]
+            )
+        except IndexError:
+            break
+        documents.append(_document)
+        paths.append(path_copy)
+        parents.append(_parent)
+        index += 1
+    return documents, paths, parents
 
 
 def schema_merge(target, src, path):  # noqa: C901
--- a/src/rpdk/core/jsonutils/utils.py
+++ b/src/rpdk/core/jsonutils/utils.py
@@ -103,6 +103,97 @@ def traverse(document, path_parts):
     return document, tuple(path), parent
 
 
+# pylint: disable=R0913,C0301
+# flake8: noqa=C901,B950
+def traverse_list_safe(
+    document,
+    path_parts,
+    parent=None,
+    path=None,
+    paths=None,
+    documents=None,
+    parents=None,
+):
+    """Traverse the document according to the reference.
+    This method traverses all the elements of the lists
+    and returns their corresponding document, path, parent
+
+    Since the document is presumed to be the reference's base, the base is
+    discarded. There is no validation that the reference is valid.
+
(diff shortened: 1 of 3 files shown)


src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,95 @@ def traverse(document, path_parts):
...
+def traverse_list_safe(
+    document,
+    path_parts,
+    parent=None,
+    path=None,
+    paths=None,
+    documents=None,
+    parents=None,
+):
+    """Traverse the document according to the reference.
```
jotompki, line 117: How does this differ from traverse? this comment makes it seem like its the same

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,95 @@ def traverse(document, path_parts):
     return document, tuple(path), parent
 
 
+# pylint: disable=R0913,W0102,C0301
```
jotompki, line 106: please try to leave some comments on why these are disabled. why so many arguments? can this be refactored to avoid all the args

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,95 @@ def traverse(document, path_parts):
...
+
+        if not parents:
+            parents = []
+        parents.append(parent)
+
+        return documents, paths, parents
+
+    temp_path = path_parts[0]
+    if isinstance(document, Sequence):
+        if temp_path == "*":
```
jotompki, line 164: what exactly is this wildcard format for?

src/rpdk/core/jsonutils/utils.py
anshikg: forgot to update the comments. I will update them in the next iteration. Essentially, this method allows you to traverse through a nested list. Lets say for example you have a resource which has writeOnlyProperties which are list of strings in an object. To traverse the list of writeOnlyProperty list you could provide * to denote all the indices of the list and traverse each of them

src/rpdk/core/jsonutils/utils.py
anshikg: explained above

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,97 @@ def traverse(document, path_parts):
     return document, tuple(path), parent
 
 
+# pylint: disable=R0913,C0301
+# flake8: noqa=C901,B950
```
anshikg, line 107: * ```C901 is too complex``` In this case, I need to create a recursive loop to get all the list properties. I don't know of a easier way of solving this. 
* ```B950 line too long (121 > 80 characters)``` In this case, the line being referred to is part of the tests being run for this method.
* ```R0913 too many arguments``` Switching to recursive method in place of the iterative approach required the addition of the parameters.
* ```C0301  Line too long``` In this case, the line being referred to is part of the tests being run for this method.

wbingli:
I think we need more details on how to support writeOnly properties.  None of this PR description nor #478 give enough information how exactly we expect customer use specify writeOnly for nested property. 

It seems the solution here is to support `/properties/DefaultActions/*/AuthenticateOidcConfig/ClientSecret`.
I think there are other 2 options:
 1. Support definitions: `/definitions/AuthenticateOidcConfig/ClientSecret`
 2. Support writeOnly as property attribute:

```
        "ClientSecret": {
          "type": "string",
          "writeOnly": true
        }
```

We should discuss those options, more on how customer will consume this feature.

anshikg:
> I think we need more details on how to support writeOnly properties. None of this PR description nor #478 give enough information how exactly we expect customer use specify writeOnly for nested property.
> 
> It seems the solution here is to support `/properties/DefaultActions/*/AuthenticateOidcConfig/ClientSecret`.
> I think there are other 2 options:
> 
> 1. Support definitions: `/definitions/AuthenticateOidcConfig/ClientSecret`
> 2. Support writeOnly as property attribute:
> 
> ```
>         "ClientSecret": {
>           "type": "string",
>           "writeOnly": true
>         }
> ```
> 
> We should discuss those options, more on how customer will consume this feature.

Let me try and explain what I am trying to do in this CR in the details section. I just want to call out, that writeOnly properties is already supported. I am trying to supported embeded list.
Lets say for example you have a writeOnly property:  ```properties/DefaultActions/*/AuthenticateOidcConfig/ClientSecret ``` where, default actions is a list of objects. Right now, we don't support a complete list but support a specific index of a list. This CR lets us get all the indices of a list by defining an *

anshikg:
> not sure why do we need additional method for just this case;
> 
> if `writeOnlyProperties` are something like this - `'writeOnlyProperties': [ '/properties/DefaultActions/AuthenticateOidcConfig/ClientSecret' ],`
> 
> this could be simply solved using flattened schema: `JsonSchemaFlattener(self.schema).flatten_schema()` =>
> 
> schema becomes
> 
> ```
> {
>     (): ...,
>     ('definitions','Action'): {...},
>     ('definitions', 'FixedResponseConfig'): {...},
>     ('definitions', 'AuthenticateCognitoConfig'): {...},
>     ...
>     ('definitions', 'AuthenticateOidcConfig'): {
>         'properties': {
>             ...
>             'ClientSecret': {...},
>         }
>     },
>     ('definitions', 'Certificate'): {...}
> }
> ```
> 
> `ClientSecret` could be taken directly from `('definitions', 'AuthenticateOidcConfig')` which is on the root level;

That is an interesting, would this get all the indices of the list too? For example, if DefaultActions is a list and we need to get all the values of the list will that work?

ammokhov:
> > not sure why do we need additional method for just this case;
> > if `writeOnlyProperties` are something like this - `'writeOnlyProperties': [ '/properties/DefaultActions/AuthenticateOidcConfig/ClientSecret' ],`
> > this could be simply solved using flattened schema: `JsonSchemaFlattener(self.schema).flatten_schema()` =>
> > schema becomes
> > ```
> > {
> >     (): ...,
> >     ('definitions','Action'): {...},
> >     ('definitions', 'FixedResponseConfig'): {...},
> >     ('definitions', 'AuthenticateCognitoConfig'): {...},
> >     ...
> >     ('definitions', 'AuthenticateOidcConfig'): {
> >         'properties': {
> >             ...
> >             'ClientSecret': {...},
> >         }
> >     },
> >     ('definitions', 'Certificate'): {...}
> > }
> > ```
> > 
> > 
> > `ClientSecret` could be taken directly from `('definitions', 'AuthenticateOidcConfig')` which is on the root level;
> 
> That is an interesting, would this get all the indices of the list too? For example, if DefaultActions is a list and we need to get all the values of the list will that work?

otherwise it wont be possible to flatten schema

src/rpdk/core/jsonutils/utils.py
aygold92: what is the `safe` distinction here? why do we need a "safe" method, thus presumably our standard method which is "unsafe"?

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,97 @@ def traverse(document, path_parts):
...
+def traverse_list_safe(
+    document,
+    path_parts,
+    parent=None,
+    path=None,
+    paths=None,
+    documents=None,
+    parents=None,
```
aygold92, line 115: I don't understand what all these arguments are or why they are needed.  In general, less arguments is better.  like document/documents, parent/parents, path/paths/path_parts... can these all be combined? if you need say that parent = parents[0] or whatever in the method thats better than passing all as parameters.

Also, why are you setting the default value to None if you're then just checking for None and setting to empty array?

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,97 @@ def traverse(document, path_parts):
...
+                        document[temp_path],
+                        path_parts[1:],
+                        document,
+                        path,
+                        paths,
+                        documents,
+                        parents,
+                    )
+                    index += 1
+                except LookupError:
```
aygold92, line 184: This try/except is too broad... what is this LookupError even from? 

I'm guessing its supposed to be from the document[temp_path] call... it looks like you're having to engineer complex just to iterate over this document, why not just do a foreach loop?

src/rpdk/core/jsonutils/utils.py
aygold92: I personally like the wildcard format as a way to nest into lists 👍 ... have you looked up if there is already a standard way to do this though?

src/rpdk/core/jsonutils/utils.py
```diff
@@ -103,6 +103,97 @@ def traverse(document, path_parts):
...
+        return documents, paths, parents
+
+    temp_path = path_parts[0]
+    if isinstance(document, Sequence):
+        if temp_path == "*":
+            index = 0
+            path_copy = path.copy()
+            while True:
+                temp_path = index
+                path = path_copy.copy()
```
aygold92, line 171: that doesn't sound right... whats the point of making the copy if we're just going to copy it again? i think if you switch to for loop you won't need to do this

src/rpdk/core/jsonutils/utils.py
anshikg: All this means is that this method should be able to iterate all the elements of the list

src/rpdk/core/jsonutils/utils.py
anshikg: The reason why these were added are because we are using a recursive approach. The singular variables like parent/path/document denote the inputs we are getting in this method while, the plural variables like documents/parents/paths denotes the list of outputs from this method. 
Pytest doe snot thing [] as default is a safe value to assign and expects None to be assigned. This also ensures the parameter list is empty every time this method is called.

src/rpdk/core/jsonutils/utils.py
anshikg: Adding an iterative approach to iterate over all the elements of the list will cause repetitive code. Therefore, I implemented a recursive one.

src/rpdk/core/jsonutils/utils.py
anshikg: that is a good catch. Will fix the extra copy.

src/rpdk/core/jsonutils/utils.py
aygold92: I think I agree with `C901 is too complex` and `R0913 too many arguments`

ddarbari (approved):

PR #553, opened 2020-08-26
adjusted summary report

ammokhov:
*Issue #, if available:*

*Description of changes:*

added messages to all assertions so it will be clear why certain test failed;
removed the full stack trace to a `short` option as it outputs the assertion message and method it comes from; 

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -385,10 +385,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                 )[0]
                 for primary_identifier in primary_identifier_paths
             ), "Every returned model MUST include the primaryIdentifier"
-        except KeyError:
+        except KeyError as e:
             raise AssertionError(
                 "Every returned model MUST include the primaryIdentifier"
-            )
+            ) from e
 
     @staticmethod
     def is_primary_identifier_equal(
@@ -404,11 +404,11 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                 )[0]
                 for primary_identifier in primary_identifier_path
             )
-        except KeyError:
+        except KeyError as e:
             raise AssertionError(
                 "The primaryIdentifier returned in every progress event must\
                      match the primaryIdentifier passed into the request"
-            )
+            ) from e
 
     def _make_payload(self, action, current_model, previous_model=None, **kwargs):
         return self.make_request(
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -368,9 +368,11 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                     resource_model, fragment_list(primary_identifier, "properties")
                 )[0]
                 for primary_identifier in primary_identifier_paths
-            ), "Every model MUST include the primaryIdentifier"
+            ), "Every returned model MUST include the primaryIdentifier"
         except KeyError:
-            raise AssertionError("Every model MUST include the primaryIdentifier")
+            raise AssertionError(
+                "Every returned model MUST include the primaryIdentifier"
+            )
 
     @staticmethod
     def is_primary_identifier_equal(
@@ -387,7 +389,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                 for primary_identifier in primary_identifier_path
             )
         except KeyError:
-            raise AssertionError("Every model MUST include the primaryIdentifier")
+            raise AssertionError(
+                "The primaryIdentifier returned in every progress event must\
+                     match the primaryIdentifier passed into the request"
+            )
 
     def _make_payload(self, action, current_model, previous_model=None, **kwargs):
         return self.make_request(
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -188,7 +188,8 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             assert not any(
                 self.key_error_safe_traverse(resource_model, write_only_property)
                 for write_only_property in self.write_only_paths
-            )
+            ), "The model MUST NOT return properties defined as \
+                writeOnlyProperties in the resource schema"
 
     @property
     def strategy(self):
@@ -361,22 +362,32 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
 
     @staticmethod
     def assert_primary_identifier(primary_identifier_paths, resource_model):
-        assert all(
-            traverse(resource_model, fragment_list(primary_identifier, "properties"))[0]
-            for primary_identifier in primary_identifier_paths
-        )
+        try:
+            assert all(
+                traverse(
+                    resource_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                for primary_identifier in primary_identifier_paths
+            ), "Every model MUST include the primaryIdentifier"
+        except KeyError:
+            raise AssertionError("Every model MUST include the primaryIdentifier")
 
     @staticmethod
     def is_primary_identifier_equal(
         primary_identifier_path, created_model, updated_model
     ):
-        return all(
-            traverse(created_model, fragment_list(primary_identifier, "properties"))[0]
-            == traverse(updated_model, fragment_list(primary_identifier, "properties"))[
-                0
-            ]
-            for primary_identifier in primary_identifier_path
-        )
+        try:
+            return all(
+                traverse(
+                    created_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                == traverse(
+                    updated_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                for primary_identifier in primary_identifier_path
+            )
+        except KeyError:
+            raise AssertionError("Every model MUST include the primaryIdentifier")
 
     def _make_payload(self, action, current_model, previous_model=None, **kwargs):
         return self.make_request(
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -1,6 +1,5 @@
 from functools import wraps
 from inspect import Parameter, signature
-from unittest.test.testmock.support import is_instance
 
 import pytest
 
@@ -82,7 +81,10 @@ def response_does_not_contain_write_only_properties(resource_client, response):
 def response_contains_resource_model_equal_current_model(
     response, current_resource_model
 ):
-    assert response["resourceModel"] == current_resource_model
+    assert (
+        response["resourceModel"] == current_resource_model
+    ), "handler MUST return a model representation that conforms\
+         to the shape of the resource schema"
 
 
 @decorate()
@@ -92,7 +94,9 @@ def response_contains_resource_model_equal_updated_model(
     assert response["resourceModel"] == {
         **current_resource_model,
         **update_resource_model,
-    }
+    }, "All properties specified in the update request MUST be present in the \
+        model returned, and they MUST match exactly, with the exception of \
+            properties defined as writeOnlyProperties in the resource schema"
 
 
 @decorate()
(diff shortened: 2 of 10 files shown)


src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -122,7 +133,11 @@ def test_delete_success(resource_client, current_resource_model):
     return response
 
 
-@failed_event(error_code=HandlerErrorCode.NotFound)
+@failed_event(
+    error_code=HandlerErrorCode.NotFound,
+    msg="A delete hander MUST return FAILED with a NotFound error code\
+         if the if the resource did not exist prior to the delete request",
```
ninavolkmuth, line 139: Typo extra "if the"

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -35,15 +35,20 @@ def deleted_resource(resource_client):
         _status, response, _error = resource_client.call_and_assert(
             Action.DELETE, OperationStatus.SUCCESS, model
         )
-        assert "resourceModel" not in response
+        assert (
+            "resourceModel" not in response
+        ), "ProgressEvent object MUST NOT contain a model"
```
ninavolkmuth, line 40: Maybe say instead: "The deletion handler's response must not contain a model"

Just to keep it more in line with the other messages since we don't usually refer to the response as the ProgressEvent

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -63,7 +68,10 @@ def contract_delete_list(resource_client, deleted_resource):
     #       remove the model from the list, however.
 
     deleted_model, _request = deleted_resource
-    assert not test_model_in_list(resource_client, deleted_model)
+    assert not test_model_in_list(
+        resource_client, deleted_model
+    ), "any subsequent list operation MUST NOT return the primaryIdentifier\
+         associated with the deleted resource instance"
```
ninavolkmuth, line 74: Maybe rephrase to something more similar to above like: "A list operation MUST NOT return the primaryIdentifier of any deleted resource instance"

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -92,7 +100,9 @@ def contract_delete_create(resource_client, deleted_resource):
             response["resourceModel"], resource_client.read_only_paths
         )
 
-        assert deleted_model == response["resourceModel"]
+        assert (
+            deleted_model == response["resourceModel"]
+        ), "A create handler MUST be idempotent"
```
ninavolkmuth, line 105: Idempotent isn't the right word here (idempotent is used in the context of multiple of the same type of calls not for two separate api calls). This is for the line in the contract that states: "Once a delete operation successfully completes, a subsequent create request with the same primaryIdentifier or additionalIdentifiers MUST NOT return FAILED with an AlreadyExists error code"

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -57,5 +58,8 @@ def contract_update_list_success(updated_resource, resource_client):
     _create_request, _created_model, _update_request, updated_model = updated_resource
     assert resource_client.is_primary_identifier_equal(
         resource_client.primary_identifier_paths, _created_model, updated_model
-    )
-    assert test_model_in_list(resource_client, updated_model)
+    ), "The primaryIdentifier returned in every progress event must match\
+         the primaryIdentifier passed into the request"
+    assert test_model_in_list(
+        resource_client, updated_model
+    ), "A list handler MUST always return a status of SUCCESS"
```
ninavolkmuth, line 65: This is checking that the updated model is returned by list so the assertion message should mention that. We don't require the list handler always return "SUCCESS". Just that it doesn't return IN_PROGRESS

src/rpdk/core/contract/suite/handler_update_invalid.py
```diff
@@ -26,7 +32,9 @@ def contract_update_create_only_property(resource_client):
             _status, response, _error_code = resource_client.call_and_assert(
                 Action.UPDATE, OperationStatus.FAILED, update_request, created_model
             )
-            assert response["message"]
+            assert response[
+                "message"
+            ], "The progress event MUST return an error message"
```
ninavolkmuth, line 37: Might want to clarify and add: "When the status if failed"

src/rpdk/core/contract/suite/handler_update_invalid.py
```diff
@@ -38,13 +46,14 @@ def contract_update_create_only_property(resource_client):
...
+         if the resource did not exist prior to the update request",
 )
 def contract_update_non_existent_resource(resource_client):
     create_request = resource_client.generate_invalid_create_example()
     update_request = resource_client.generate_update_example(create_request)
     _status, response, _error = resource_client.call_and_assert(
         Action.UPDATE, OperationStatus.FAILED, update_request, create_request
     )
-    assert response["message"]
+    assert response["message"], "The progress event MUST return an error message"
```
ninavolkmuth, line 58: Might want to clarify and add: "When the status if failed"

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -106,11 +110,12 @@ def response_contains_primary_identifier(resource_client, response):
 def response_contains_unchanged_primary_identifier(
     resource_client, response, current_resource_model
 ):
-    resource_client.is_primary_identifier_equal(
+    assert resource_client.is_primary_identifier_equal(
```
anshikg, line 113: is_primary_identifier_equal is already asserting do we need to assert here too?

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -45,7 +45,8 @@ def contract_update_read_success(updated_resource, resource_client):
     _create_request, _created_model, _update_request, updated_model = updated_resource
     assert resource_client.is_primary_identifier_equal(
         resource_client.primary_identifier_paths, _created_model, updated_model
-    )
+    ), "The primaryIdentifier returned in every progress event must match\
```
anshikg, line 48: this is not being checked for every progress evert but only when a handler returns failed/success. We might ant to rephrase this to reflect that

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -57,5 +58,8 @@ def contract_update_list_success(updated_resource, resource_client):
     _create_request, _created_model, _update_request, updated_model = updated_resource
     assert resource_client.is_primary_identifier_equal(
         resource_client.primary_identifier_paths, _created_model, updated_model
-    )
-    assert test_model_in_list(resource_client, updated_model)
+    ), "The primaryIdentifier returned in every progress event must match\
```
anshikg, line 61: same as above

src/rpdk/core/contract/suite/handler_update.py
anshikg: I would leave the success statement out. It does not require to be added here

src/rpdk/core/data/pytest-contract.ini
```diff
@@ -1,8 +1,11 @@
 [pytest]
+log_cli = True
```
anshikg, line 2: can we add a parameter to allow people to get a more verbose error message if they want

src/rpdk/core/contract/resource_client.py
```diff
@@ -361,22 +362,32 @@ def assert_time(self, start_time, end_time, action):
...
-        )
+        try:
+            assert all(
+                traverse(
+                    resource_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                for primary_identifier in primary_identifier_paths
+            ), "Every model MUST include the primaryIdentifier"
+        except KeyError:
+            raise AssertionError("Every model MUST include the primaryIdentifier")
```
anshikg, line 373: can we throw key error instead with the same error message

src/rpdk/core/contract/resource_client.py
```diff
@@ -361,22 +362,32 @@ def assert_time(self, start_time, end_time, action):
...
+                traverse(
+                    created_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                == traverse(
+                    updated_model, fragment_list(primary_identifier, "properties")
+                )[0]
+                for primary_identifier in primary_identifier_path
+            )
+        except KeyError:
+            raise AssertionError("Every model MUST include the primaryIdentifier")
```
anshikg, line 390: can we throw KeyError instead of assertion error?

src/rpdk/core/contract/resource_client.py
ammokhov: this assertion is aimed to make sure that primary identifier is returned within the `ResourceModel` object but in case it's missing or `null` we get `KeyError` instead of a failed assertion, which itself is extremely misleading. 

leaving `KeyError` will mean that the assertion message will never get thrown, bc it will be either a successful assertion (just if the key exists) or key error; 

I feel like `cli` with a testing component in it should not expose any unhandled runtime errors and instead surface what's causing incorrect behavior;

src/rpdk/core/contract/resource_client.py
ammokhov: same as above, I dont think we should expose runtime errors and rather handle them accordingly (ideally have custom exception classes for each case like that)

src/rpdk/core/contract/suite/contract_asserts.py
ammokhov: [is_primary_identifier_equal](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L389) has expected `boolean` output; it does not assert

src/rpdk/core/contract/suite/handler_update.py
ammokhov: oh, good catch

src/rpdk/core/data/pytest-contract.ini
ammokhov: `--verbose` parameter is there. if you mean traceback printing, it's set to short which outputs only assertion messages and associated methods; I am not sure what do we get from printing out full traceback but might be a good idea to give an option;

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -85,7 +84,11 @@ def response_does_not_contain_write_only_properties(resource_client, response):
 def response_contains_resource_model_equal_current_model(
     response, current_resource_model
 ):
-    assert response["resourceModel"] == current_resource_model
+    assert (
+        response["resourceModel"] == current_resource_model
+    ), "All properties specified in the request MUST be present in the model \
+        returned, and they MUST match exactly, with the exception of properties\
+             defined as writeOnlyProperties in the resource schema"
```
PatMyron, line 91: think this error message might be misleading because there should be more exceptions:
[readOnlyProperties, writeOnlyProperties, defaults, insertionOrder, uniqueItems, etc.](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics)

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -95,7 +98,9 @@ def response_contains_resource_model_equal_updated_model(
     assert response["resourceModel"] == {
         **current_resource_model,
         **update_resource_model,
-    }
+    }, "All properties specified in the update request MUST be present in the \
+        model returned, and they MUST match exactly, with the exception of \
+            properties defined as writeOnlyProperties in the resource schema"
```
PatMyron, line 103: think this error message might be misleading because there should be more exceptions:
[readOnlyProperties, writeOnlyProperties, defaults, insertionOrder, uniqueItems, etc.](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics)

src/rpdk/core/contract/suite/contract_asserts.py
ammokhov: this been taken from the handler contract agreement

PR #554, opened 2020-08-26
fix(test) - increasing default handler timeout for cfn test

PatMyron:
https://github.com/aws-cloudformation/cloudformation-cli/issues/481, https://github.com/aws-cloudformation/cloudformation-cli/pull/486

[SAM testing](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test.html#resource-type-develop-test) results in:
```bash
AssertionError: Handler <Action.READ: 'READ'> timed out.
AssertionError: Handler <Action.LIST: 'LIST'> timed out.
```

[diff without `black` formatting changes](https://github.com/aws-cloudformation/cloudformation-cli/pull/554/files/1214f6db8aa5df839a29ffebcd444ec06377e9b6)

PR #555, opened 2020-08-27
Add clientToken to contract test

anshikg:
*Issue #, if available:*

*Description of changes:*
* Client Token is required by some handlers to complete a handler execution. Without the bearerToken (clientToken) handlers throw NPE. Therefore adding clientToken to the request payload to the handlers.
* Pylint added some linting changes too

### DIFF
--- a/src/rpdk/core/contract/suite/handler_commons.py
+++ b/src/rpdk/core/contract/suite/handler_commons.py
@@ -50,7 +50,8 @@ def test_create_failure_if_repeat_writeable_id(resource_client, current_resource
 @response_contains_resource_model_equal_current_model
 def test_read_success(resource_client, current_resource_model):
     primay_identifier_only_model = create_model_with_properties_in_path(
-        current_resource_model.copy(), resource_client.primary_identifier_paths,
+        current_resource_model.copy(),
+        resource_client.primary_identifier_paths,
     )
     _status, response, _error_code = resource_client.call_and_assert(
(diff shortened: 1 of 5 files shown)


src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -50,8 +50,7 @@ def test_create_failure_if_repeat_writeable_id(resource_client, current_resource
 @response_contains_resource_model_equal_current_model
 def test_read_success(resource_client, current_resource_model):
     primay_identifier_only_model = create_model_with_properties_in_path(
-        current_resource_model.copy(),
-        resource_client.primary_identifier_paths,
+        current_resource_model.copy(), resource_client.primary_identifier_paths,
```
ninavolkmuth, line 53: You need to update your black python module and may have to run the files individually. When the build was failing for me, these files weren't getting changed when I ran pre-commit for all files and needed to specifically run black on these specific files (though not sure if you'll have that same issue). It's definitely annoying that these linter/formatting will randomly fail if there was an update and things changed. Not sure if there's a way to have some configuration that will prevent updates to the linter/formatter causing changes to code we haven't touched

PatMyron: +1, hit unrelated linting failures in https://github.com/aws-cloudformation/cloudformation-cli/pull/554 and seems weird to add unrelated formatting changes to these PRs

Issue #248, opened 2019-06-10
Remove example schemas in favour of repo pointers


rjlohan:
Once we have actual provider repos in place, we can remove the example schemas from this repo in favour of some canonical repository pointers. Static files in this repo will go stale over time.

We do have a couple of repositories available already, but they reflect simpler resources, so I'll wait on a couple more before doing this.

Issue #374, opened 2019-12-23
Publish AWS common definitions


dpantke:
I noticed that in your examples, for AWS resources ( which you are using internally I'd imagine ), you have a common definition file for types that can be included for reference. The one I'm specifically referring to here is aws.common.types.v1.json . I would like that to be published at a well-known URL so that I can reference it to get easy reference when dealing with common parameters I want validated like availability zone, region, ARN, etc....

rjlohan:
These schemas are planned to be published soon - probably in the next couple of weeks. Stay tunes.

tobywf:
I believe this is now available in some regions, for example: https://schema.cloudformation.us-east-1.amazonaws.com/aws.common.types.v1.json

benkehoe:
Is there a root URL that makes these types discoverable?

rjlohan:
> Is there a root URL that makes these types discoverable?

Yeah this should be announced soon, once it's available in all regions is usually the process.

PatMyron:
[The GovCloud partition is incorrect and the isolated partitions are missing in the current `Arn` definition](https://docs.aws.amazon.com/sdk-for-go/api/aws/endpoints/#pkg-constants)

There will be more partitions in the future as well, so let's just generalize it beyond the current partitions

---

The `AvailabilityZone` definition also does not match [`us-west-2-lax-1a`](https://aws.amazon.com/blogs/aws/aws-now-available-from-a-local-zone-in-los-angeles/)

PR #559, opened 2020-08-31
remove example schemas

PatMyron:
fixes https://github.com/aws-cloudformation/cloudformation-cli/issues/248

[public examples here](https://github.com/aws-cloudformation?q=aws-cloudformation-resource-providers-&type=public)

[although none of them seem to be using any of the common definitions yet](https://github.com/aws-cloudformation/cloudformation-cli/issues/374)


rjlohan (approved):

LaikaN57 (approved):
From a consumer perspective, there are better public resources to use rather than looking at these examples. I agree with removing these.

Issue #540, opened 2020-08-12
Additional Identifiers throwing Validation Errors


advaj:
My Schema is as follows: 

```
  "properties": {
        "TPSCode": {
            "description": "A TPS Code is automatically generated on creation and assigned as the unique identifier.",
            "type": "string",
            "pattern": "^[A-Z]{3,5}[0-9]{8}-[0-9]{4}$"
        },
        "TestValue": {
            "description": "The value of the API key. Must be at least 20 characters long.",
            "type": "string",
            "minLength": 20
        },
        "Title": {
            "description": "The title of the TPS report is a mandatory element.",
            "type": "string",
            "minLength": 20,
            "maxLength": 250
        },
        "CoverSheetIncluded": {
            "description": "Required for all TPS Reports submitted after 2/19/1999",
            "type": "boolean"
        },
        "DueDate": {
            "$ref": "#/definitions/InitechDateFormat"
        },
        "ApprovalDate": {
            "$ref": "#/definitions/InitechDateFormat"
        },
        "Memo": {
            "$ref": "#/definitions/Memo"
        },
        "SecondCopyOfMemo": {
            "description": "In case you didn't get the first one.",
            "$ref": "#/definitions/Memo"
        },
        "TestCode": {
            "type": "string",
            "enum": [
                "NOT_STARTED",
                "CANCELLED"
            ]
        },
        "Authors": {
            "type": "array",
            "items": {
                "type": "string"
            }
        }
    },
    "additionalProperties": false,
    "required": [
        "TestCode",
        "Title"
    ],
    "additionalIdentifiers": [
        "/properties/TestValue"
    ],
    "readOnlyProperties": [
        "/properties/TPSCode"
    ],
    "primaryIdentifier": [
        "/properties/TPSCode"
    ],
```
When I execute cfn validate on this schema, the following error is thrown - 

```
$ cfn validate
Resource specification is invalid: '/properties/TestValue' is not of type 'array'

Failed validating 'type' in schema['properties']['additionalIdentifiers']['items']:
    {'items': {'format': 'json-pointer', 'type': 'string'},
     'minItems': 1,
     'type': 'array'}

On instance['additionalIdentifiers'][0]:
    '/properties/TestValue'
```

ammokhov:
`additionalIdentifier` is supposed to be a list of lists, so the correct value would be:

```
{
    ...
    "additionalIdentifiers": [
        ["/properties/TestValue"]
    ],
    ...
}
```

PR #562, opened 2020-09-01
adding additional identifiers validation

ammokhov:
*Issue #, if available:* #540 

*Description of changes:*

enhancing validation on additional identifiers


aygold92 (changes requested):
as discussed offline, additionalIdentifiers SHOULD be a list of lists.  However, our README documentation should be updated to clearly reflect that, as it is confusing now

PR #564, opened 2020-09-02
0.1.8

anshikg:
*Issue #, if available:*

*Description of changes:* bumping the version to release the latest changes to PyPI

### DIFF
--- a/src/rpdk/core/__init__.py
+++ b/src/rpdk/core/__init__.py
@@ -1,5 +1,5 @@
 import logging
 
-__version__ = "0.1.7"
+__version__ = "0.1.8"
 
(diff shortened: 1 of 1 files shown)


src/rpdk/core/__init__.py
```diff
@@ -1,5 +1,5 @@
 import logging
 
-__version__ = "0.1.7"
+__version__ = "0.1.8"
```
jotompki, line 3: unrelated to the release, but we should really try to make a change to 1.x.0 at some point for versioning. doesn't really make sense for this product to be in the 0.1.x versioning almost a year after initial release.

ammokhov (approved):

PR #565, opened 2020-09-02
Restructure request to match handlers

jotompki:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/issues/112 https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/issues/109

*Description of changes:* The request to the resource handlers for contract tests is slightly wrong in structure, causing hard failures while testing via the python plugin (partition isn't a field in the [request](https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/blob/master/src/cloudformation_cli_python_lib/utils.py#L83-L95)).

This should fix the request. Tested with the python plugin and tests ran succesfully with some other changes to the lib that will go out after this. Also bumping the version to expedite a release.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -347,7 +347,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             "region": region,
             "awsAccountId": account,
             "action": action,
-            "requestContext": {"callbackContext": callback_context},
+            "callbackContext": callback_context,
             "bearerToken": token,
             **kwargs,
         }
@@ -416,7 +416,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         payload_to_log = {
             key: payload[key]
             for key in [
-                "requestContext",
+                "callbackContext",
                 "action",
                 "requestData",
                 "region",
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -133,7 +133,6 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         )
         self.region = region
         self.account = get_account(self._session, self._creds)
-        self.partition = self._get_partition()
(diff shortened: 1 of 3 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -161,13 +160,6 @@ def __init__(
         self._inputs = inputs
         self._timeout_in_seconds = int(timeout_in_seconds)
 
-    def _get_partition(self):
```
anshikg, line 164: for handlers which use partition. How will that be handled?

src/rpdk/core/contract/resource_client.py
```diff
@@ -354,10 +345,9 @@ def make_request(
                 "previousResourceProperties": previous_resource_state,
             },
             "region": region,
-            "awsPartition": partition,
             "awsAccountId": account,
             "action": action,
-            "callbackContext": callback_context,
+            "requestContext": {"callbackContext": callback_context},
```
anshikg, line 350: good catch

jotompki: nevermind on this. callbackContext is right

src/rpdk/core/contract/resource_client.py
jotompki: we infer on the handler side: https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/blob/master/src/cloudformation_cli_python_lib/utils.py#L135

https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/a4bf268462367ea43c6a3f5269a9ede42d1b2a89/python/rpdk/java/templates/generate/HandlerWrapper.java#L127-L142

anshikg: thanks for the info. It seems that the prod entry should pick up the partition information

anshikg (approved):

ammokhov (approved):

PR #572, opened 2020-09-09
Fix handler delete test failures

prerna-p:
*Issue #, if available:*
N/A
*Description of changes:*
**Refining the check for error code "Not Found" when the handler fails to delete a resource**.
 When we delete a resource and then try to fetch it, handler should return "Not Found" error code. However, for resources which have writeable identifiers, this is not the case. The test fails due this error code not being returned for this specific case and therefore I have added a condition to avoid failure for such resources

### DIFF
--- a/src/rpdk/core/contract/suite/handler_delete.py
+++ b/src/rpdk/core/contract/suite/handler_delete.py
@@ -7,10 +7,7 @@ import pytest
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
 from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
-from rpdk.core.contract.resource_client import (
-    create_model_with_properties_in_path,
-    prune_properties_from_model,
-)
+from rpdk.core.contract.resource_client import prune_properties_from_model
 from rpdk.core.contract.suite.handler_commons import (
     test_create_success,
     test_delete_failure_not_found,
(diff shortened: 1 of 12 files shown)


anshikg (approved):
The changes look good. Please add a little more description about this change in the details section

prerna-p:
> The changes look good. Please add a little more description about this change in the details section

updated description

src/rpdk/core/contract/suite/handler_delete.py
```diff
@@ -42,8 +42,12 @@ def deleted_resource(resource_client):
     finally:
         status, response = resource_client.call(Action.DELETE, model)
 
-        # a failed status is allowed if the error code is NotFound
-        if status == OperationStatus.FAILED:
+        # a failed status is allowed if the error code is NotFound for
+        # resource having read-only primary identifier
+        if (
+            not resource_client.has_writable_identifier()
```
ammokhov, line 48: could you plz clarify this a bit `However, for resources which have writeable identifiers, this is not the case.`? why writable is different from non writable id?

prerna-p: the delete fails with different error codes for this case

ammokhov: how come

anshikg: hey Anton, for resources which have readOnly properties which are primaryIdentifiers the final delete fails with different error codes depending on the service because, the readOnly property is null. The service does not null primary identifier to be passed. We can not expect them to throw NotFound error code in this case.

anshikg: I think there is a little bit of misunderstanding. The condityion should be following:
```
if (
           resource_client.has_writable_identifier() and status == OperationStatus.FAILED
):

jotompki (changes requested):
do we have an example of where this is ocurring? i feel like instead of making exceptions here the exceptions for this specific use case should be rolled into the handler logic. this seems like more of a one off thing than something we need to apply to all resource types

Issue #410, opened 2020-03-23
Support CLI arguments instead of stdin


iann0036:
Specifically for `cfn init`, I would like to be able to do something like:

`cfn init --type-name Test::Test::Test --provider python37 --use-docker`

ammokhov:
should be implemented in #574

PR #574, opened 2020-09-11
Support CLI arguments for cfn init

chuchocd85:
*Issue #, if available:*
#410 

*Description of changes:*
Add the ability to pass as CLI arguments all the required values to create projects on all the three public language plugins. These new arguments are:

* --language
* --type-name
* --use-docker
* --namespace
* --codegen-model
* --import-path

Ideally this PR should be merged (or at least the changes. released) after [#121](https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/pull/121), [#309](https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/309) and [#168](https://github.com/aws-cloudformation/cloudformation-cli-go-plugin/pull/168).


jotompki (commented):
The only thing that I am hesitant about this approach is that running `cfn init --help` will produce a lot very verbose set all arguments used across plugins. I know we kind of went over this previously, but maybe it is possible to add a unique parser for each plugin, then surface this using the plugin registry? 

Though this may just be something we look at down the road. Other than that the current approach looks good, just want someone else to take a look at it.

src/rpdk/core/init.py
```diff
@@ -172,7 +196,52 @@ def setup_subparser(subparsers, parents):
...
+        "--use-docker",
+        action="store_true",
+        help="""Use docker for python platform-independent packaging.
+            This is highly recommended unless you are experienced
+            with cross-platform Python packaging.""",
+    )
+
+    parser.add_argument(
+        "-n",
+        "--namespace",
```
ammokhov, line 229: @jotompki do you think we could use something like a config file to specify all the metadata - [ConfigArgParse](https://pypi.org/project/ConfigArgParse/) is one of the options

src/rpdk/core/plugin_registry.py
```diff
@@ -5,7 +5,13 @@
     for entry_point in pkg_resources.iter_entry_points("rpdk.v1.languages")
 }
 
-PLUGIN_CHOICES = sorted(PLUGIN_REGISTRY.keys())
+
+def get_plugin_choices():
+    plugin_choices = [
+        entry_point.name
+        for entry_point in pkg_resources.iter_entry_points("rpdk.v1.languages")
+    ]
```
ammokhov, line 13: I would use set if you dont rely on element position

ammokhov (changes requested):

src/rpdk/core/init.py
```diff
@@ -2,13 +2,12 @@
...
 import logging
 import re
-from argparse import SUPPRESS
 from functools import wraps
 
 from colorama import Fore, Style
 
 from .exceptions import WizardAbortError, WizardValidationError
-from .plugin_registry import PLUGIN_CHOICES
+from .plugin_registry import get_plugin_choices
```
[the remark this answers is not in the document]
chuchocd85: For testing we add a dummy plugin on the fly but at that point in time `PLUGIN_CHOICES` is already a constant that does not includes it, making it a method allows to get the latest installed plugins.

src/rpdk/core/init.py
```diff
@@ -42,7 +41,7 @@ def validate_type_name(value):
         return value
     LOG.debug("'%s' did not match '%s'", value, TYPE_NAME_REGEX)
     raise WizardValidationError(
```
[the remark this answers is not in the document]
chuchocd85: That is indeed retried, `input_with_validation` handles the retrying.

src/rpdk/core/init.py
```diff
@@ -139,14 +138,39 @@ def init(args):
...
         language = input_language()
 
-    project.init(type_name, language)
+    project.init(
+        type_name,
+        language,
+        {
+            "use_docker": args.use_docker,
+            "namespace": args.namespace,
+            "codegen_template_path": args.codegen_model,
```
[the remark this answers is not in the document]
chuchocd85: At the moment, each plugin looks for the arguments it needs in `project.settings`, the fact that additional un-related information is also coming in does not affects the plugin. We sure can do some validation and only allow arguments related to the selected plugin but I don't know if is worth doing that validation or how it will impact the user experience, if the user passes a wrong flag s/he would need to try again (kind of defeating the purpose of being able to use just a one-liner). What do you think?

src/rpdk/core/plugin_registry.py
chuchocd85: Before my changes the list of plugins was already being sorted which converts the set back to a list. I f this is not required anymore I'll be ok with making it a set.

tests/test_init.py
```diff
@@ -17,62 +16,93 @@
...
 
-def test_init_method_noninteractive_language():
-    type_name = object()
+def test_init_method_noninteractive():
+    add_dummy_language_plugin()
+
+    args = get_args()
+    mock_project, patch_project = get_mock_project()
+
+    patch_tn = patch("rpdk.core.init.input_typename")
```
jotompki, line 56: do these need to be patched if we are going the non-interactive route?

tests/test_init.py
```diff
@@ -17,62 +16,93 @@
...
+    patch_l = patch("rpdk.core.init.input_language")
```
jotompki, line 57: same here. no reason to patch if we are just asserting we aren't calling

tests/utils.py
```diff
@@ -67,6 +73,79 @@ def chdir(path):
...
+            "type_name",
+            "settings",
+        ]
+    )
+
+    args.language = (
+        None if interactive else ("dummy" if language else "invalid_language")
+    )
+    args.type_name = (
+        None if interactive else ("Test::Test::Test" if type_name else "Test")
```
jotompki, line 111: let's just pass in a type name into this get_args util instead of passing in boolean that sets between two values. makes the function more straightforward

tests/utils.py
```diff
@@ -67,6 +73,79 @@ def chdir(path):
...
+    args = Mock(
+        spec_set=[
+            "language",
+            "type_name",
+            "settings",
+        ]
+    )
+
+    args.language = (
+        None if interactive else ("dummy" if language else "invalid_language")
```
jotompki, line 108: same here as below. no need to obfuscate what is going on

tests/test_init.py
```diff
@@ -17,62 +16,93 @@
...
+
+    mock_tn.assert_not_called()
+    mock_l.assert_not_called()
+    mock_parser.assert_called_once()
+
+    mock_project.load_settings.assert_called_once_with()
+    mock_project.init.assert_called_once_with(
+        args.type_name,
+        args.language,
+        args.settings,
```
jotompki, line 74: usually args doesn't have a settings member, correct? let's keep args as simple as possible and in line with what the cli expects its structure to be. it looks like here it is only used to keep track of test values. can always pull that logic out of the get_args method and inline it here.

tests/test_init.py
```diff
@@ -17,62 +16,93 @@
...
 
-    args = Mock(spec_set=["force", "language"])
-    args.force = False
-    args.language = None
-
-    mock_project = Mock(spec=Project)
-    mock_project.load_settings.side_effect = FileNotFoundError
-    mock_project.settings_path = ""
-    mock_project.root = Path(".")
+    args = get_args(interactive=True)
```
jotompki, line 29: is this even being used in this test other than the settings?

jotompki (commented):
overall logic looks good i just think the get_args method is being abused a bit to keep track of testing values

src/rpdk/core/init.py
```diff
@@ -1,14 +1,14 @@
...
 import logging
 import re
-from argparse import SUPPRESS
 from functools import wraps
 
 from colorama import Fore, Style
 
 from .exceptions import WizardAbortError, WizardValidationError
-from .plugin_registry import PLUGIN_CHOICES
+from .plugin_registry import PARSER_REGISTRY, get_plugin_choices
```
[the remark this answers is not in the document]
chuchocd85: Discussed.

tests/test_init.py
```diff
@@ -17,62 +16,105 @@
...
+            "force": False,
+            "type_name": args.type_name,
+            "language": args.language,
+            "dummy": True,
+        },
+    )
+    mock_project.generate.assert_called_once_with()
+
+
+def test_init_method_noninteractive_invalid_type_name():
```
jotompki, line 86: confused on this. how does init reach project init method if the typename is invalid?

jotompki (approved):

ammokhov (approved):

PR #579, opened 2020-09-17
Initial support using a docker image directly instead of using sam cli

ninavolkmuth:
*Description of changes:* This introduces a new command cfn build-image and changes to resource_client.

The command build-image will initially only support the java plugin, but can be expanded to other plugins. The command will create an image based off of the jar in the target folder. The image can then be run with docker in order to provide an alternative to sam cli. The entry point to the container is a new entry point that does not have a dependency on lambda.

When an image name and entry point is specified for the resource client, it will start a docker container instead of invoking a lambda.

Relevant java plugin cr: aws-cloudformation/cloudformation-cli-java-plugin#308

Tested with unit tests and manually with a personal resource

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,6 +54,16 @@ cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD 
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 ```
 
+### Command: build-image
+
+To build an image
+
+```bash
+cfn test
+cfn test --image-name my-handler --executable target/myjar.jar
+```
+
+
 
 ## Development
 
--- a/README.md
+++ b/README.md
@@ -56,13 +56,19 @@ cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 
 ### Command: build-image
 
-To build an image
+To build an image. This image provides a minimalistic execution environment for the handler that does not depend on AWS Lambda in anyway. This image can be used during cfn invoke and cfn test instead of using sam cli.
 
 ```bash
 cfn test
 cfn test --image-name my-handler --executable target/myjar.jar
 ```
 
+The resulting image can be run in a container by executing the following command:
+
+```
+docker run IMAGE_NAME HANDLER_ENTRYPOINT PAYLOAD
+docker run my-test-resource com.my.test.resource.ExecutableHandlerWrapper PAYLOAD_JSON # Example for a java based-project
+```
 
 
 ## Development
--- a/README.md
+++ b/README.md
@@ -59,8 +59,8 @@ cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 To build an image. This image provides a minimalistic execution environment for the handler that does not depend on AWS Lambda in anyway. This image can be used during cfn invoke and cfn test instead of using sam cli.
 
 ```bash
-cfn test
-cfn test --image-name my-handler --executable target/myjar.jar
+cfn build-image
+cfn build-image --image-name my-handler --executable target/myjar.jar
 ```
 
 The resulting image can be run in a container by executing the following command:
--- a/setup.py
+++ b/setup.py
@@ -49,6 +49,7 @@ setup(
         "requests>=2.22",
         "hypothesis>=4.32",
         "colorama>=0.4.1",
+        "docker>=4.3.1",
     ],
     entry_points={
         "console_scripts": ["cfn-cli = rpdk.core.cli:main", "cfn = rpdk.core.cli:main"]
--- a/src/rpdk/core/build_image.py
+++ b/src/rpdk/core/build_image.py
@@ -1,0 +1,64 @@
+"""This sub command will build an image for the executable
+
+Projects can be created via the 'init' sub command.
+"""
+import logging
+import os
+
+import docker
+
+from .project import Project
+
+LOG = logging.getLogger(__name__)
+VALID_IMAGE_RUNTIMES = [
+    "java8",
+    "java11",
+]
+
+
+def _get_executable_name(runtime, type_info):
+    name = None
+    if "java" in runtime:
+        name = "-".join(type_info).lower() + "-handler-1.0-SNAPSHOT"
+    return name
+
+
+def build_image(_args):
+    project = Project()
+    project.load()
+
+    if project.runtime not in VALID_IMAGE_RUNTIMES:
+        raise ValueError(
+            "Runtime '{}' is not supported for building an image".format(
+                project.runtime
+            )
+        )
+
+    executable_name = _get_executable_name(project.runtime, project.type_info)
+    dockerfile_path = (
+        os.path.dirname(os.path.realpath(__file__))
+        + "/data/build-image-src/Dockerfile-"
+        + project.runtime
+    )
+    project_path = os.path.dirname(os.path.realpath(__name__))
+    image_name = (
+        _args.image_name if _args.image_name else "-".join(project.type_info).lower()
+    )
+
+    docker_client = docker.from_env()
+    LOG.warning("Creating image")
+    docker_client.images.build(
+        path=project_path,
+        dockerfile=dockerfile_path,
+        tag=image_name,
+        buildargs={"executable_name": executable_name},
+    )
+    LOG.warning("Image '%s' created", image_name)
+
+
+def setup_subparser(subparsers, parents):
+    # see docstring of this file
+    parser = subparsers.add_parser("build-image", description=__doc__, parents=parents)
+    parser.set_defaults(command=build_image)
+
+    parser.add_argument("--image-name", help="Image name")
--- a/src/rpdk/core/build_image.py
+++ b/src/rpdk/core/build_image.py
@@ -13,19 +13,19 @@ from .project import Project
 LOG = logging.getLogger(__name__)
 
 
-def build_image(_args):
+def build_image(args):
     project = Project()
     project.load()
 
     config = project.generate_image_build_config()
 
-    if _args.executable:
-        executable_name = _args.executable
+    if args.executable:
+        executable_name = args.executable
     else:
         executable_name = config["executable_name"]
 
     image_name = (
-        _args.image_name if _args.image_name else "-".join(project.type_info).lower()
+        args.image_name if args.image_name else "-".join(project.type_info).lower()
     )
 
     docker_client = docker.from_env()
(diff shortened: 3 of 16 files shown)


src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,64 @@
...
+import logging
+import os
+
+import docker
+
+from .project import Project
+
+LOG = logging.getLogger(__name__)
+VALID_IMAGE_RUNTIMES = [
+    "java8",
```
wbingli, line 14: Could we delegate the docker image and runtime configuration to each plugin?

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,64 @@
...
+
+    if project.runtime not in VALID_IMAGE_RUNTIMES:
+        raise ValueError(
+            "Runtime '{}' is not supported for building an image".format(
+                project.runtime
+            )
+        )
+
+    executable_name = _get_executable_name(project.runtime, project.type_info)
+    dockerfile_path = (
```
wbingli, line 38: Same, could we delegate the runtime and docker configuration to each plugin?

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,64 @@
...
+
+    docker_client = docker.from_env()
+    LOG.warning("Creating image")
+    docker_client.images.build(
+        path=project_path,
+        dockerfile=dockerfile_path,
+        tag=image_name,
+        buildargs={"executable_name": executable_name},
+    )
+    LOG.warning("Image '%s' created", image_name)
```
wbingli, line 56: Could we print more details of the image? such as Image ID.

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,64 @@
...
+        + "/data/build-image-src/Dockerfile-"
+        + project.runtime
+    )
+    project_path = os.path.dirname(os.path.realpath(__name__))
+    image_name = (
+        _args.image_name if _args.image_name else "-".join(project.type_info).lower()
+    )
+
+    docker_client = docker.from_env()
+    LOG.warning("Creating image")
```
wbingli, line 49: Add more information to the log, like the image name to the log.

src/rpdk/core/data/build-image-src/Dockerfile-java11
```diff
@@ -0,0 +1,4 @@
+FROM openjdk:11-alpine
+ARG executable_name
+ADD target/${executable_name}.jar handler.jar
+ENTRYPOINT ["java", "-Xmx256M", "-cp", "handler.jar"]
```
wbingli, line 4: Use 512M, the same as current runtime in service.

src/rpdk/core/build_image.py
ninavolkmuth: Could you clarify what you mean by delegate the image and runtime configuration to the plugin? Are you saying create the dockerfile as part of the plugin? So, for instance, that it would be created when running cfn generate?

For me, it makes more sense to have this logic as part of the cloudformation cli. The plugins shouldn't really care how the executables are being consumed. Furthermore, I thought that putting everything in the plugin would require more effort for existing projects to onboard this feature. We would probably need to add the executable name as an additional field in .rpdk_config which looked like was only generated once during project initialization.

ninavolkmuth: Talked offline. The plugin can generate the image settings in a separate method like:

```
image_config = project._plugin.generate_image_config()
```

Made these changes

README.md
```diff
@@ -54,6 +54,22 @@ cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 ```
 
+### Command: build-image
+
+To build an image. This image provides a minimalistic execution environment for the handler that does not depend on AWS Lambda in anyway. This image can be used during cfn invoke and cfn test instead of using sam cli.
+
+```bash
+cfn test
```
wbingli, line 62: Could we have example of this command itself? `cfn build-image ....`?

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,59 @@
+"""This sub command will build an image for the executable
+
+Projects can be created via the 'init' sub command.
```
wbingli, line 3: nit: This can be removed.  We have this comment for almost all the commands, but it's not useful.

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,59 @@
...
+        executable_name = _args.executable
+    else:
+        executable_name = config["executable_name"]
+
+    image_name = (
+        _args.image_name if _args.image_name else "-".join(project.type_info).lower()
+    )
+
+    docker_client = docker.from_env()
+    LOG.warning("Creating image with name '%s'", image_name)
```
wbingli, line 32: why it's `warning`? not `info`?

src/rpdk/core/contract/resource_client.py
```diff
@@ -413,20 +421,39 @@ def _call(self, payload):
...
+                    self._docker_image,
+                    self._executable_entrypoint + " '" + payload + "'",
+                )
+                .decode()
+                .strip()
+            )
+            LOG.debug("=== Handler execution logs ===")
+            LOG.debug(result)
+            # pylint: disable=W1401
+            regex = "StartResponse-([\s\S]*)-EndResponse"  # noqa: W605
```
wbingli, line 441: I'm afraid  `StartResponse-` and `-EndResponse` is not special enough,  it's possible customer may have property value like this.. Think about something like `__CFN_RESOURCE_START_RESPONSE__`

src/rpdk/core/project.py
```diff
@@ -427,6 +437,15 @@ def generate_docs(self):
         )
         self.safewrite(readme_path, contents)
 
+    def generate_image_build_config(self):
+        if not hasattr(self._plugin, "generate_image_build_config"):
+            raise InvalidProjectError(
```
wbingli, line 442: nice

src/rpdk/core/project.py
```diff
@@ -222,14 +225,21 @@ def write_settings(self):
...
+                else {}
+            )
             json.dump(
                 {
                     "typeName": self.type_name,
                     "language": self.language,
                     "runtime": self.runtime,
                     "entrypoint": self.entrypoint,
                     "testEntrypoint": self.test_entrypoint,
+                    "executableEntrypoint": self.executable_entrypoint,
```
jotompki, line 240: what's the difference between adding this key here and dumping the contents of the other dict?

README.md
ninavolkmuth: Fixed

src/rpdk/core/build_image.py
ninavolkmuth: Removed

src/rpdk/core/build_image.py
ninavolkmuth: I wanted to print something to the terminal at all times so that the developer knows something is happening

src/rpdk/core/contract/resource_client.py
ninavolkmuth: Changed

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,59 @@
...
+import docker
+from docker.errors import APIError, BuildError
+
+from .exceptions import DownstreamError
+from .project import Project
+
+LOG = logging.getLogger(__name__)
+
+
+def build_image(_args):
```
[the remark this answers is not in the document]
ninavolkmuth: Removed

src/rpdk/core/build_image.py
```diff
@@ -0,0 +1,59 @@
...
+            path=config["project_path"],
+            dockerfile=config["dockerfile_path"],
+            tag=image_name,
+            buildargs={"executable_name": executable_name},
+        )
+    except (BuildError, APIError, TypeError) as e:
+        raise DownstreamError("An error occurred when building the image") from e
+    LOG.debug("=== Image build logs ===")
+    for log in logs:
+        LOG.debug(log)
```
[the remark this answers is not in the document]
ninavolkmuth: I think that may be a bit messy since we don't actually need the list for anything? I think this way is a bit easier to read

src/rpdk/core/contract/resource_client.py
```diff
@@ -109,6 +112,8 @@ def __init__(
         inputs=None,
         role_arn=None,
         timeout_in_seconds="30",
+        docker_image=None,
+        executable_entrypoint=None,
```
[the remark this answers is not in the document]
ninavolkmuth: I agree that will make the code a bit cleaner. I can do that in a separate pr

src/rpdk/core/invoke.py
```diff
@@ -80,4 +86,9 @@ def setup_subparser(subparsers, parents):
         "exiting. If not specified, will continue to "
         "re-invoke until terminal status is reached.",
     )
+    parser.add_argument(
+        "--docker-image",
+        help="Docker image name to run. If specified, invoke will use docker instead "
+        "of SAM",
```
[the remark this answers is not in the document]
ninavolkmuth: Mutually exclusive groups unfortunately do not work if you want to specify one argument or multiple arguments. I added manual validation for this instead.

src/rpdk/core/project.py
ninavolkmuth: Good catch. I should not have been adding it. Updated unit tests to catch this case

wbingli (approved):

jotompki (approved):

setup.py
```diff
@@ -49,6 +49,7 @@ def find_version(*file_paths):
         "requests>=2.22",
         "hypothesis>=4.32",
         "colorama>=0.4.1",
+        "docker>=4.3.1",
```
kddejong, line 52: This line conflicts with https://github.com/aws/aws-sam-cli/blob/develop/requirements/base.txt#L11 resulting in conflicts when using this with sam-cli.

PatMyron: reported here as well: https://github.com/aws-cloudformation/cloudformation-cli-python-plugin/issues/147

PatMyron: tbh [SAM CLI requirements are unreasonably tight](https://github.com/aws/aws-sam-cli/blob/a8525fc8157d507c4b102477ded4d221deaed145/requirements/base.txt#L11) because [they insist it shouldn't be installed with pip](https://github.com/aws/aws-sam-cli/issues/1424)

Issue #580, opened 2020-09-18
[contract test] Drift parity


anshikg:
Currently in contract test we compare if the input to a handler is equal to out for create and update handlers. This functionality check needs to be modified to work similar to Drift detection.

- [x] In drift, if a handler input does not contain a property but the output has the property set the resource is not drifted. We should maintain the same functionality in contract test.

- [x] In drift, if a resource has a list and it does not care about the order of the list we can add ```insertionOrder``` flag and this allows the drift to pass. 

Contract test need to have same functionality to ensure the un ordered list resources pass the contract test.

prerna-p:
About #1, isn't this just handled by [these pruning steps](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_commons.py#L150-L159)?

roolerzz:
On the similar lines, does the output of Create Handler also needs to match field by field to the model returned by the read handler?(for e.g. The current behavior I am seeing with our service is say the model returned by Create has fields A, B, C which was set in the create input, but the backend resource created has the attributes A, B, C, D, E(D & E being the defaults.)).
Now when the read handler runs, it fetches the state of the resource from the backend and populates and returns a model having the attributes(A, B, C, D, E)(Read won't necessarily remember what the inputs were to create.) Seems like some of the contract tests seem to fail validating asserting output of create(A, B, C) should be equal to output of Read(A, B, C, D, E). 

Q: Should there be this expectation that model output of CREATE should be deep equals to model output of READ, or similar to the drift you mentioned above, the system should support extra attributes in the model output of READ?

PR #589, opened 2020-09-25
Fix for contract test drift parity issue #1

prerna-p:
*Issue #* #580

*Description of changes:*
Contract test is failing for resources when the output model has properties that are not present in input model. To fix this, I am truncating the output model if it has any properties that are not present in input model. 


prerna-p:
I was not able to test my change locally, still working on that. I also think we can consider using [deep diff](https://pypi.org/project/deepdiff/) since we resource models can be complex json objects so python dict comparisons may not suffice

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -154,8 +154,15 @@ def test_input_equals_output(resource_client, input_model, output_model):
...
-
-    assert pruned_input_model == pruned_output_model
+    # only comparing properties in input model to those in output model and
+    # ignoring extraneous properties that maybe present in output model.
+    trunc_output_model = {}
+    for key in pruned_input_model:
+        if key in pruned_output_model:
+            trunc_output_model[key] = pruned_output_model[key]
+
+    assert pruned_input_model == trunc_output_model
```
ammokhov, line 168: can we rewrite this a bit?

```python
assert all([pruned_input_model[key] == pruned_output_model[key] for key in pruned_input_model if key in pruned_output_model]), "Some alerting details why it failed"
```

ammokhov (changes requested):

src/rpdk/core/contract/suite/handler_commons.py
prerna-p: Can I use same error message as contract_asserts::response_contains_resource_model_equal_current_model ?
>All properties specified in the request MUST be present in the model  returned, and they MUST match exactly, with the exception of properties defined as writeOnlyProperties in the resource schema

ammokhov: not sure how ambiguous that message is for this comparison; also DeepDiff seems to be a bit overkill for this usecase

anshikg (approved):

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -154,8 +154,15 @@ def test_input_equals_output(resource_client, input_model, output_model):
     pruned_output_model = prune_properties_if_not_exist_in_path(
         output_model.copy(), pruned_input_model, resource_client.read_only_paths
     )
+
     pruned_output_model = prune_properties_if_not_exist_in_path(
         pruned_output_model, pruned_input_model, resource_client.create_only_paths
     )
```
PatMyron, line 160: Nice, I think we can remove this `createOnlyProperty` pruning now
```suggestion
```
https://github.com/aws-cloudformation/cloudformation-cli/pull/531#discussion_r469462358

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -154,8 +154,15 @@ def test_input_equals_output(resource_client, input_model, output_model):
...
-
-    assert pruned_input_model == pruned_output_model
+    # only comparing properties in input model to those in output model and
+    # ignoring extraneous properties that maybe present in output model.
+    assert all(
+        pruned_input_model[key] == pruned_output_model[key]
+        for key in pruned_input_model
```
ammokhov, line 165: hmm, if output model does not return all properties then we will have a runtime error, which will not surface correctly the issue; we should catch KeyError and raise AssertionError

ammokhov (changes requested):

ammokhov (approved):

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -154,8 +154,23 @@ def test_input_equals_output(resource_client, input_model, output_model):
...
     pruned_output_model = prune_properties_if_not_exist_in_path(
         pruned_output_model, pruned_input_model, resource_client.create_only_paths
     )
 
-    assert pruned_input_model == pruned_output_model
+    assertion_error_message = (
+        "All properties specified in the request MUST "
+        "be present in the model returned, and they MUST"
+        " match exactly, with the exception of properties"
+        " defined as writeOnlyProperties in the resource schema"
```
PatMyron, line 166: think this error message might be misleading because there should be more exceptions:
[readOnlyProperties, writeOnlyProperties, defaults, insertionOrder, uniqueItems, etc.](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics)

PR #590, opened 2020-09-29
Remove cloudformation.amazonaws.com Service principal

wbingli:
*Issue #, if available:* N/A

*Description of changes:*. The `LogAndMetricsDeliveryRole` will only be assumed-role by `resources.cloudformation.amazonaws.com`. Remove unused service principal.

### DIFF
--- a/src/rpdk/core/data/managed-upload-infrastructure.yaml
+++ b/src/rpdk/core/data/managed-upload-infrastructure.yaml
@@ -106,7 +106,6 @@ Resources:
         - Effect: Allow
           Principal:
             Service:
-              - cloudformation.amazonaws.com
               - resources.cloudformation.amazonaws.com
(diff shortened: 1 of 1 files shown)


ammokhov (approved):

jotompki:
Looks good. Did we test the submission end to end?

anshikg (commented):
Is there some way we can verify this is not used by any edge condition?

wbingli:
> Looks good. Did we test the submission end to end?

Yes, I did an end-2-end testing. With this change, the resource type provisioning and logging work fine.

> Is there some way we can verify this is not used by any edge condition?

We can ensure this is not used somewhere else, and it should NOT use somewhere.

jotompki (approved):

anshikg (approved):

Issue #585, opened 2020-09-22
Fix writable identifier check for composite key primary identifier


prerna-p:
Description:
When a primary key is composite having createOnly AND readOnly properties, the has_writable_identifier() returns true. This is incorrect and needs to be fixed so that we return true only when none of the primary identifiers are readOnly.

PR #598, opened 2020-10-06
Fix writable identifier check for composite key primary identifier

prerna-p:
*Issue:* #585

*Description of changes:*
The has_writable_identifier() is called during the execution of the contract_create_duplicate() test. 
- We only need to check primaryIdentifiers here, additionalIdentifiers  does not appear to be used. 
- Updated to check if all primaryIdentifiers are readOnly

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -173,7 +173,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for identifier in additional_identifiers
         ]
 
-    def is_primary_identifier_create_only(self):
+    def has_only_writable_identifiers(self):
         return all(
             path in self.create_only_paths for path in self.primary_identifier_paths
         )
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -173,15 +173,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
             for identifier in additional_identifiers
         ]
 
-    def has_writable_identifier(self):
-        for path in self.primary_identifier_paths:
-            if path not in self.read_only_paths:
-                return True
-        for identifier_paths in self._additional_identifiers_paths:
-            for path in identifier_paths:
-                if path not in self.read_only_paths:
-                    return True
-        return False
+    def is_primary_identifier_create_only(self):
+        return all(
+            path in self.create_only_paths for path in self.primary_identifier_paths
+        )
 
     def assert_write_only_property_does_not_exist(self, resource_model):
         if self.write_only_paths:
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -449,7 +449,7 @@ def test_generate_update_example_create_override(resource_client):
     assert example == {"a": 5, "b": 2}
 
 
-def test_has_writable_identifier_primary_is_read_only(resource_client):
+def test_is_primary_identifier_create_only_primary_is_read_only(resource_client):
     resource_client._update_schema(
         {
             "primaryIdentifier": ["/properties/foo"],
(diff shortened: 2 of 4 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -174,14 +174,9 @@ def _update_schema(self, schema):
         ]
 
     def has_writable_identifier(self):
```
anshikg, line 176: please change the name of the function to reflect the functionality change

tests/contract/test_resource_client.py
```diff
@@ -478,28 +478,28 @@ def test_has_writable_identifier_primary_and_additional_are_read_only(resource_c
...
+# def test_has_writable_identifier_compound_is_writeable(resource_client):
+#     resource_client._update_schema(
+#         {
+#             "primaryIdentifier": ["/properties/foo"],
+#             "additionalIdentifiers": [["/properties/bar", "/properties/baz"]],
+#             "readOnlyProperties": ["/properties/foo", "/properties/baz"],
+#         }
+#     )
+#
+#     assert resource_client.has_writable_identifier()
```
anshikg, line 502: please remove the unit test if not required. Also, could you update the unit test with the new functionality

src/rpdk/core/contract/resource_client.py
```diff
@@ -173,15 +173,10 @@ def _update_schema(self, schema):
...
-    def has_writable_identifier(self):
-        for path in self.primary_identifier_paths:
-            if path not in self.read_only_paths:
-                return True
-        for identifier_paths in self._additional_identifiers_paths:
-            for path in identifier_paths:
-                if path not in self.read_only_paths:
-                    return True
-        return False
+    def is_primary_identifier_create_only(self):
```
anshikg, line 176: in this case we are checking if any readonly value exists we will return false. The name does not reflect that

tests/contract/test_resource_client.py
```diff
@@ -449,57 +449,61 @@ def test_generate_update_example_create_override(resource_client):
...
+def test_is_primary_identifier_create_only_composite_primary_is_read_only(
+    resource_client,
+):
     resource_client._update_schema(
         {
-            "primaryIdentifier": ["/properties/foo"],
-            "additionalIdentifiers": [["/properties/bar"]],
+            "primaryIdentifier": ["/properties/foo", "/properties/bar"],
             "readOnlyProperties": ["/properties/foo"],
+            "createOnlyProperties": ["/properties/bar"],
```
anshikg, line 489: why did we change this?

tests/contract/test_resource_client.py
```diff
@@ -449,57 +449,61 @@ def test_generate_update_example_create_override(resource_client):
...
+def test_is_primary_identifier_create_only_composite_primary_is_create_only(
+    resource_client,
+):
     resource_client._update_schema(
         {
-            "primaryIdentifier": ["/properties/foo"],
-            "additionalIdentifiers": [["/properties/bar", "/properties/baz"]],
-            "readOnlyProperties": ["/properties/foo", "/properties/baz"],
+            "primaryIdentifier": ["/properties/foo", "/properties/bar"],
+            "createOnlyProperties": ["/properties/bar", "/properties/foo"],
```
anshikg, line 502: lets not remove additional identifiers just change the results to match what we expect

tests/contract/test_resource_client.py
prerna-p: I added a test to have composite primary identifiers. The old test does not work with the change we're making.

tests/contract/test_resource_client.py
prerna-p: Since the test is only updated primary identifiers and testing them, it maybe confusing to a reader why the additionalIdentifiers are being updated.

ammokhov (approved):

src/rpdk/core/contract/resource_client.py
anshikg, line 176: Please name this more appropriately. Something like ```has_only_writable_identifiers```

PR #602, opened 2020-10-18
Ensure required property is present in generated documentation

eduardomourar:
*Issue #, if available:*

*Description of changes:*

Whenever the required property was part of a combiner (e.g. `anyOf`) in a resource schema, the generated documentation would ignored it. This changed will ensure that required fields will be flattened and then used by the documentation generator.

You can check a documentation example [here](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/tree/master/aws-cloudformation-resourcedefaultversion/docs) where the required property are ignored.

### DIFF
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -421,6 +421,7 @@ class Project:  # pylint: disable=too-many-instance-attributes
             name: self._set_docs_properties(name, value, (name,))
             for name, value in self._flattened_schema[()]["properties"].items()
         }
+        docs_schema["required"] = self._flattened_schema[()].get("required")
 
         LOG.debug("Finished documenting nested properties")
 


ammokhov (approved):

PatMyron:
> You can check a documentation example [here](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/tree/master/aws-cloudformation-resourcedefaultversion/docs) where the required property is ignored.

What does it look like after this change? [Because none of the properties look like they're required by themselves](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cloudformation/blob/733ccc32c77f0b605d553145454c6990de10ddc7/aws-cloudformation-resourcedefaultversion/aws-cloudformation-resourcedefaultversion.json#L27-L39)

eduardomourar:
You can find an example [here](https://github.com/org-formation/aws-resource-providers/blob/master/code-commit/repository-association/docs/README.md) that I ran with this PR's branch. The fields are required true now instead false.

PatMyron:
> You can find an example [here](https://github.com/org-formation/aws-resource-providers/blob/master/code-commit/repository-association/docs/README.md) that I ran with this PR's branch. The fields are required true now instead false.

I'd argue only [`RepositoryNames`](https://github.com/org-formation/aws-resource-providers/blob/9e4f40c2b972df302df6d036a63a14c2e2466eea/code-commit/repository-association/community-codecommit-repositoryassociation.json#L39-L52) is required there. The other properties are only conditionally required

iann0036:
For reference, the official spec uses `Conditional` as an explicit state when representing conditionally required (i.e. anyOf) properties:

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-as-scheduledaction.html#cfn-as-scheduledaction-maxsize

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-s3-bucket-lifecycleconfig-rule.html#cfn-s3-bucket-rule-abortincompletemultipartupload

eduardomourar:
I agree that putting Conditional is the right way, but flattening will not be enough for that logic.

PR #604, opened 2020-10-21
fixed composite id validation

ammokhov:
*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -92,8 +92,8 @@ def get_file_base_uri(file):
 
 
 def _is_in(schema, key):
-    def contains(values):
-        return set(values).issubset(set(schema.get(key, [])))
+    def contains(value):
+        return value in schema.get(key, [])
 
     return contains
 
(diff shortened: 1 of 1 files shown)


PatMyron:
seems it was already there, but confusing for [resource schemas](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/) to be referred to as [Resource Specs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cfn-resource-specification.html) 
https://github.com/aws-cloudformation/cloudformation-cli/pull/425

src/rpdk/core/data_loaders.py
```diff
@@ -92,8 +92,8 @@ def get_file_base_uri(file):
 
 
 def _is_in(schema, key):
-    def contains(values):
-        return set(values).issubset(set(schema.get(key, [])))
+    def contains(value):
+        return value in schema.get(key, [])
 
     return contains
```
PatMyron, line 98: seems like a weird layer of misdirection to have this function return a function rather than just having a function that takes all 3 parameters

PatMyron (approved):

prerna-p (approved):
Agree with Pat's comment

src/rpdk/core/data_loaders.py
ammokhov: this is closure, i'd keep it in case we need to reuse or enhance the logic

PatMyron: just seems unnecessary when the check is a one-liner with or without that function

PatMyron:
Ran this check to find [existing resource provider schemas](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/resource-type-schemas.html) with violations:
```
AWS::Athena::DataCatalog
AWS::ECS::Service
AWS::FMS::NotificationChannel
AWS::IoTWireless::Destination
AWS::Kendra::DataSource
AWS::MediaPackage::Asset
AWS::MediaPackage::OriginEndpoint
AWS::MediaPackage::PackagingConfiguration
AWS::S3::AccessPoint
AWS::SSO::InstanceAccessControlAttributeConfiguration
AWS::WAFv2::IPSet
AWS::WAFv2::RegexPatternSet
AWS::WAFv2::RuleGroup
AWS::WAFv2::WebACL
```

Issue #609, opened 2020-10-28
Readonly properties changing on update causing contract tests to fail


privetjr:
### Motivation for Change
When running contract tests for a resource that has a read-only property that is modified automatically during an update, `contract_update_read_success` and `contract_update_list_success` contract tests fail with:

"AssertionError: All properties specified in the request MUST be present in the model returned, and they MUST match exactly, with the exception of properties defined as writeOnlyProperties in the resource schema"

See [the source here](https://github.com/aws-cloudformation/cloudformation-cli/blob/009e0965d32f51ce94f84af2c5a33698273bc515/src/rpdk/core/contract/suite/handler_commons.py#L149-L176)

Read-only properties that change automatically(e.g. a timestamp for when the resource was last updated) should not need to match. CloudFormation is not able to modify/write the property, but that doesn't mean read-only properties never change. The `test_input_equals_output` method should also prune read-only properties.

anshikg:
Closing this issue as the code has been merged. Feel free to reopen this issue if I have missed something.

PR #610, opened 2020-10-28
prune readonly paths from input model for test_input_equals_output

privetjr:
If read only paths change with updates, contract tests fail because the readonly property value does not match the previous value.
This change prunes readonly paths from the input model so when comparing input model to output model, readonly paths are not considered.

*Issue #, if available:* [609](https://github.com/aws-cloudformation/cloudformation-cli/issues/609)

*Description of changes:* prune readonly paths from input model in `test_input_equals_output`, and update the message accordingly. 

### DIFF
--- a/src/rpdk/core/contract/suite/handler_update.py
+++ b/src/rpdk/core/contract/suite/handler_update.py
@@ -6,6 +6,7 @@ import pytest
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
 from rpdk.core.contract.interface import Action, OperationStatus
+from rpdk.core.contract.resource_client import prune_properties_from_model
 from rpdk.core.contract.suite.handler_commons import (
     test_input_equals_output,
     test_model_in_list,
@@ -26,6 +27,11 @@ def updated_resource(resource_client):
         updated_input_model = update_request = resource_client.generate_update_example(
             created_model
         )
+
+        updated_input_model = prune_properties_from_model(
+            updated_input_model, resource_client.read_only_paths
+        )
+
         _status, response, _error = resource_client.call_and_assert(
             Action.UPDATE, OperationStatus.SUCCESS, update_request, created_model
         )


src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -147,10 +147,15 @@ def test_delete_failure_not_found(resource_client, current_resource_model):
...
 def test_input_equals_output(resource_client, input_model, output_model):
+
     pruned_input_model = prune_properties_from_model(
         input_model.copy(), resource_client.write_only_paths
     )
 
+    pruned_input_model = prune_properties_from_model(
+        pruned_input_model, resource_client.read_only_paths
+    )
+
```
anshikg, line 158: why does the input model have readOnly property? ReadOnly properties should be set by the service when creating the resources

privetjr: Good question. The input model shouldn't have a readOnly property, but for the update contract tests, the output model for the create is used as the input model for the subsequent update. See [handler_update.py#L47-L53](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update.py#L47-L53), [handler_update.py#L23-L28](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update.py#L23-L28), and [resource_client.py#L255-L260](https://github.com/aws-cloudformation/cloudformation-cli/blob/009e0965d32f51ce94f84af2c5a33698273bc515/src/rpdk/core/contract/resource_client.py#L255-L260).

privetjr: Do you think the pruning should happen in `resource_client.generate_update_example` instead?

anshikg: Checked with the team, read_only properties should be updatable using update handler. I can go ahead approve this PR, could you please move the change to update?

privetjr: Yup, sounds good. I will move these lines into `generate_update_example`.

privetjr: Actually, after taking another look, `generate_update_example` is used for more than the update_read_success and update_list_success contract tests. Instead, I'll move this into the `pytest.fixture` - `updated_resource`, which is only used by the two affected tests.

anshikg (approved):

src/rpdk/core/contract/suite/handler_update.py
```diff
@@ -26,6 +27,11 @@ def updated_resource(resource_client):
         updated_input_model = update_request = resource_client.generate_update_example(
             created_model
         )
+
+        updated_input_model = prune_properties_from_model(
+            updated_input_model, resource_client.read_only_paths
```
wbingli, line 32: I like it. Some resources use the readOnlyProperties as to support `GetAtt`  and mixed the readOnly with resource properties. This one can also help to detection such mis-use.

wbingli (approved):

PR #612, opened 2020-11-02
Add propertyTransform to transform the properties to expected output

anshikg:
*Issue #, if available:*

*Description of changes:* Adding a property transform function to define what the read handler would return for created/updated property. This change is ensuring the cfn validate does not fail on propertyTransform.

*Test:*
* pre-commit run --all-files
* run ```cfn validate``` for a schema file with ```propertyTransform``` set.

### DIFF
--- a/src/rpdk/core/data/schema/meta-schema.json
+++ b/src/rpdk/core/data/schema/meta-schema.json
@@ -241,5 +241,13 @@
             "$ref": "#"
         }
     },
+    "propertyTransform": {
+        "type": "object",
+        "patternProperties": {
+            "^[A-Za-z0-9]{1,64}$": {
+                "type": "string"
+            }
+        }
+    },
     "default": true
 }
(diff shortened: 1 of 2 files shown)


ammokhov (approved):

wbingli (approved):

PatMyron:
@anshikg
[thinking about automatically syncing `provider.definition.schema.v1.json` from the `resource-schema` repo](https://github.com/aws-cloudformation/cloudformation-cli/pull/645#discussion_r541484704), and this was the main difference. Should this be added [there](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/master/src/main/resources/schema/provider.definition.schema.v1.json)? Or can it be removed from both places? Or are these two intentionally drifted?

jotompki:
#650 <- We only added `src/rpdk/core/data/schema/meta-schema.json` locally to support local development. This schema should not be modified

Issue #613, opened 2020-11-03
Contract test failed because TypeError: unhashable type: 'dict'


YangMann:
I noticed our contract tests are failing after #608 was merged. 

```
=============================================================================== FAILURES ===============================================================================
________________________________________________________________________ contract_create_delete ________________________________________________________________________
/Users/zyng/Downloads/cloudformation-cli/venv/lib/python3.9/site-packages/rpdk/core/contract/suite/handler_create.py:50: in contract_create_delete
    test_input_equals_output(resource_client, input_model, delete_model)
/Users/zyng/Downloads/cloudformation-cli/venv/lib/python3.9/site-packages/rpdk/core/contract/suite/handler_commons.py:173: in test_input_equals_output
    pruned_input_model[key]
E   TypeError: unhashable type: 'dict'
======================================================================= short test summary info ========================================================================
FAILED handler_create.py::contract_create_delete - TypeError: unhashable type: 'dict'
```

This is because https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_commons.py#L174 uses the value in `pruned_input_model` to check existence in set `resource_client.properties_without_insertion_order`. This creates the aforementioned error when `pruned_input_model` contains nested objects. 

I looked into `resource_client.properties_without_insertion_order`, it seems to be a set of property keys instead of property objects.

PR #614, opened 2020-11-03
use dict key instead of value when checking properties without insert…

YangMann:
*Issue #, if available:* #613 

*Description of changes:* Use dict key of `pruned_input_model` to check property key existence in `resource_client.properties_without_insertion_order`, instead of dict value. Otherwise Python will complain "TypeError: unhashable type: 'dict'" when input model has nested dicts (json objects)


ammokhov (approved):
looks like a bug on our side, thanks for the contribution

anshikg (approved):

PR #615, opened 2020-11-03
Remove redundant check

anshikg:
*Issue #, if available:*

*Description of changes:* Read handler output does not need to be compared with the create handler output. We already make a check for read handler output with create handler input [here](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_create.py#L88-L90)

*Test:*
* pre-commit run --all-files
* cfn test

### DIFF
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -78,17 +78,6 @@ def response_does_not_contain_write_only_properties(resource_client, response):
 
 
 @decorate()
-def response_contains_resource_model_equal_current_model(
-    response, current_resource_model
-):
-    assert (
-        response["resourceModel"] == current_resource_model
-    ), "All properties specified in the request MUST be present in the model \
-        returned, and they MUST match exactly, with the exception of properties\
-             defined as writeOnlyProperties in the resource schema"
-
-
-@decorate()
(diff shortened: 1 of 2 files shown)


src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -77,17 +77,6 @@ def response_does_not_contain_write_only_properties(resource_client, response):
     resource_client.assert_write_only_property_does_not_exist(response["resourceModel"])
 
 
-@decorate()
```
ammokhov, line 80: how come it's redundant?

anshikg: because, we have added input == output for read handler response with create handler input [here](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_create.py#L88-L90). Not sure we need to check read handler output with create handler output

ammokhov: yes but this check asserts that during the execution the model does not change

jotompki: create  response model should match the read handler model. the model in the final progress of the create handler should match what is returned by read

PR #621, opened 2020-11-10
Include format in ResolvedType

jotompki:
*Issue #, if available:*  https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/151

*Description of changes:* This allows the [format keyword](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/blob/master/src/main/resources/schema/provider.definition.schema.v1.json#L201-L203) to be passed to plugins to allow construction of types with specific format akin to the OpenAPI specification for certain primitives (thanks @bkabrda for the suggestion.) This is to be reviewed alongside this java plugin PR: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/328

TL;DR (for specific https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/issues/151 issue):
Customers can now add format to their integer schemas to specify int or long for java. The below schema would result in an Long type being generated in java when both prs are merged & released (omitting format preserves existing functionality)
```
{
    "type": "integer",
    "format": "int64"
}
```
Results in a Long type being generated in resource model pojos.

### DIFF
--- a/src/rpdk/core/jsonutils/resolver.py
+++ b/src/rpdk/core/jsonutils/resolver.py
@@ -7,7 +7,7 @@ from .utils import BASE, fragment_encode
 
 UNDEFINED = "undefined"
 MULTIPLE = "multiple"
-DEFAULT = "default"
+FORMAT_DEFAULT = "default"
 
 
 class ContainerType(Enum):
@@ -20,7 +20,7 @@ class ContainerType(Enum):
 
 
 class ResolvedType:
-    def __init__(self, container, item_type, type_format=DEFAULT):
+    def __init__(self, container, item_type, type_format=FORMAT_DEFAULT):
         self.container = container
         self.type = item_type
         self.type_format = type_format
@@ -135,7 +135,9 @@ class ModelResolver:
     @staticmethod
     def _get_primitive_lang_type(schema_type, property_schema):
         return ResolvedType(
-            ContainerType.PRIMITIVE, schema_type, property_schema.get("format", DEFAULT)
+            ContainerType.PRIMITIVE,
+            schema_type,
+            property_schema.get("format", FORMAT_DEFAULT),
         )
 
     def _get_array_lang_type(self, property_schema):
--- a/src/rpdk/core/jsonutils/resolver.py
+++ b/src/rpdk/core/jsonutils/resolver.py
@@ -7,6 +7,7 @@ from .utils import BASE, fragment_encode
 
 UNDEFINED = "undefined"
 MULTIPLE = "multiple"
+DEFAULT = "default"
 
 
 class ContainerType(Enum):
@@ -19,9 +20,10 @@ class ContainerType(Enum):
 
 
 class ResolvedType:
-    def __init__(self, container, item_type):
+    def __init__(self, container, item_type, type_format=DEFAULT):
         self.container = container
         self.type = item_type
+        self.type_format = type_format
 
     def __repr__(self):
         return f"<ResolvedType({self.container}, {self.type})>"
@@ -114,7 +116,7 @@ class ModelResolver:
             return self._get_array_lang_type(property_schema)
         if schema_type == "object":
             return self._get_object_lang_type(property_schema)
-        return self._get_primitive_lang_type(schema_type)
+        return self._get_primitive_lang_type(schema_type, property_schema)
 
     @staticmethod
     def _get_array_container_type(property_schema):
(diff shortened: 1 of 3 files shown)


src/rpdk/core/jsonutils/resolver.py
```diff
@@ -7,6 +7,7 @@
 
 UNDEFINED = "undefined"
 MULTIPLE = "multiple"
+DEFAULT = "default"
```
ammokhov, line 10: `default` is a reserved name in schema engine, could we use some alternatives as it's a bit confusing

src/rpdk/core/jsonutils/resolver.py
```diff
@@ -131,16 +133,18 @@ def _get_multiple_lang_type(schema_type):
         return ResolvedType(ContainerType.MULTIPLE, schema_type)
 
     @staticmethod
-    def _get_primitive_lang_type(schema_type):
-        return ResolvedType(ContainerType.PRIMITIVE, schema_type)
+    def _get_primitive_lang_type(schema_type, property_schema):
+        return ResolvedType(
+            ContainerType.PRIMITIVE, schema_type, property_schema.get("format", DEFAULT)
+        )
```
ammokhov, line 139: did you test it with python/go plugin?

src/rpdk/core/jsonutils/resolver.py
```diff
@@ -131,16 +133,18 @@ def _get_multiple_lang_type(schema_type):
         return ResolvedType(ContainerType.MULTIPLE, schema_type)
 
     @staticmethod
-    def _get_primitive_lang_type(schema_type):
-        return ResolvedType(ContainerType.PRIMITIVE, schema_type)
+    def _get_primitive_lang_type(schema_type, property_schema):
+        return ResolvedType(
+            ContainerType.PRIMITIVE, schema_type, property_schema.get("format", DEFAULT)
```
ammokhov, line 138: what if the format is not supported? this looks it can cause uncaught runtime exceptions. 
```json
{
    "type": "integer",
    "format": "int0"
}
```

ammokhov (changes requested):

src/rpdk/core/jsonutils/resolver.py
wbingli: `default` is still a good name as long as it's allowed to use.  Maybe the constant name should be changed to more meaningful, such as `FORMAT_DEFAULT`

wbingli (commented):
LGTM, but should check Anton's comment, we should have a list of supported formats and fail if provide invalid one.

src/rpdk/core/jsonutils/resolver.py
jotompki: it's not used by them currently so should be fine. if you look it's just adding a member variable, so other plugins should behave the same. can test to ensure though.

src/rpdk/core/jsonutils/resolver.py
jotompki: see aws-cloudformation/cloudformation-cli-java-plugin#328. Since language plugin would decide if a format is supported or not, was just going to default if not found.

```
            LOG.error(
                "Could not find specified format '%s' for type '%s'. "
                "Defaulting to '%s'",
                resolved_type.type_format,
                resolved_type.type,
                primitive_format,
            )
```

ammokhov: yep, looked at both as one PR. maybe we should have a map of allowed formats on the cli side? so we can default it once and not rewrite logic on plugin side. lmk what u think

src/rpdk/core/jsonutils/resolver.py
jotompki: i can change this variable name. "default" is just used in the mapping of resolved schema types to language primitives, so it wouldn't be something that the user would specify in the schema:
https://github.com/jotompki/cloudformation-cli-java-plugin/blob/d006d5d22afd455d8146d375e6438c82644b3e45/python/rpdk/java/resolver.py#L6-L14

Either way, `default` is a reserved key, whereas this would be a value for the `format` key, so if a user did want to explicitly specify the default for the languages, they would do the following:
```
{
    "type": "integer",
    "format": "default"
}
```

ammokhov (approved):

src/rpdk/core/jsonutils/resolver.py
jotompki: i feel like that be cumbersome to maintain if we start adding a bunch of them to all of the different plugins. what if a certain type like long is supported in java [but not in python?](https://www.python.org/dev/peps/pep-0237/). I think that these lists should be provided by each plugin instead.

jotompki:
> LGTM, but should check Anton's comment, we should have a list of supported formats and fail if provide invalid one.

I think going the route of soft failing/defaulting to the normal format and logging an error to the customer in each of the language plugins might prove better. In the java PR you can see I added a table of supported formats.

wbingli (approved):

PR #627, opened 2020-11-12
Improve the check for created_model == read_model

anshikg:
*Issue #, if available:*

*Description of changes:*
This change ensures the created_model is equal to read_model. This improves that assertion by checking the insertion order.

### DIFF
--- a/src/rpdk/core/contract/suite/contract_asserts.py
+++ b/src/rpdk/core/contract/suite/contract_asserts.py
@@ -78,17 +78,6 @@ def response_does_not_contain_write_only_properties(resource_client, response):
 
 
 @decorate()
-def response_contains_resource_model_equal_current_model(
-    response, current_resource_model
-):
-    assert (
-        response["resourceModel"] == current_resource_model
-    ), "All properties specified in the request MUST be present in the model \
-        returned, and they MUST match exactly, with the exception of properties\
-             defined as writeOnlyProperties in the resource schema"
-
-
-@decorate()
 def response_contains_resource_model_equal_updated_model(
     response, current_resource_model, update_resource_model
 ):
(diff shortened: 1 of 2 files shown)


src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -159,31 +159,6 @@ def test_input_equals_output(resource_client, input_model, output_model):
         pruned_output_model, pruned_input_model, resource_client.create_only_paths
     )
 
-    assertion_error_message = (
-        "All properties specified in the request MUST "
-        "be present in the model returned, and they MUST"
-        " match exactly, with the exception of properties"
-        " defined as writeOnlyProperties in the resource schema"
+    response_contains_resource_model_equal_current_model(
+        resource_client, pruned_input_model, pruned_output_model
```
ammokhov, line 163: why dont you want to include whole method under decorate? so it could be used as annotation

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -139,3 +128,34 @@ def wrapper(*args, **kwargs):
...
+            else:
+                assert input_model[key] == output_model[key], assertion_error_message
+    except KeyError as e:
+        raise AssertionError(assertion_error_message) from e
+
+
+def test_unordered_list_match(inputs, outputs):
+    assert len(inputs) == len(outputs)
+    try:
+        assert all(input in outputs for input in inputs)
```
ammokhov, line 159: maybe instead out `input` and `output` we should be a bit more specific? like handler output keys or something like that

ammokhov (commented):
looks good overall but I would include all parts of `test_input_equals_output` into decorated method so it would be able to use as annotation to any method where we need

anshikg:
> looks good overall but I would include all parts of `test_input_equals_output` into decorated method so it would be able to use as annotation to any method where we need

the only reason I have done that is because, `test_input_equals_output` is used for fixtures. Was not sure if decorators can be used with fixtures

src/rpdk/core/contract/suite/contract_asserts.py
anshikg: this was non_discript for a reason it is being used in multiple places. Like we check input to create handler to output of create handler or input to create handler with output of read handler.

src/rpdk/core/contract/suite/contract_asserts.py
```diff
@@ -139,3 +128,34 @@ def wrapper(*args, **kwargs):
...
+    assertion_error_message = (
+        "All properties specified in the request MUST "
+        "be present in the model returned, and they MUST"
+        " match exactly, with the exception of properties"
+        " defined as writeOnlyProperties in the resource schema"
+    )
+    output_model = output_model["resourceModel"]
+    try:
+        for key in input_model:
+            if key in resource_client.properties_without_insertion_order:
```
jotompki, line 148: won't this only check top level lists without insertion order? what if there's a model like 
```
{
    "a": {
        "anotherList": ["a", "b", "c"]
    }
}
```

src/rpdk/core/contract/suite/contract_asserts.py
tayloryork, line 159: I am experiencing a bug that I think can be resolved in this PR.

Can this be changed to work with deep objects?
`expected = { "top": { "inner1": "1" } }`
`actual = { "top": { "inner1": "1", "inner2": "2" } }`

For me, a deep object is coming back with extra parameters so the test fails.

anshikg: hey Taylor, This is a separate issue. It seems your service is adding generated defaults. Let's talk about this offline.

PR #631, opened 2020-11-17
Create temporary credential for each invocation and limit duration to 15 mins

wbingli:
*Issue #, if available:*. N/A

*Description of changes:*

There are 2 problems to fix here:
 - Only create temporary credential once during the whole contract test is not right.  The resource stabilization could be more than default credential time 1 hour and it will fail. Also this doesn't match how cloudformation get temporary credential for each invocation.

 - With above change, we can limit the credential to 15 mins to enhance the security.

### Testing

Run contract test with log group. (some contract test failed with log group but it's not because of this change)

### DIFF
--- a/src/rpdk/core/boto_helpers.py
+++ b/src/rpdk/core/boto_helpers.py
@@ -44,7 +44,7 @@ def get_temporary_credentials(session, key_names=BOTO_CRED_KEYS, role_arn=None):
         )
         try:
             response = sts_client.assume_role(
-                RoleArn=role_arn, RoleSessionName=session_name
+                RoleArn=role_arn, RoleSessionName=session_name, DurationSeconds=900
             )
         except ClientError:
             # pylint: disable=W1201
@@ -64,7 +64,7 @@ def get_temporary_credentials(session, key_names=BOTO_CRED_KEYS, role_arn=None):
             creds = (frozen.access_key, frozen.secret_key, frozen.token)
         else:
             try:
-                response = sts_client.get_session_token()
+                response = sts_client.get_session_token(DurationSeconds=900)
             except ClientError as e:
                 LOG.debug(
                     "Getting session token resulted in unknown ClientError", exc_info=e
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -117,6 +117,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     ):  # pylint: disable=too-many-arguments
         self._schema = schema
         self._session = create_sdk_session(region)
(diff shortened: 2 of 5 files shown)


src/rpdk/core/boto_helpers.py
```diff
@@ -64,7 +64,7 @@ def get_temporary_credentials(session, key_names=BOTO_CRED_KEYS, role_arn=None):
             creds = (frozen.access_key, frozen.secret_key, frozen.token)
         else:
             try:
-                response = sts_client.get_session_token()
+                response = sts_client.get_session_token(DurationSeconds=900)
```
anshikg, line 67: this change looks good. I am curious, how will we handler contract test run for resources which take time to create/update.

wbingli: It actually will re-invoke the handler if the response is IN_PROGRESS: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L506

src/rpdk/core/contract/resource_client.py
```diff
@@ -117,6 +117,7 @@ def __init__(
     ):  # pylint: disable=too-many-arguments
         self._schema = schema
         self._session = create_sdk_session(region)
+        self._role_arn = role_arn
         self._creds = get_temporary_credentials(
```
jotompki, line 121: is this _creds member variable needed if we are calling get_temporary_credentials every time? if not we may just be able to remove it.

wbingli: Good catch. It will be used below to get the account ID once. I will remove it as instance member var.

PatMyron (approved):

src/rpdk/core/contract/resource_client.py
PatMyron: some tests failing from this:
https://github.com/aws-cloudformation/cloudformation-cli/blob/88a4d8762d855e5d9e10609915fe93d41fe21be8/tests/contract/test_resource_client.py#L85
https://github.com/aws-cloudformation/cloudformation-cli/blob/88a4d8762d855e5d9e10609915fe93d41fe21be8/tests/contract/test_resource_client.py#L127

jotompki (approved):

src/rpdk/core/contract/resource_client.py
wbingli: Pretty funny, my local `pytest` works fine.

wbingli: OK. realize I need to do `pip3 install .` before the run the `pytest`. Am I doing something wrong?  can I run the `pytest` again the code in local workspace instead of installed package?

Issue #626, opened 2020-11-12
Update Contract test input payload should not have values from Create Output. Assertion failures for AWS::Redshift::Cluster 


shwetayakkali:
Facing issue with update contract tests. After a read on successful resource creation, the input parameters being passed to the Update Handler are expected to have a null value (if a parameter is not present in input payload to update Handler) but looks like the create handler output is combined with the input payload.

Example : create returns params {A = "a", B= "b", C="c")
input to update = {A="a", B= "bb"}

input being passed to Update Handler = {A="a", B= "bb, C="c"}

expected input passed to Update Handler= {A="a", B= "bb, C="null"}

anshikg:
Hi shwetayakkali,
This is expected behavior in update. Because, we dont have a good mechanism to fetch the ReadOnly properties in update we merged the created resource with the inputs_update.json file. This allows us to fetch this property. Is this causing contract test failures?

prerna-p:
Why is C="c" being set at all?
Update test will first create a resource and then update it with the properties from inputs_#_update.json.

anshikg:
> Why is C="c" being set at all?
> Update test will first create a resource and then update it with the properties from inputs_#_update.json.

thats not true Prerna. We merge the created model with the inputs_#_update.json file: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L271

prerna-p:
> > Why is C="c" being set at all?
> > Update test will first create a resource and then update it with the properties from inputs_#_update.json.
> 
> thats not true Prerna. We merge the created model with the inputs_#_update.json file: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L271

my confusion is why C="c" is passed in inputs_#_ create.json if it is required as null for the update call.

shwetayakkali:
> Hi shwetayakkali,
> This is expected behavior in update. Because, we dont have a good mechanism to fetch the ReadOnly properties in update we merged the created resource with the inputs_update.json file. This allows us to fetch this property. Is this causing contract test failures?

Yes, I am seeing failures in update contract tests due to this.

shwetayakkali:
> > > Why is C="c" being set at all?
> > > Update test will first create a resource and then update it with the properties from inputs_#_update.json.
> > 
> > 
[...]

PR #632, opened 2020-11-18
Fix to copy only unique identifiers into update example

prerna-p:
*Issue #, if available:* #626 
*Description of changes:*
Problem: 
payload for update request was created by merging all properties from the resourceModel** and inputs_#_update
Fix: 
- create the payload by merging only unique identifiers i.e primaryIdentifiers and additionalIdentifiers from the resourceModel into inputs_#_update
- check that createOnly primaryIdentifier(s) are not changed between the resourceModel and inputs_#_update

**resourceModel is model returned after a successful create.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -267,9 +267,39 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         example = self.invalid_strategy.example()
         return override_properties(example, self._overrides.get("CREATE", {}))
 
+    def get_unique_keys_for_model(self, create_model):
+        return {
+            k: v
+            for k, v in create_model.items()
+            if self.is_property_in_path(k, self.primary_identifier_paths)
+            or self.is_property_in_path(k, self._additional_identifiers_paths)
+        }
+
+    @staticmethod
+    def is_property_in_path(key, paths):
+        for path in paths:
+            prop = fragment_list(path, "properties")[0]
+            if prop == key:
+                return True
+        return False
+
+    def validate_update_example_keys(self, unique_identifiers, update_example):
+        for primary_identifier in self.primary_identifier_paths:
+            if primary_identifier in self.create_only_paths:
+                primary_key = fragment_list(primary_identifier, "properties")[0]
+                assert update_example[primary_key] == unique_identifiers[primary_key], (
+                    "Any createOnlyProperties specified in update handler input "
+                    "MUST NOT be different from their previous state"
+                )
+
     def generate_update_example(self, create_model):
         if self._inputs:
-            return {**create_model, **self._inputs["UPDATE"]}
+            unique_identifiers = self.get_unique_keys_for_model(create_model)
+            update_example = self._inputs["UPDATE"]
+            if self.create_only_paths:
+                self.validate_update_example_keys(unique_identifiers, update_example)
+            update_example.update(unique_identifiers)
+            return update_example
         overrides = self._overrides.get("UPDATE", self._overrides.get("CREATE", {}))
         example = override_properties(self.update_strategy.example(), overrides)
         return {**create_model, **example}
--- a/src/rpdk/core/contract/suite/handler_update.py
+++ b/src/rpdk/core/contract/suite/handler_update.py
@@ -16,7 +16,8 @@ from rpdk.core.contract.suite.handler_commons import (
 
 @pytest.fixture(scope="module")
 def updated_resource(resource_client):
-    create_request = input_model = model = resource_client.generate_create_example()
+    create_request = input_model = resource_client.generate_create_example()
+    model = {}
     try:
         _status, response, _error = resource_client.call_and_assert(
             Action.CREATE, OperationStatus.SUCCESS, create_request
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -54,6 +54,31 @@ SCHEMA_WITH_MULTIPLE_WRITE_PROPERTIES = {
     "writeOnlyProperties": ["/properties/d", "/properties/a"],
 }
 
+SCHEMA_ = {
+    "properties": {
+        "a": {"type": "number"},
+        "b": {"type": "number"},
+        "c": {"type": "number"},
+        "d": {"type": "number"},
+    },
+    "readOnlyProperties": ["/properties/b"],
+    "createOnlyProperties": ["/properties/c"],
+    "primaryIdentifier": ["/properties/c"],
+    "writeOnlyProperties": ["/properties/d"],
+}
+
+SCHEMA_WITH_COMPOSITE_KEY = {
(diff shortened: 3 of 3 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+    def get_unique_keys_for_model(self, create_model):
+        return {
+            k: v
+            for k, v in create_model.items()
+            if self.is_property_in_path(k, self.primary_identifier_paths)
+            or self.is_property_in_path(k, self._additional_identifiers_paths)
+        }
+
+    # Disable the no-self-use violations in this function
+    # pylint: disable=no-self-use
```
ammokhov, line 279: you should use `@staticmethod` if you dont need `self` reference

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+                    return True
+        return False
+
+    def validate_update_example_keys(self, unique_identifiers, update_example):
+        for primary_identifier in self.primary_identifier_paths:
+            if primary_identifier in self.create_only_paths:
+                primary_key = fragment_list(primary_identifier, "properties")[0]
+                assert update_example[primary_key] == unique_identifiers[primary_key], (
+                    "primaryIdentifier at the time of update must match"
+                    " primaryIdentifier used at the time of create"
```
ammokhov, line 293: this is what we have in handler contract [`Any createOnlyProperties specified in update handler input MUST NOT be different from their previous state`](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test-contract.html#resource-type-test-contract-update)

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+        for path in paths:
+            for key in path:
+                if prop == key:
+                    return True
+        return False
+
+    def validate_update_example_keys(self, unique_identifiers, update_example):
+        for primary_identifier in self.primary_identifier_paths:
+            if primary_identifier in self.create_only_paths:
+                primary_key = fragment_list(primary_identifier, "properties")[0]
```
ammokhov, line 290: why index `0`?

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+                    "primaryIdentifier at the time of update must match"
+                    " primaryIdentifier used at the time of create"
+                )
+
     def generate_update_example(self, create_model):
         if self._inputs:
-            return {**create_model, **self._inputs["UPDATE"]}
+            unique_identifiers = self.get_unique_keys_for_model(create_model)
+            update_example = self._inputs["UPDATE"]
+            if self.create_only_paths:
```
ammokhov, line 300: i think we have this [check](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/contract_asserts.py#L124)

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+    # Disable the no-self-use violations in this function
+    # pylint: disable=no-self-use
+    def is_property_in_path(self, prop, paths):
+        for path in paths:
+            for key in path:
+                if prop == key:
+                    return True
+        return False
+
+    def validate_update_example_keys(self, unique_identifiers, update_example):
```
ammokhov, line 287: might be a good idea to use it as a `@decorator` - [lib](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/contract_asserts.py)

ammokhov (changes requested):

src/rpdk/core/contract/resource_client.py
prerna-p: this is because of the weird construction of self.primary_identifier_paths. Same code is already in use [here](https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L393)

src/rpdk/core/contract/resource_client.py
prerna-p: I think that check is relevant where you want to know if all primary keys are createOnly. Here, even if one primary key is createOnly we want to check its value hasn't changed

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,40 @@ def generate_invalid_create_example(self):
...
+            k: v
+            for k, v in create_model.items()
+            if self.is_property_in_path(k, self.primary_identifier_paths)
+            or self.is_property_in_path(k, self._additional_identifiers_paths)
+        }
+
+    # Disable the no-self-use violations in this function
+    # pylint: disable=no-self-use
+    def is_property_in_path(self, prop, paths):
+        for path in paths:
```
jotompki, line 281: suggest you change this to a list comprehension of some kind. also the `in` boolean operator will save you a loop.

```
return any(prop in path for path in paths)
```

src/rpdk/core/contract/resource_client.py
```diff
@@ -267,9 +267,39 @@ def generate_invalid_create_example(self):
         example = self.invalid_strategy.example()
         return override_properties(example, self._overrides.get("CREATE", {}))
 
+    def get_unique_keys_for_model(self, create_model):
+        return {
+            k: v
+            for k, v in create_model.items()
+            if self.is_property_in_path(k, self.primary_identifier_paths)
```
jotompki, line 274: does this work for nested properties? say we have a schema 
```
{
    "properties": {
        "a": {"type": "number"},
        "b": {"type": "number"},
        "c": {"type": "object",
                "properties": {
                     "e": {"type": "string"}
                 }
               },
        "d": {"type": "number"},
    },
    "readOnlyProperties": ["/properties/b"],
    "createOnlyProperties": ["/properties/c"],
    "primaryIdentifier": ["/properties/c/e"],
    "writeOnlyProperties": ["/properties/d"],
}
```

Will we be able to copy over the c/e property? it seems like we only do a check on the resource model's top level keys. It seems like it would currently include the entire c object and not just the c/e key

prerna-p: That is a good catch and indeed, it will not work correctly in that case. In my opinion, this is an issue due to the construction of primary_identifier_paths rather than the code this PR is addressing. It is something we should log as a bug to fix, possibly across all  <value>_identifier_paths we are creating.

ammokhov (approved):

jotompki (approved):

prerna-p:
Tested with the resource that was reporting error (tests are passing):
https://paste.amazon.com/show/prerp/1609277600

PR #634, opened 2020-11-24
Add support for Modules

MalikAtalla-AWS:
This release adds support for creating Modules using the CFN CLI
 * Initialize a Module project with `cfn init --artifact-type MODULE --type-name <type-name>`
 * For example: `cfn init --artifact-type MODULE --type-name My::Test::Example::MODULE`
 * Using `cfn init` without parameters will now first ask if you want to create a resource or a module. If you choose to create a module, it will create an example module template under fragments/sample.json which you can use as a starting point to make your changes.
 * Use `cfn validate` to validate your module before submitting
 * Use `cfn submit` to register your module, making it available for use by CloudFormation in your AWS account
 * Find detailed documentation about creating Modules [here](docs.aws.amazon.com/cloudformation-cli/latest/userguide/modules.html)

 Backwards-incompatible changes:
  * To use `cfn init` without going into interactive mode, you now need to specify the artifact type. So, to initialize a resource project, you would use
    * `cfn init --artifact-type RESOURCE --type-name <type-name>`
    * or for short: `cfn init -a r -t <type-name>`

*Description of changes:*

### DIFF
--- a/src/rpdk/core/__init__.py
+++ b/src/rpdk/core/__init__.py
@@ -1,5 +1,5 @@
 import logging
 
-__version__ = "0.1.14"
+__version__ = "0.2.0"
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
--- a/src/rpdk/core/fragment/generator.py
+++ b/src/rpdk/core/fragment/generator.py
@@ -1,0 +1,347 @@
+"""
+This class has two responsibilities:
+1. generating a sample template fragment so the user has some initial
+file fragment as an example.
+The method "generate_sample_fragment" will be called as part of the init command.
+2. generating schema for provided template fragments.
+The method "generate_schema" will be called right before submission.
+"""
+import json
+import logging
+import os
+from pathlib import Path
+
+import yaml
+
+from rpdk.core.data_loaders import resource_json
+from rpdk.core.exceptions import FragmentValidationError
+
+LOG = logging.getLogger(__name__)
+FRAGMENT_DIR = "fragments"
+SAMPLE_FRAGMENT_OUTPUT = "sample.json"
+SCHEMA_NAME = "schema.json"
+SAMPLE_FRAGMENT = "../data/examples/module/sample.json"
+ALLOWED_EXTENSIONS = {".json", ".yaml", ".yml"}
+RESOURCE_LIMIT = 500
+OUTPUT_LIMIT = 200
+MAPPING_LIMIT = 200
+MAPPING_ATTRIBUTE_LIMIT = 200
+TEMPLATE_FILE_SIZE_IN_BYTES_LIMIT = 1500000
+
+
+class TemplateFragment:  # pylint: disable=too-many-instance-attributes
+    def __init__(self, type_name, root=None):
+        self.root = Path(root) if root else Path.cwd()
+        self.fragment_dir = self.root / FRAGMENT_DIR
+        self.type_name = type_name
+        self.resource_limit = RESOURCE_LIMIT
+        self.output_limit = OUTPUT_LIMIT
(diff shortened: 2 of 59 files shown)


src/rpdk/core/__init__.py
```diff
@@ -1,5 +1,5 @@
 import logging
 
-__version__ = "0.1.14"
+__version__ = "0.2.0"
```
jotompki, line 3: https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/setup.py#L39 important to note this. might need to bump this

miparnisari (approved):

wbingli (approved):

jotompki (approved):

src/rpdk/core/project.py
```diff
@@ -346,15 +414,34 @@ def load(self):
...
+            self.schema = template_fragment.generate_schema()
+            self.fragment_dir = template_fragment.fragment_dir
+        else:
+            LOG.info("Validating your resource specification...")
+            try:
+                self.load_schema()
+            except FileNotFoundError as e:
+                self._raise_invalid_project("Resource specification not found.", e)
+            except SpecValidationError as e:
+                msg = "Resource specification is invalid: " + str(e)
```
PatMyron, line 434: partially reverted https://github.com/aws-cloudformation/cloudformation-cli/pull/425

src/rpdk/core/fragment/generator.py
```diff
@@ -0,0 +1,347 @@
...
+FRAGMENT_DIR = "fragments"
+SAMPLE_FRAGMENT_OUTPUT = "sample.json"
+SCHEMA_NAME = "schema.json"
+SAMPLE_FRAGMENT = "../data/examples/module/sample.json"
+ALLOWED_EXTENSIONS = {".json", ".yaml", ".yml"}
+RESOURCE_LIMIT = 500
+OUTPUT_LIMIT = 200
+MAPPING_LIMIT = 200
+MAPPING_ATTRIBUTE_LIMIT = 200
+TEMPLATE_FILE_SIZE_IN_BYTES_LIMIT = 1500000
```
PatMyron, line 29: Isn't this over [CloudFormation's template size limit](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/cloudformation-limits.html)?

PatMyron:
>Backwards-incompatible changes:
> * To use `cfn init` without going into interactive mode, you now need to specify the artifact type

Couldn't `--artifact-type` default to `RESOURCE`? Main benefit of [non-interactive mode](https://github.com/aws-cloudformation/cloudformation-cli/issues/410) is a stable interface to automate against, so we should keep it backwards-compatible

tests/test_project.py
```diff
@@ -539,6 +643,29 @@ def test_load_invalid_schema(project):
...
 
 
+def test_load_module_project_succeeds(project):
+    project.artifact_type = "MODULE"
+    project.type_name = "Unit::Test::Malik::MODULE"
+    patch_load_settings = patch.object(
+        project, "load_settings", return_value={"artifact_type": "MODULE"}
+    )
+    with patch_load_settings:
+        project.load()
```
PatMyron, line 653: This test leaks a `schema.json` file in the root directory

src/rpdk/core/fragment/generator.py
PatMyron: Also think a lot of this validation this isn't specific to modules can be removed since [`cfn-lint` is being called now](https://github.com/aws-cloudformation/cloudformation-cli/pull/644)

tests/test_project.py
MalikAtalla-AWS: This has been fixed since, btw

PR #637, opened 2020-11-25
Fix flaky unit test

MalikAtalla-AWS:
The unit test in question was depending on a specific folder to exist. There are other tests which will create that folder, but I suspect the tests are executed in a deterministic order. Fixed this by changing the test to use a shared method which includes the creation of said folder.

### DIFF
--- a/tests/fragments/test_generator.py
+++ b/tests/fragments/test_generator.py
@@ -14,23 +14,15 @@ TIMEOUT_IN_SECONDS = 10
 
 directory = os.path.dirname(__file__)
 
-test_root = "build"
-
 
 @pytest.fixture
-def template_fragment():
-    return TemplateFragment(type_name, test_root)
+def template_fragment(tmpdir):
+    return TemplateFragment(type_name, tmpdir)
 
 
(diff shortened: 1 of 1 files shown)


tests/fragments/test_generator.py
```diff
@@ -23,13 +23,7 @@ def template_fragment():
```
jotompki, line 23: i think a better approach is to add a fixture that makes a temporary directory in the [tmpdir](https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-fixture). Tests that need it can just rely on the fixture and it will get cleaned up afterwards.

jotompki: [tmpdir factory might be better](https://docs.pytest.org/en/stable/tmpdir.html#the-tmpdir-factory-fixture)

MalikAtalla-AWS: Makes sense. Addressed.

jotompki (approved):

miparnisari (approved):

PR #642, opened 2020-12-04
Add additional checks on structured properties to not leak the additional properties to handlers during validation

srujithpoondla03:
*Description of changes:*
Currently, in the provider definition schema we do not restrict the structured property to specify the additionalProperties as false. Structured properties should define the additionalProperties as false otherwise properties with different case or pattern could not be caught during validation.

Tested by running the `cfn validate`
```
⋊> ~/c/NrqlAlertCondition on master ⨯ cfn validate                                                                                                                                                       
[Warning] Resource spec validation would fail from next major version. Provider should mark additionalProperties as false if the property is defined as a structured property. Please fix the warnings: 'additionalProperties' is a required property

Failed validating 'required' in schema['properties']['definitions']['patternProperties']['^[A-Za-z0-9]{1,64}$']['allOf'][0]['dependencies']['properties']:
    {'$comment': 'An object cannot have both defined and undefined '
                 'properties; therefore, patternProperties is not allowed '
                 'when properties is specified.',
     'not': {'required': ['patternProperties']},
     'required': ['additionalProperties']}

On instance['definitions']['Term']:
```

### DIFF
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -80,7 +80,10 @@ def make_resource_validator_with_additional_properties_check():
     properties_check = {
         "properties": {
             "$comment": "An object cannot have both defined and undefined \
-properties; therefore, patternProperties is not allowed when properties is specified.",
+properties; therefore, patternProperties is not allowed when properties is specified.\
+ Provider should mark additionalProperties as false if the \
+property is of object type and has properties defined \
+in it.",
             "not": {"required": ["patternProperties"]},
             "required": ["additionalProperties"],
         }
@@ -88,7 +91,9 @@ properties; therefore, patternProperties is not allowed when properties is speci
     pattern_properties_check = {
         "patternProperties": {
             "$comment": "An object cannot have both defined and undefined \
-properties; therefore, properties is not allowed when patternProperties is specified.",
+properties; therefore, properties is not allowed when patternProperties is specified. \
+Provider should mark additionalProperties as false if the property is of object type \
+and has patternProperties defined in it.",
             "not": {"required": ["properties"]},
             "required": ["additionalProperties"],
         }
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -93,9 +93,11 @@ properties; therefore, properties is not allowed when patternProperties is speci
             "required": ["additionalProperties"],
         }
     }
-    dependencies.update(properties_check)
-    dependencies.update(pattern_properties_check)
-    schema["definitions"]["validations"]["dependencies"] = dependencies
+    schema["definitions"]["validations"]["dependencies"] = {
+        **dependencies,
+        **properties_check,
+        **pattern_properties_check,
+    }
     return make_validator(schema)
 
 
@@ -147,7 +149,8 @@ def load_resource_spec(resource_spec_file):  # noqa: C901
         LOG.warning(
             "[Warning] Resource spec validation would fail from next \
 major version. Provider should mark additionalProperties as false if the \
-property is defined as a structured property. Please fix the warnings: %s",
+property is of object type and has properties or patternProperties defined \
+in it. Please fix the warnings: %s",
             str(e),
         )
     in_readonly = _is_in(resource_spec, "readOnlyProperties")
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -74,6 +74,33 @@ def make_resource_validator():
     return make_validator(schema)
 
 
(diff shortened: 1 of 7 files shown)


src/rpdk/core/data_loaders.py
```diff
@@ -74,6 +74,31 @@ def make_resource_validator():
...
+        "patternProperties": {
+            "$comment": "An object cannot have both defined and undefined \
+properties; therefore, properties is not allowed when patternProperties is specified.",
+            "not": {"required": ["properties"]},
+            "required": ["additionalProperties"],
+        }
+    }
+    dependencies.update(properties_check)
+    dependencies.update(pattern_properties_check)
+    schema["definitions"]["validations"]["dependencies"] = dependencies
```
ammokhov, line 98: can u just use 
```python
dep = schema["definitions"]["validations"]["dependencies"]
schema["definitions"]["validations"]["dependencies"] = {**dep, ** properties_check, **pattern_prop_check}
```

priyap286:
Do we need to define what a structured property is? Also "additionalProperties" are a property of object schemas and are **required** only there.

src/rpdk/core/data_loaders.py
srujithpoondla03: Updated

src/rpdk/core/data_loaders.py
```diff
@@ -74,6 +74,33 @@ def make_resource_validator():
     return make_validator(schema)
 
 
+def make_resource_validator_with_additional_properties_check():
```
priyap286, line 77: We must also make sure that resource owners do not specify `additionalProperties` when `properties`/`patternProperties` are not specified.

sungkkim (approved):

src/rpdk/core/data_loaders.py
srujithpoondla03: Will handle this in a different PR

priyap286 (approved):

src/rpdk/core/data_loaders.py
```diff
@@ -74,6 +74,33 @@ def make_resource_validator():
...
+    schema = resource_json(__name__, "data/schema/provider.definition.schema.v1.json")
+    dependencies = schema["definitions"]["validations"]["dependencies"]
+    properties_check = {
+        "properties": {
+            "$comment": "An object cannot have both defined and undefined \
+properties; therefore, patternProperties is not allowed when properties is specified.",
+            "not": {"required": ["patternProperties"]},
```
ammokhov, line 84: is `not` supported?

priyap286: "`properties` and `patternProperties` it is not valid to use both properties and patternProperties together in the same shape, as a shape should not contain both defined and undefined values. In order to implement this, the set of undefined values should itself be a subshape."
From the ReadMe of [aws-cloudformation-resource-schema
](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema)

srujithpoondla03: Yeah discussed with @ammokhov. I will add couple more unit tests to test the case where providers set "additionalProperties" as true. We should not allow this as well.

ammokhov (approved):
lgtm

PatMyron:
seems explicitly documented that `additionalProperties` should not be used:

["use of `additionalProperties` is not valid for a resource property"](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#constraints)

Should try to check why that was documented and update that documentation if that's no longer the case

srujithpoondla03:
Use of `additionalProperties` is not valid means provider should not model their schema to accept additionalProperties instead they should define all the properties or use pattern properties.

PatMyron:
@srujithpoondla03 got it, maybe just re-worded to clarify then?

src/rpdk/core/data_loaders.py
```diff
@@ -107,12 +139,25 @@ def load_resource_spec(resource_spec_file):  # noqa: C901
...
 
+    try:
+        additional_properties_validator.validate(resource_spec)
+    except ValidationError as e:
+        LOG.warning(
+            "[Warning] Resource spec validation would fail from next \
+major version. Provider should mark additionalProperties as false if the \
+property is of object type and has properties or patternProperties defined \
+in it. Please fix the warnings: %s",
+            str(e),
```
PatMyron, line 159: This generates thousands of lines of output [when run across all existing resource schemas](https://github.com/aws-cloudformation/cloudformation-cli/pull/604#issuecomment-756521889). Wondering if just outputting the `message` (`'additionalProperties' is a required property`) instead would be less noisy:
```suggestion
            e.message,
```

PR #644, opened 2020-12-09
Call cfn-lint for module fragments

miparnisari:
*Issue #, if available:* N/A

*Description of changes:* when calling `cfn validate` on a module, run `cfn-lint` on the fragment for extra validations.

(Future improvement: support `cfn validate --region` so that we can pass the `region` value to `cfn-lint`. Related: https://github.com/aws-cloudformation/cfn-python-lint/issues/607)

Sample output:

```bash
MacBook@ ~/Desktop/modules/cfnlint $ cat fragments/sample.json
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Resources": {
        "S3Bucket": {
            "Type": "AWS::S3::Bucket",
            "DeletionPolicy": "Something",
            "UpdateReplacePolicy": "Retain"
        }
    }
}
MacBook@ ~/Desktop/modules/cfnlint $ cfn validate
Module fragment is probably valid, but there are warnings/errors from cfn-lint (https://github.com/aws-cloudformation/cfn-python-lint):
	DeletionPolicy should be only one of Delete, Retain, Snapshot at Resources/S3Bucket/DeletionPolicy (from rule E3035: Check DeletionPolicy values for Resources)
MacBook@ ~/Desktop/modules/cfnlint $ cat fragments/sample.json
{
    "AWSTemplateFormatVersion": "2010-09-09",
    "Resources": {
        "S3Bucket": {
            "Type": "AWS::S3::Bucket",
            "DeletionPolicy": "Retain",
            "UpdateReplacePolicy": "Retain"
        }
    }
}
MacBook@ ~/Desktop/modules/cfnlint $ cfn validate
Module fragment is valid.
```

---

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,9 +57,9 @@ cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 
 ### Command: validate
 
-To validate the auto-generated code, use the `validate` command.
+To validate the schema, use the `validate` command.
 
-This command is automatically run whenever one attempts to submit a resource or module. Any module fragments will be automatically validated via `cfn-lint`, however any warnings or errors detected by `cfn-lint` will not cause this step to fail.
+This command is automatically run whenever one attempts to submit a resource or module. Any module fragments will be automatically validated via [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/), however any warnings or errors detected by [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/) will not cause this step to fail.
 
 ```bash
 cfn validate
--- a/README.md
+++ b/README.md
@@ -55,6 +55,16 @@ cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD 
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 ```
 
+### Command: validate
+
+To validate the auto-generated code, use the `validate` command.
+
+This command is automatically run whenever one attempts to submit a resource or module. Any module fragments will be automatically validated via `cfn-lint`, however any warnings or errors detected by `cfn-lint` will not cause this step to fail.
+
+```bash
+cfn validate
+```
+
 ### Command: build-image
 
 To build an image for a resource type. This image provides a minimalistic execution environment for the resource handler that does not depend on AWS Lambda in anyway. This image can be used during cfn invoke and cfn test instead of using sam cli.
--- a/setup.py
+++ b/setup.py
@@ -51,7 +51,7 @@ setup(
         "colorama>=0.4.1",
         "docker>=4.3.1",
         "ordered-set>=4.0.2",
-        "cfn-lint>=0.35.0",
+        "cfn-lint>=0.43.0",
     ],
     entry_points={
         "console_scripts": ["cfn-cli = rpdk.core.cli:main", "cfn = rpdk.core.cli:main"]
--- a/setup.py
+++ b/setup.py
@@ -51,6 +51,7 @@ setup(
         "colorama>=0.4.1",
         "docker>=4.3.1",
         "ordered-set>=4.0.2",
+        "cfn-lint>=0.35.0",
     ],
     entry_points={
         "console_scripts": ["cfn-cli = rpdk.core.cli:main", "cfn = rpdk.core.cli:main"]
--- a/src/rpdk/core/fragment/generator.py
+++ b/src/rpdk/core/fragment/generator.py
@@ -86,6 +86,10 @@ class TemplateFragment:  # pylint: disable=too-many-instance-attributes
                 "warnings/errors from cfn-lint "
                 "(https://github.com/aws-cloudformation/cfn-python-lint):"
             )
+            self.__print_cfn_lint_warnings(lint_warnings)
+
+    @staticmethod
+    def __print_cfn_lint_warnings(lint_warnings):
         for lint_warning in lint_warnings:
             print(
                 "\t{} (from rule {})".format(lint_warning.message, lint_warning.rule),
--- a/src/rpdk/core/fragment/generator.py
+++ b/src/rpdk/core/fragment/generator.py
@@ -11,6 +11,8 @@ import logging
 import os
 from pathlib import Path
 
+import cfnlint.config
+import cfnlint.core
 import yaml
 
 from rpdk.core.data_loaders import resource_json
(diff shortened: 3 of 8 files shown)


tests/data/sample_fragments/template_without_parameter_section.json
```diff
@@ -5,6 +5,7 @@
         "S3Bucket": {
             "Type": "AWS::S3::Bucket",
             "DeletionPolicy": "Retain",
+            "UpdateReplacePolicy": "Retain",
```
miparnisari, line 8: This change is needed to bring coverage to 100% and not have to write an extra test

setup.py
```diff
@@ -51,6 +51,7 @@ def find_version(*file_paths):
         "colorama>=0.4.1",
         "docker>=4.3.1",
         "ordered-set>=4.0.2",
+        "cfn-lint>=0.35.0",
```
PatMyron, line 54: any particular reason for this version btw? Thinking we want the latest version since older versions bundle outdated static data (assuming [`0.35.0`](https://github.com/aws-cloudformation/cfn-python-lint/releases/tag/v0.35.0) was just the latest version when this work started since nothing too notable about that release)
```suggestion
        "cfn-lint>=0.43.0",
```

README.md
```diff
@@ -55,6 +55,16 @@ cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 ```
 
+### Command: validate
+
+To validate the auto-generated code, use the `validate` command.
+
+This command is automatically run whenever one attempts to submit a resource or module. Any module fragments will be automatically validated via `cfn-lint`, however any warnings or errors detected by `cfn-lint` will not cause this step to fail.
```
PatMyron, line 62:
```suggestion
This command is automatically run whenever one attempts to submit a resource or module. Any module fragments will be automatically validated via [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/), however any warnings or errors detected by [`cfn-lint`](https://github.com/aws-cloudformation/cfn-python-lint/) will not cause this step to fail.
```

src/rpdk/core/fragment/generator.py
```diff
@@ -75,11 +77,43 @@ def validate_fragments(self):
         self.__validate_no_transforms_present(raw_fragments)
         self.__validate_outputs(raw_fragments)
         self.__validate_mappings(raw_fragments)
+        lint_warnings = self.__validate_fragment_through_cfn_lint(raw_fragments)
+        if not lint_warnings:
+            LOG.warning("Module fragment is valid.")
+        else:
+            LOG.warning(
+                "Module fragment is valid, but there are warnings from cfn-lint "
```
PatMyron, line 85: 1) We don't really know the module fragment is valid yet, right?

2) Might be worth distinguishing terminology from one of [`cfn-lint`s rule levels](https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#rule-levels)?
```suggestion
                "Module fragment might be valid, but there are warnings from cfn-lint "
```

PatMyron:
> Future improvement: support `cfn validate --region` so that we can pass the `region` value to `cfn-lint`

Other options like [`--registry-schemas`](https://github.com/aws-cloudformation/cfn-python-lint/pull/1732) come to mind as well

README.md
```diff
@@ -55,6 +55,16 @@ cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
 ```
 
+### Command: validate
+
+To validate the auto-generated code, use the `validate` command.
```
jotompki, line 60: we really are validating the schema with this command

src/rpdk/core/fragment/generator.py
jotompki: will this only appear if the user running `validate` specifies increased verbosity?

src/rpdk/core/fragment/generator.py
```diff
@@ -75,11 +77,43 @@ def validate_fragments(self):
...
         self.__validate_output_limit(raw_fragments)
 
+    @staticmethod
+    def __validate_fragment_through_cfn_lint(raw_fragment):
+        filename = "temporary_fragment.json"
+
+        with open(filename, "w") as outfile:
+            json.dump(raw_fragment, outfile, indent=4)
+
+        template = cfnlint.decode.cfn_json.load(filename)
```
jotompki, line 104: Can we just use the yaml loads() so we don't have to write to a temporary file: https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/decode/cfn_yaml.py#L183?

src/rpdk/core/fragment/generator.py
miparnisari: @PatMyron right. I'll edit both.

@jotompki no, for modules this runs all the time.

src/rpdk/core/fragment/generator.py
miparnisari: The problem is that `cfnlint.core.run_checks` needs a `filename` to run. I haven't seen an api that validates a template in memory.

src/rpdk/core/fragment/generator.py
```diff
@@ -75,11 +77,45 @@ def validate_fragments(self):
...
+        else:
+            LOG.warning(
+                "Module fragment is probably valid, but there are "
+                "warnings/errors from cfn-lint "
+                "(https://github.com/aws-cloudformation/cfn-python-lint):"
+            )
+        for lint_warning in lint_warnings:
+            print(
+                "\t{} (from rule {})".format(lint_warning.message, lint_warning.rule),
+            )
```
MalikAtalla-AWS, line 92: Would you mind wrapping this in a new method? Something, like print `cfn_lint_warnings`

src/rpdk/core/fragment/generator.py
```diff
@@ -75,11 +77,45 @@ def validate_fragments(self):
...
+        with open(filename, "w") as outfile:
+            json.dump(raw_fragment, outfile, indent=4)
+
+        template = cfnlint.decode.cfn_json.load(filename)
+
+        # Initialize the ruleset to be applied (no overrules, no excludes)
+        # Runs Warning and Error rules
+        rules = cfnlint.core.get_rules([], [], [], [], False, [])
+
+        regions = ["us-east-1"]
```
MalikAtalla-AWS, line 111: Should this region be hardcoded? What's the effect of this?

PatMyron: `us-east-1` is the default for `cfn-lint` since it's close to a superset for AWS types

miparnisari: Source: https://github.com/aws-cloudformation/cfn-python-lint/blob/master/src/cfnlint/config.py#L519

miparnisari: I'm unsure whether there is a lot of benefit in having RPDK `validate` command receive a `region` parameter and passing this to cfn-lint. John also seemed to suggest above that we could make this whole validation optional. I don't want to make the `validate` CLI too complicated but i'm open to suggestions.

src/rpdk/core/fragment/generator.py
MalikAtalla-AWS: Sorry, I wasn't super clear. I actually meant the entire thing. This way the validate_fragments method will be really short and concise and just have a list of other validate calls. Don't want to block the PR though.

MalikAtalla-AWS (approved):

src/rpdk/core/fragment/generator.py
MalikAtalla-AWS: Thanks for the clarification. Makes sense to me.

src/rpdk/core/fragment/generator.py
jotompki: is filename being used anywhere? 

if it's really necessary recommend: https://docs.python.org/3/library/tempfile.html

miparnisari: @jotompki yep, it's used by ` cfnlint.core.run_checks`. I tried using `NamedTemporaryFile`, I don't see the benefit, I have to write the file, then read it back, and then manually remove it?

src/rpdk/core/fragment/generator.py
miparnisari: refactored!

MalikAtalla-AWS (approved):

kddejong:
This is great to see.  One of the things that may be worth adding as a rule (Warning level) is using parameters in conditions inside your module.  Not sure if you would want to add custom rules but I got tripped up on this.

miparnisari:
> This is great to see. One of the things that may be worth adding as a rule (Warning level) is using parameters in conditions inside your module. Not sure if you would want to add custom rules but I got tripped up on this.

@kddejong this can be added as an extra validation in RPDK, right? @MalikAtalla-AWS has context on all the validations were are currently doing.

PatMyron (approved):

PR #647, opened 2020-12-12
automating CloudFormation Resource Provider Definition MetaSchema updates with Github actions

PatMyron:
[tested in my fork](https://github.com/PatMyron/cloudformation-cli/pull/1/files)

currently picks up https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/84 and reverts https://github.com/aws-cloudformation/cloudformation-cli/pull/612 unless it's added there too

### DIFF
--- a/.github/workflows/schema-updater.yaml
+++ b/.github/workflows/schema-updater.yaml
@@ -1,0 +1,16 @@
+on:
+  schedule:
+    - cron: '*/5 * * * *'
+jobs:
+  job:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


aygold92 (approved):
LGTM. It would be nice if we could somehow get informative commit messages but not sure how we would do that.

anshikg (approved):

PatMyron:
> LGTM. It would be nice if we could somehow get informative commit messages but not sure how we would do that.

@aygold92 agreed, at least the commit messages can be edited while squashing and merging for now if people want

PR #667, opened 2021-01-15
Add property transform to contract test

anshikg:
*Issue #, if available:*

*Description of changes:* Add property Transform in contract test

*Test:*
```
platform darwin -- Python 3.8.5, pytest-6.0.1, py-1.9.0, pluggy-0.13.1 -- /Users/anshikg/workspace/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/anshikg/workspace/aws-cloudformation-resource-providers-cloudwatch/aws-cloudwatch-alarm/.hypothesis/examples')
rootdir: /Users/anshikg/workspace/aws-cloudformation-resource-providers-cloudwatch/aws-cloudwatch-alarm, configfile: ../../../../../private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_yok5l5i6.ini
plugins: random-order-1.0.4, localserver-0.5.0, cov-2.10.0, hypothesis-5.19.3, black-0.3.10
collected 16 items                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [  6%]
handler_create.py::contract_invalid_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                               [ 12%]
handler_create.py::contract_create_duplicate PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             [ 18%]
handler_create.py::contract_create_read_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 25%]
handler_create.py::contract_create_list_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 31%]
handler_delete.py::contract_delete_read PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  [ 37%]
handler_delete.py::contract_delete_list PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  [ 43%]
handler_delete.py::contract_delete_update PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 50%]
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 56%]
handler_delete.py::contract_delete_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                [ 62%]
handler_misc.py::contract_check_asserts_work PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             [ 68%]
handler_read.py::contract_read_without_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                            [ 75%]
handler_update.py::contract_update_read_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 81%]
handler_update.py::contract_update_list_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          [ 87%]
handler_update_invalid.py::contract_update_create_only_property PASSED                                                                                                                                                                                                                                                                                                                                                                                                                          [ 93%]
handler_update_invalid.py::contract_update_non_existent_resource PASSED   


```


.pre-commit-config.yaml
```diff
@@ -47,14 +47,14 @@ repos:
...
-    # have to skip B322, as there is no way to indicate the codebase is Python 3 only (input only vulnerable in Py2)
-    args: ["--skip", "B101,B322"]
+#- repo: https://github.com/PyCQA/bandit
+#  rev: f5a6f0ca62  # TODO: update once a release > 1.5.1 hits with this change in
+#  hooks:
+#  - id: bandit
+#    files: "^src/"
+#    # have to skip B101, contract tests use it and there's no way to skip for specific files
+#    # have to skip B322, as there is no way to indicate the codebase is Python 3 only (input only vulnerable in Py2)
+#    args: ["--skip", "B101,B322,B602,B607"]
```
anshikg, line 57: missed this. Will remove the comments

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -187,3 +195,57 @@ def test_unordered_list_match(inputs, outputs):
...
+
+def transform(property_path, input_model, property_transform):
+    path = set_to_string(property_path)
+    file = open("./transformation.js", "a")
+    file.write("var jsonata =require('jsonata')\n")
+    file.write("var model = {0};\n".format(input_model))
+    property_transform_value = property_transform[path].replace('"', '\\"')
+    file.write('var expression = jsonata("{0}");\n'.format(property_transform_value))
+    file.write("var result = expression.evaluate(model);\n")
+    file.write("console.log(result);\n")
```
PatMyron, line 223: Let's not write other programming languages inside other programming languages line-by-line. Let's just read  another `.js` file if needed so local tooling like autocompletion, syntax highlighting, static analysis, etc. works

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -187,3 +195,57 @@ def test_unordered_list_match(inputs, outputs):
...
+        LOG.debug("Override failed.\nPath %s\nDocument %s", property_path, input_model)
+        LOG.warning("Override with path %s not found, skipping", property_path)
+    else:
+        key = resolved_path[-1]
+        parent[key] = transformed_property
+
+
+def execute_js():
+    subprocess.getoutput("npm install jsonata")
+    jsonata_output = subprocess.getoutput("node transformation.js")
```
PatMyron, line 249: [Still think we should re-consider pulling in another programming language just for this](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/111#discussion_r555242773)

PatMyron (dismissed):

src/rpdk/core/contract/suite/handler_commons.py
jotompki: use the  [data directory](https://github.com/aws-cloudformation/cloudformation-cli/tree/master/src/rpdk/core/data) if we have anything we want to load in from the package

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -187,3 +195,57 @@ def test_unordered_list_match(inputs, outputs):
...
+    file.write("var model = {0};\n".format(input_model))
+    property_transform_value = property_transform[path].replace('"', '\\"')
+    file.write('var expression = jsonata("{0}");\n'.format(property_transform_value))
+    file.write("var result = expression.evaluate(model);\n")
+    file.write("console.log(result);\n")
+    file.close()
+    return execute_js()
+
+
+def set_to_string(property_path):
```
jotompki, line 228: just use string [join](https://docs.python.org/3/library/stdtypes.html#str.join)

src/rpdk/core/contract/suite/handler_commons.py
jotompki: Also could look into some evaluation lib: https://github.com/sony/v8eval

jotompki (commented):
are there any native libraries that handle jsonata format?

src/rpdk/core/contract/suite/handler_commons.py
anshikg: Still think we should re-consider pulling in another programming language just for this
> dont think this is a possibility now. We have had design conversations on this and the backend service is already implemented for jsonata

anshikg: Also could look into some evaluation lib: https://github.com/sony/v8eval\
> I looked into adding both v8eval & js2py. Both are running into issues when fetching jsonata into code. Right now, subprocess is the easiest way to include jsonata code. The other option is to write our own python wrapper for jsonata. We could consider doing that for the long run

src/rpdk/core/contract/suite/handler_commons.py
anshikg: good point. thanks

src/rpdk/core/contract/suite/handler_commons.py
anshikg: Let's not write other programming languages inside other programming languages line-by-line. Let's just read another .js file if needed so local tooling like autocompletion, syntax highlighting, static analysis, etc. works
> As of now jsonata python wrapper is not present for use. We could write our own python wrapper but, for unblocking this functionality this seems to be the quickest solution

anshikg: use the data directory if we have anything we want to load in from the package
> the ```transformation.js``` file is created in the local directory (which is the provider package). Is there some way to fetch the data directory information? May be I am missing something.

PatMyron: specifically asking to not use `file.write()` here. Unrelated to broader discussion of JavaScript usage

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -187,3 +195,57 @@ def test_unordered_list_match(inputs, outputs):
...
+    except LookupError:
+        LOG.debug("Override failed.\nPath %s\nDocument %s", property_path, input_model)
+        LOG.warning("Override with path %s not found, skipping", property_path)
+    else:
+        key = resolved_path[-1]
+        parent[key] = transformed_property
+
+
+def execute_js():
+    subprocess.getoutput("npm install jsonata")
```
jotompki, line 248: this will fail if the end customer doesn't have npm installed correct? we don't list this as a dependency. if that's the case we need some documentation on installing npm if people plan to run contract tests with tranformations.

jotompki: also should we be attempting to install on every transformation?

src/rpdk/core/contract/suite/handler_commons.py
jotompki: we can render this as a jinja template if planning on doing this. consider also using tmpfile as well so we aren't leaving files behind after execution.

jotompki (commented):
can we add some unit tests for some of this as well? i feel like the contract module has had a grey area in the past because we are writing tests, but i feel as some of this should be tested, especially the process calls out to npm

src/rpdk/core/contract/suite/handler_commons.py
anshikg: makes sense. We need npm installed. Will move this so that we are calling this once every test

anshikg:
> can we add some unit tests for some of this as well? i feel like the contract module has had a grey area in the past because we are writing tests, but i feel as some of this should be tested, especially the process calls out to npm

makes sense. Will add unit test for it.

anshikg:
we can render this as a jinja template if planning on doing this. consider also using tmpfile as well so we aren't leaving files behind after execution.
> Could you help me understand the advantage of using jinja templates. I am not sure I understand. Will look into using tmpfiles

src/rpdk/core/contract/suite/handler_commons.py
PatMyron: We also need to think about whether we want `jsonata` to be a versioned dependency instead of always just installing the latest version on the fly

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -187,3 +197,64 @@ def test_unordered_list_match(inputs, outputs):
...
+
+
+def transform_model(
+    input_model,
+    output_model,
+    property_transform_keys,
+    property_transform,
+    jinja_template,
+):
+    subprocess.getoutput("npm install jsonata")
```
PatMyron, line 209: This is still re-installing every run, right?
[Wondering if it's possible to install `jsonata` when `cloudformation-cli` itself is installed](https://stackoverflow.com/questions/49953785/which-options-do-exist-for-defining-a-python-package-with-node-js-dependencies)
If not, I think that's still my biggest concern with pulling in `npm` dependencies

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -168,6 +171,13 @@ def test_input_equals_output(resource_client, input_model, output_model):
     # only comparing properties in input model to those in output model and
     # ignoring extraneous properties that maybe present in output model.
     try:
+        transform_model(
```
jotompki, line 174: since this is so dependent on the resource client might it make sense to have it as a method on the client? could probably reduce the number of arguments then and get rid of transform keys, property transform, and jinja template

src/rpdk/core/contract/resource_client.py
```diff
@@ -168,12 +176,14 @@ def _update_schema(self, schema):
...
         self.properties_without_insertion_order = self.get_metadata()
-
+        self.property_transform_keys = self._properties_to_paths("propertyTransform")
+        self.property_transform = self._schema.get("propertyTransform")
         additional_identifiers = self._schema.get("additionalIdentifiers", [])
         self._additional_identifiers_paths = [
             {fragment_decode(prop, prefix="") for prop in identifier}
             for identifier in additional_identifiers
         ]
+        self.jinja_template = self.env.get_template("transformation.template")
```
jotompki, line 186: nit: remove jinja from name and replace with something like "transformation" to indicate what the template actually contains and not the library used to transform it.

src/rpdk/core/contract/templates/transformation.template
```diff
@@ -0,0 +1,5 @@
+var jsonata = require('jsonata')
+var model = {{input_model}}
+var expression = jsonata("{{jsonata_expression}}");
```
jotompki, line 3: thinking about this a bit more do you think it just makes sense to make this a static nodejs file that is a function with two arguments input_model and jsonata_expression?

README.md
```diff
@@ -15,6 +15,7 @@ This tool can be installed using [pip](https://pypi.org/project/pip/) from the P
 ```bash
 pip install cloudformation-cli cloudformation-cli-java-plugin cloudformation-cli-go-plugin cloudformation-cli-python-plugin
 ```
+You will need npm to run contract test (cfn test). Please ensure you have npm installed. Follow the instructions: https://www.npmjs.com/get-npm to get npm installed on your machine.
```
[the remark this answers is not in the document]
anshikg: not sure, if they can the way the code stands they will need npm because jsonata is getting installed before the transformation check is being made. It may be worth changing the code to make sure customers don't need to instal npm if they are not using the functionality.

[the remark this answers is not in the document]
PatMyron: > the way the code stands they will need npm because jsonata is getting installed before the transformation check is being made

could we check and not install `jsonata` during `cfn test` if they're not using `propertyTransform`?

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -1,4 +1,6 @@
 import logging
+import subprocess  # nosec
```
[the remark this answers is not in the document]
anshikg: yes it does.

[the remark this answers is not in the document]
benbridts: For python >=3.5 the documentation recommends using subprocess.run, if possible.
I don't know if that influences the `# nosec`

README.md
```diff
@@ -15,6 +15,7 @@ This tool can be installed using [pip](https://pypi.org/project/pip/) from the P
 ```bash
 pip install cloudformation-cli cloudformation-cli-java-plugin cloudformation-cli-go-plugin cloudformation-cli-python-plugin
 ```
+You will need npm to run contract test (cfn test) if your resource schema requires property transform. Please ensure you have npm installed. Follow the instructions: https://www.npmjs.com/get-npm to get npm installed on your machine.
```
jotompki, line 18:
```suggestion
You will need npm to run contract tests (cfn test) if your resource schema requires property transform. Please ensure you have npm installed. Follow the instructions: https://www.npmjs.com/get-npm to get npm installed on your machine.
```

src/rpdk/core/contract/resource_client.py
```diff
@@ -201,6 +215,78 @@ def get_metadata(self):
...
+    # removing coverage for this method as it is not possible to check both npm
+    # exists and does not exist condition in unit test
+    @staticmethod
+    def check_npm():  # pragma: no cover
+        output = subprocess.getoutput("npm list jsonata")
+        if "npm: command not found" not in output:
+            if "jsonata@" not in output:
+                subprocess.getoutput("npm install jsonata")
+        else:
+            LOG.error(
```
jotompki, line 270: should we make this method return something? if a user doesn't have npm installed since this method doesn't return anything looks like we will just continue to run the transformation function.

src/rpdk/core/contract/resource_client.py
```diff
@@ -201,6 +215,78 @@ def get_metadata(self):
...
+            if document_input != document_output:
+                transformed_property = self.transform(prop, input_model)
+                self.update_transformed_property(
+                    prop, transformed_property, input_model
+                )
+
+    def transform(self, property_path, input_model):
+
+        path = "/" + "/".join(property_path)
+        property_transform_value = self.property_transform[path].replace('"', '\\"')
```
jotompki, line 237:
```suggestion
        property_transform_value = json.dumps(self.property_transform[path])
```

Might help to catch [other escaped characters](https://stackoverflow.com/questions/18886596/replace-all-quotes-in-a-string-with-escaped-quotes)

jotompki (approved):
Good overall just want to be sure we don't run transformations when npm isn't installed

prerna-p (approved):

PR #669, opened 2021-01-21
Add cli version information in package artifact during cfn submit

omkhegde:
*Issue #, if available:*

*Description of changes:*
Adding CLI version information in a metadata file named _cfn_metadata.json. This file will be included in the package artifact zip being uploaded. 

### DIFF
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -51,7 +51,7 @@ DEFAULT_ROLE_TIMEOUT_MINUTES = 120  # 2 hours
 MIN_ROLE_TIMEOUT_SECONDS = 3600  # 1 hour
 MAX_ROLE_TIMEOUT_SECONDS = 43200  # 12 hours
 
-CFN_METADATA_FILE_NAME = "_cfn_metadata.json"
+CFN_METADATA_FILENAME = ".cfn_metadata.json"
 
 LAMBDA_RUNTIMES = {
     "noexec",  # cannot be executed, schema only
@@ -500,7 +500,7 @@ class Project:  # pylint: disable=too-many-instance-attributes,too-many-public-m
                             "%s not found. Not writing to package.", INPUTS_FOLDER
                         )
 
-                    metadata_file = open(CFN_METADATA_FILE_NAME, "w+")
+                    metadata_file = open(CFN_METADATA_FILENAME, "w+")
                     self._plugin.package(self, zip_file)
 
                     # Plugin package method adds the plugin language and version info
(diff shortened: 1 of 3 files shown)


src/rpdk/core/project.py
```diff
@@ -50,6 +51,7 @@
 MIN_ROLE_TIMEOUT_SECONDS = 3600  # 1 hour
 MAX_ROLE_TIMEOUT_SECONDS = 43200  # 12 hours
 
+CFN_METADATA_FILE_NAME = "_cfn_metadata.json"
```
jotompki, line 54: prefer a dot prefix

src/rpdk/core/project.py
```diff
@@ -499,6 +501,22 @@ def submit(
                         )
                     self._plugin.package(self, zip_file)
 
+                    # Plugin package method adds the plugin language and version info
+                    # to the metadata file. Update the file with cli version info and copy
+                    # to the zip artifact
+                    try:
+                        version_metadata = {}
+                        with open(CFN_METADATA_FILE_NAME, "r") as metadata_file:
+                            version_metadata = json.load(metadata_file)
```
jotompki, line 510: if file doesn't exist what's the behavior here?

src/rpdk/core/project.py
```diff
@@ -499,6 +501,22 @@ def submit(
...
+
+                        with open(CFN_METADATA_FILE_NAME, "w") as metadata_file:
+                            version_metadata["cli-version"] = __version__
+                            json.dump(version_metadata, metadata_file)
+
+                        zip_file.write(CFN_METADATA_FILE_NAME)
```
jotompki, line 516: if we are creating this then just using it to write to the zipfile, then removing want to use writestr to remove the initial opening of the file and removal of file after? https://docs.python.org/3/library/zipfile.html#zipfile.ZipFile.writestr

jotompki (changes requested):
unit tests for path?

PatMyron:
Can this change and all the plugin changes be picked up with the other if someone upgrades one and not the other?

src/rpdk/core/project.py
omkhegde: Since the plugins and cli are writing the version info independently, and zip archives cannot be updated, I believe we have to either write to a file and then copy it over to the archive, or create a new archive after the plugins have finished writing the version metadata.

omkhegde:
@PatMyron , yes since the plugins and the cli will be updating the metadata file independent of each other.

src/rpdk/core/project.py
jotompki: so plugins write to a zip file, then we open it here and write to the same file? maybe im missing the plugin portion as it would be in a subsequent PR. I think a better approach would be to implement a get_version_information method in each plugin code, pass that back to the main class, then write once

omkhegde: Sounds good, updated the PR. Corresponding Java plugin PR -> https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/pull/349/files

ammokhov (approved):
since it's a cross repo change, make sure to add install requirements similar to this - https://github.com/aws-cloudformation/cloudformation-cli-java-plugin/blob/master/setup.py#L39

omkhegde:
@ammokhov since this change is backward incompatible (the cli handles the case where the plugins do not provide version info), I dont think we need to update the minimum requirement

wbingli (approved):

jotompki (approved):
Looks great

PR #687, opened 2021-02-15
Add Regex and OR support for transform

anshikg:
*Issue #, if available:*

*Description of changes:* Add Regex and OR support for property Transform
Property Transform uses regex to compare ARN and OR to support multiple conditions. This change helps contract test support these conditions and make sure we are at parity with drift.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -231,8 +231,8 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                     property_transform_value = self.property_transform[path].replace(
                         '"', '\\"'
                     )
-                    if "OR" in property_transform_value:
-                        transform_functions = property_transform_value.split("OR")
+                    if "$OR" in property_transform_value:
+                        transform_functions = property_transform_value.split("$OR")
                         for transform_function in transform_functions:
                             transformed_property = self.transform(
                                 transform_function, input_model
@@ -268,9 +268,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     def compare(document_input, document_output):
         try:
             if isinstance(document_input, str):
-                regex = r"^{}$".format(document_input)
-                re.compile(regex)
-                return re.match(regex, document_output)
+                return re.match(f"^{document_input}$", document_output)
             return document_input == document_output
         except re.error:
             return document_input == document_output
@@ -497,7 +495,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         creds,
         token,
         callback_context=None,
-        **kwargs
+        **kwargs,
     ):
         return {
             "requestData": {
@@ -573,7 +571,7 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                 self._session, LOWER_CAMEL_CRED_KEYS, self._role_arn
             ),
             self.generate_token(),
-            **kwargs
+            **kwargs,
         )
 
     def _call(self, payload):
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -215,28 +215,67 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                 and properties[prop]["insertionOrder"] == "false"
             }
 
-    def transform_model(self, input_model, output_model):
+    def transform_model(self, input_model, output_model):  # pylint: disable=R0914
         if self.property_transform_keys:
             self.check_npm()
-        for prop in self.property_transform_keys:
-            document_input, _path_input, _parent_input = traverse(
-                input_model, list(prop)[1:]
-            )
-            document_output, _path_output, _parent_output = traverse(
-                output_model, list(prop)[1:]
-            )
-            if document_input != document_output:
-                transformed_property = self.transform(prop, input_model)
-                self.update_transformed_property(
-                    prop, transformed_property, input_model
+        for prop in self.property_transform_keys:  # pylint: disable=R1702
+            try:
+                document_input, _path_input, _parent_input = traverse(
+                    input_model, list(prop)[1:]
                 )
+                document_output, _path_output, _parent_output = traverse(
+                    output_model, list(prop)[1:]
+                )
+                if not self.compare(document_input, document_output):
+                    path = "/" + "/".join(prop)
+                    property_transform_value = self.property_transform[path].replace(
+                        '"', '\\"'
+                    )
+                    if "OR" in property_transform_value:
+                        transform_functions = property_transform_value.split("OR")
+                        for transform_function in transform_functions:
+                            transformed_property = self.transform(
+                                transform_function, input_model
+                            )
+                            value = self.convert_type(
+                                document_input, transformed_property
+                            )
+                            if self.compare(value, document_output):
+                                self.update_transformed_property(
+                                    prop, value, input_model
+                                )
+                    else:
+                        transformed_property = self.transform(
+                            property_transform_value, input_model
+                        )
+                        value = self.convert_type(document_input, transformed_property)
+                        self.update_transformed_property(prop, value, input_model)
 
-    def transform(self, property_path, input_model):
+            except KeyError:
+                pass
 
-        path = "/" + "/".join(property_path)
-        property_transform_value = json.dumps(self.property_transform[path])
-        # self.property_transform[path].replace('"', '\\"')
+    @staticmethod
+    def convert_type(document_input, transformed_property):
+        if isinstance(document_input, bool):
+            return bool(transformed_property)
+        if isinstance(document_input, int):
+            return int(transformed_property)
+        if isinstance(document_input, float):
+            return float(transformed_property)
+        return transformed_property
 
+    @staticmethod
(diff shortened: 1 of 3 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -215,28 +215,67 @@ def get_metadata(self):
...
+            return float(transformed_property)
+        return transformed_property
 
+    @staticmethod
+    def compare(document_input, document_output):
+        try:
+            if isinstance(document_input, str):
+                regex = r"^{}$".format(document_input)
+                re.compile(regex)
+                return re.match(regex, document_output)
```
benbridts, line 273: Some regex nitpicking:

The `r` isn't really needed here. It indicates that `\` should not be treated as a special character, and it has no effect on the part that is in the format.

```
>>> print('^{}$'.format('l1\nl2'))
^l1
l2$
```
If we're targeting python 3.6 or higher, we could use f-strings instead of `.format`, but in this case there isn't a big readability difference:

the re.compile(regex) line does nothing here, as you work with the return value of that if you want to compile a regex.
On the other hand: There are not a lot of reasons to compile a regex that you're only using once, so we don't need it anyway.

Putting all those things together, we cloud replace these three lines with

```
return re.match(f"^{document_input}$", document_output)
```

Note that we assume that document_input is a raw string or has its `\`s escaped.

tests/contract/test_resource_client.py
```diff
@@ -1238,7 +1248,24 @@ def test_transform_model_equal_output(resource_client):
     output_model = TRANSFORM_OUTPUT.copy()
 
     resource_client.transform_model(input_model, output_model)
-    assert input_model == TRANSFORM_OUTPUT
+    assert input_model == output_model
+
+
+def test_transform_model_equal_output_OR(resource_client):
+    input_model = {"e": "new", "a": "ValueA", "c": 1}
+    output_model = {"e": "newValue", "a": "ValueA", "c": 1}
```
benbridts, line 1256: Do we have a test for the other side of the OR? It's not 100% clear of what the OR function should do, but I assume that having 
```
output_model = {"e": "newTest", "a": "ValueA", "c": 1}
```
should also make input and output be seen as equivalent

tests/contract/test_resource_client.py
```diff
@@ -1238,7 +1248,24 @@ def test_transform_model_equal_output(resource_client):
...
+
+    resource_client.transform_model(input_model, output_model)
+    assert input_model == output_model
+
+
+def test_transform_model_not_equal_output_OR(resource_client):
```
benbridts, line 1262: It suprised me a little bit that we're testing that `"c": 5` cause a not equal. I would have expected the test to be focussed around `e` not being equal to the input. E.g. with `"e": "not-newValue"`

benbridts (commented):
Feel free to ignore my nitpicking, I initially was concerned about the `split("OR")`, and while looking at how that worked, noticed some other things that are a lot more personal preference.

tests/contract/test_resource_client.py
anshikg: makes sense. I will update the test to test the OR fail condition.

tests/contract/test_resource_client.py
anshikg: will add parameterized output to check both the conditions

src/rpdk/core/contract/resource_client.py
```diff
@@ -215,28 +215,67 @@ def get_metadata(self):
...
+                document_output, _path_output, _parent_output = traverse(
+                    output_model, list(prop)[1:]
+                )
+                if not self.compare(document_input, document_output):
+                    path = "/" + "/".join(prop)
+                    property_transform_value = self.property_transform[path].replace(
+                        '"', '\\"'
+                    )
+                    if "OR" in property_transform_value:
+                        transform_functions = property_transform_value.split("OR")
```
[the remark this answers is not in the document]
anshikg: OR (in upper case) is not supported in jsonata. It makes sense that I append $ for OR to denote a function. As for the space, I don't think it will matter. In this case trailing or leading spaces will not matter for jsonata, I would add a trim function to avoid any issues though. 
Please let me know if I missing the point.

tests/contract/test_resource_client.py
```diff
@@ -85,15 +85,21 @@
...
     "readOnlyProperties": ["/properties/b"],
     "createOnlyProperties": ["/properties/c"],
-    "primaryIdentifier": ["/properties/c"],
+    "primaryIdentifier": ["/properties/b"],
     "writeOnlyProperties": ["/properties/d"],
-    "propertyTransform": {"/properties/a": '$join([a, "Test"])'},
+    "propertyTransform": {
+        "/properties/a": '$join([a, "Test"])',
+        "/properties/c": "$power(2, 2)",
+        "/properties/e": '$join([e, "Test"]) $OR $join([e, "Value"])',
```
jotompki, line 97: is something like
```
"/properties/e": '$join([e, "T$OR"]) $OR $join([e, "Value"])',
```

currently supported? if not we should probably document the limitations of using the OR support

anshikg: ```$join([e, "T$OR"])``` is not supported. I am not sure I understand the use case when this is valid. Could you elaborate on that. I can add this to drift documentation.

src/rpdk/core/contract/resource_client.py
```diff
@@ -215,28 +215,67 @@ def get_metadata(self):
...
+                            )
+                            value = self.convert_type(
+                                document_input, transformed_property
+                            )
+                            if self.compare(value, document_output):
+                                self.update_transformed_property(
+                                    prop, value, input_model
+                                )
+                    else:
+                        transformed_property = self.transform(
```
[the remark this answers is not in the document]
anshikg: That is a valid point. But, I want contract test to be consistent with the functionality of the backend service. For example:
In the following test, I expect that at line 7 the input model is updated to ```{'a': 'ValueATest', "c": 1}``` where value of property a is ```ValueATest``` the transformed value and not ```ValueA``` the original value.
```
1. def test_transform_model_unequal_models(resource_client):
2.    input_model = INPUT.copy() #{'a': 'ValueA', 'c': 1}
3.    output_model = INVALID_OUTPUT.copy() #{"a": "ValueB", "c": 1}
4.
5.    resource_client.transform_model(input_model, output_model)
6.    assert input_model != output_model
7.    assert input_model == TRANSFORM_OUTPUT_WITHOUT_C  #{'a': 'ValueA', "c": 1} != {'a': 'ValueATest', "c": 1}
```

tests/contract/test_resource_client.py
jotompki: i think just letting people know that $OR is reserved and can't be used in transformations other than to specify multiple would be good

jotompki (approved):
Might be good to move some of the nested blocks to other functions out if it makes sense to cut down on the pylint disabling

src/rpdk/core/contract/resource_client.py
```diff
@@ -215,28 +215,65 @@ def get_metadata(self):
                 and properties[prop]["insertionOrder"] == "false"
             }
 
-    def transform_model(self, input_model, output_model):
+    def transform_model(self, input_model, output_model):  # pylint: disable=R0914
```
wbingli, line 218: It's not related with this PR, But I'm thinking the transform should not be used to transform the model into other model, but should only be used for comparing.  We can do such as `compare(input_model, output_model, transforms)`. The transform semantic here really means comparing, instead of transform A to B.

Other comment, which we have such method in `ResourceClient` we need to have a separated class for model based utils but leave the `ResourceClient` to manage the call with local server.

wbingli (approved):

src/rpdk/core/contract/resource_client.py
anshikg: Hey Wenbing, I am not sure I understand the first item. I am happy to set up a chat to talk about them. For now I am merging this CR and can send out a new CR with the suggested changes.

src/rpdk/core/contract/resource_client.py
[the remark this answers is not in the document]
benbridts: I phrased that a bit unclear. consider this (bit contrived schema fragment):

```
"propertyTransform": {
     "/properties/a": "$join([a, 'ORGANIZATION'])"
}
```
`property_transform_value.split("OR")` will return `['$join([a, "', 'GANIZATION"])']`.

The safest way to avoid that is to not split the string at all:
```
# this could be like IAM policies, either a string if there is only a single element or a list for OR behaviour
"propertyTransform": {
     "/properties/a": [
        "$join([a, 'ORGANIZATION'])",
        "$join([a, 'something else'])"
}
```

The mitigation I proposed to also look at the `$` being behind the OR  to reduce the likelihood of the separator being part of a word is probably less useful than I originally thought. There is a chance of both parts starting with `$`, but it is not guaranteed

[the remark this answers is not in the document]
benbridts: Edit: That might be more of a schema discussion than a contract test discussion, but since I haven't seen the OR behaviour documented, it might make sense to have that discussion anyway

PatMyron:
One thing to keep in mind for this is having protections in place against https://en.wikipedia.org/wiki/ReDoS

Issue #692, opened 2021-02-26
Invalid date in Module fragment - error is not clear to the user


priyanka-amzn:
When a yaml template is used with a non-string date, the error message to the user is not clear. 

If the module fragment starts with
```
AWSTemplateFormatVersion: 2010-09-09
```
instead of
```
AWSTemplateFormatVersion: "2010-09-09"
```

The output for `cfn validate` is:
```
=== Unhandled exception ===
Please report this issue to the team.
Issue tracker: https://github.com/aws-cloudformation/aws-cloudformation-rpdk/issues
Please include the log file 'rpdk.log'
```

priyanka-amzn:
fragments/sample.json
```
AWSTemplateFormatVersion: 2010-09-09
Description: A secure S3 Bucket. The features are Versioning and DeletionPolicy.
Parameters:
  BucketName:
    Description: Name for the bucket
    Type: String
Resources:
  S3Bucket:
    Type: AWS::S3::Bucket
    DeletionPolicy: Retain
    Properties:
      BucketName: !Ref 'BucketName'
      VersioningConfiguration:
        Status: Enabled
```

rpdk.log
```
[2021-02-26T01:05:36Z] INFO     - Validating your module fragments...
[2021-02-26T01:05:36Z] DEBUG    - Fragment directory: /Users/priyaha/testDate/fragments
[2021-02-26T01:05:36Z] DEBUG    - Unhandled exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/cli.py", line 100, in main
    args.command(args)
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/validate.py", line 14, in validate
    project.load()
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/project.py", line 421, in load
    self._validate_fragments(template_fragment)
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/project.py", line 440, in _validate_fragments
    template_fragment.validate_fragments()
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/fragment/generator.py", line 80, in validate_fragments
    self.__validate_fragment_thru_cfn_lint(raw_fragments)
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/fragment/generator.py", line 83, in __validate_fragment_thru_cfn_lint
    lint_warnings = self.__get_cfn_lint_matches(raw_fragments)
[...]

PR #694, opened 2021-03-01
Issue-692 [Modules] Don't fail validation when version string with no quotes

MalikAtalla-AWS:
[Issue 692](https://github.com/aws-cloudformation/cloudformation-cli/issues/692)

*Description of changes:*
Don't throw an internal error when template uses the version string without quotes
```
AWSTemplateFormatVersion: 2010-09-09
```
Instead of 
```
```

### DIFF
--- a/src/rpdk/core/fragment/generator.py
+++ b/src/rpdk/core/fragment/generator.py
@@ -105,7 +105,7 @@ class TemplateFragment:  # pylint: disable=too-many-instance-attributes
         filename = "temporary_fragment.json"
 
         with open(filename, "w") as outfile:
-            json.dump(raw_fragment, outfile, indent=4)
+            json.dump(raw_fragment, outfile, indent=4, default=str)
 
         template = cfnlint.decode.cfn_json.load(filename)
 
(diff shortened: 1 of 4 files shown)


jotompki (approved):

src/rpdk/core/fragment/generator.py
```diff
@@ -105,7 +105,7 @@ def __get_cfn_lint_matches(raw_fragment):
         filename = "temporary_fragment.json"
 
         with open(filename, "w") as outfile:
-            json.dump(raw_fragment, outfile, indent=4)
+            json.dump(raw_fragment, outfile, indent=4, default=str)
```
jotompki, line 108: does this affect any other values in templates? i guess cloudformation interprets everything as a string so it shouldnt matter?

MalikAtalla-AWS: I'm not aware of another case, but I think exporting as string is the right behaviour. This is just about getting the linting results when creating a module, so the blast radius of this change is pretty small.

miparnisari (approved):

PR #698, opened 2021-03-05
Enable Encryption in Modules example fragment

MalikAtalla-AWS:
Enable Encryption in Modules example fragment, because that's good practice in S3

### DIFF
--- a/fragments/sample.json
+++ b/fragments/sample.json
@@ -15,6 +15,15 @@
                 "BucketName": {
                     "Ref": "BucketName"
                 },
+                "BucketEncryption": {
+                    "ServerSideEncryptionConfiguration": [
+                        {
+                            "ServerSideEncryptionByDefault": {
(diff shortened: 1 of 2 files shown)


fragments/sample.json
```diff
@@ -15,6 +15,15 @@
...
                 },
+                "BucketEncryption": {
+                    "ServerSideEncryptionConfiguration": [
+                        {
+                            "ServerSideEncryptionByDefault": {
+                                "SSEAlgorithm": "AES256"
+                            }
+                        }
+                    ]
+                },
```
MalikAtalla-AWS, line 26: This whole file really doesn't need to be here. It's an artifact of the modules unit tests. I wasn't able to figure out how to get rid of it today with my time constraints, but will do this in a separate PR.

PatMyron (approved):

miparnisari (approved):

Issue #693, opened 2021-02-26
Module yaml fragment: block mapping parsing error


priyanka-amzn:
4 files are part of the attached zip file: 2 module fragments(1 yaml, 1 json) and 2 rpdk(1 for yaml case, 1 for json case) log files.

When I ran `cfn-validate` for `ec2withsub.json` fragment, the output was:
```
Module fragment is probably valid, but there are warnings/errors from cfn-lint (https://github.com/aws-cloudformation/cfn-python-lint):
	At least one of [ImageId, LaunchTemplate] should be specified for Resources/EC2Instance/Properties (from rule E2522: Check Properties that need at least one of a list of properties)
	Fn::Sub isn't needed because there are no variables at Resources/EC2Instance/Properties/UserData/Fn::Base64/Fn::Sub (from rule W1020: Sub isn't needed if it doesn't have a variable defined)
```
But when I ran `cfn-validate` for `ec2withsub.yaml` which has the same cfn resource as `ec2withsub.json` but is in yaml, the output was:
```
Invalid template fragment: Fragment file '/Users/priyaha/testDate/fragments/sample.json' is invalid: while parsing a block mapping
  in "<unicode string>", line 8, column 9:
            Fn::Base64: Sub |
            ^
expected <block end>, but found '<scalar>'
  in "<unicode string>", line 10, column 11:
              sudo yum install -y https://s3.r ... 
              ^
```

[module fragments and logs.zip](https://github.com/aws-cloudformation/cloudformation-cli/files/6046847/module.fragments.and.logs.zip)

MalikAtalla-AWS:
Thanks for reporting this. I'm able to reproduce the issue from your description.

MalikAtalla-AWS:
Will be resolved in versions >= 0.2.6

Issue #689, opened 2021-02-19
Module validation reports incorrect cfn-lint warnings 


michaelwittig:
I have the following fragment (`fragments/module.yml`):
```yaml
---
AWSTemplateFormatVersion: '2010-09-09'
Description: 'VPC Subnet'
Parameters:
  VpcId:
    Description: 'VPC id'
    Type: String
  VpcCidrBlock:
    Description: 'VPC CIDR block'
    Type: String
  InternetGatewayId:
    Description: 'VPC Internet Gateway id'
    Type: String
  Reach:
    Description: 'Should the subnet has a route to the Internet?'
    Type: String
    AllowedValues: [private, public]
  AZIndex:
    Description: 'Index of the AZ where the subnet should be created in'
    Type: Number
    MinValue: 0
    MaxValue: 2
  AZChar:
    Description: 'Availability zone char'
    Type: String
    AllowedValues: [A, B, C]
  SubnetIndex:
    Description: 'Index of the subnet'
    Type: Number
    MinValue: 0
    MaxValue: 5
  SubnetCount:
    Description: 'To slice the IP address ranges you need to specify how many subnets you want to create in the VPC'
    Type: Number
    MinValue: 1
    MaxValue: 6
Conditions:
  IsPublic: !Equals [!Ref Reach, 'public']
Resources:
  Subnet:
    Type: 'AWS::EC2::Subnet'
    Properties:
      AvailabilityZone: !Select [!Ref AZIndex, !GetAZs '']
      CidrBlock: !Select [!Ref SubnetIndex, !Cidr [!Ref VpcCidrBlock, !Ref SubnetCount, 12]]
      VpcId: !Ref VpcId
      Tags:
      - Key: Name
        Value: !Sub '${AZChar} ${Reach}'
      - Key: Reach
        Value: !Ref Reach
  RouteTable:
    Type: 'AWS::EC2::RouteTable'
    Properties:
      VpcId: !Ref VpcId
      Tags:
      - Key: Name
        Value: !Sub '${AZChar} ${Reach}'
  SubnetRouteTableAssociation:
    Type: 'AWS::EC2::SubnetRouteTableAssociation'
    Properties:
      SubnetId: !Ref Subnet
      RouteTableId: !Ref RouteTable
  RouteInternet:
    Type: 'AWS::EC2::Route'
    Condition: IsPublic
    Properties:
      RouteTableId: !Ref RouteTable
[...]

PR #699, opened 2021-03-10
[Modules] Support parsing Yaml fragments with CFN short forms

MalikAtalla-AWS:
Issues #693 and #689 

*Description of changes:*
 * Using [cfn-flip](https://github.com/awslabs/aws-cfn-template-flip) to parse the Module fragments so we can support the full range of CFN yaml syntax. This is solving the problem that we can't currently parse templates like the one below:

```yaml
Resources:
  EC2Instance:
    Type: AWS::EC2::Instance
    Properties:
      ImageId: ami-09c5e030f74651050
      UserData:
        Fn::Base64: !Sub |
          #!/bin/bash -e
          sudo yum install -y https://s3.region.amazonaws.com/amazon-ssm-region/latest/linux_amd64/amazon-ssm-agent.rpm
          sudo systemctl status amazon-ssm-agent
          sudo systemctl enable amazon-ssm-agent
          sudo systemctl start amazon-ssm-agent
```
 * Apply cfn-lint on the original, unmodified module fragment file to avoid any false positives from cfn-lint
 * This PR also extracts some module-reading logic from generator.py into a separate module.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -52,6 +52,7 @@ setup(
         "docker>=4.3.1",
         "ordered-set>=4.0.2",
         "cfn-lint>=0.43.0",
+        "cfn_flip>=1.2.3",
         "nested-lookup",
     ],
     entry_points={
--- a/src/rpdk/core/fragment/generator.py
+++ b/src/rpdk/core/fragment/generator.py
@@ -13,17 +13,17 @@ from pathlib import Path
 
 import cfnlint.config
 import cfnlint.core
-import yaml
 
 from rpdk.core.data_loaders import resource_json
 from rpdk.core.exceptions import FragmentValidationError
+
+from .module_fragment_reader import get_template_file_size_in_bytes, read_raw_fragments
 
 LOG = logging.getLogger(__name__)
 FRAGMENT_DIR = "fragments"
 SAMPLE_FRAGMENT_OUTPUT = "sample.json"
 SCHEMA_NAME = "schema.json"
 SAMPLE_FRAGMENT = "../data/examples/module/sample.json"
-ALLOWED_EXTENSIONS = {".json", ".yaml", ".yml"}
 RESOURCE_LIMIT = 500
 OUTPUT_LIMIT = 200
 MAPPING_LIMIT = 200
@@ -45,7 +45,7 @@ class TemplateFragment:  # pylint: disable=too-many-instance-attributes
         LOG.debug("Fragment directory: %s", self.fragment_dir)
 
     def generate_schema(self):
-        raw_fragments = self._read_raw_fragments()
+        raw_fragments = read_raw_fragments(self.fragment_dir)
 
         schema = {}
         properties = {}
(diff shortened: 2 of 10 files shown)


setup.py
```diff
@@ -52,6 +52,7 @@ def find_version(*file_paths):
         "docker>=4.3.1",
         "ordered-set>=4.0.2",
         "cfn-lint>=0.43.0",
+        "cfn_flip>=1.2.3",
```
priyanka-amzn, line 55: Is this an Amazon approved cfn flip? If it is the same as https://cfnflip.com/, we might have to switch to an internal tool.

src/rpdk/core/fragment/generator.py
```diff
@@ -100,14 +104,12 @@ def __validate_outputs(self, raw_fragments):
...
-        with open(filename, "w") as outfile:
-            json.dump(raw_fragment, outfile, indent=4, default=str)
-
-        template = cfnlint.decode.cfn_json.load(filename)
+    def __get_cfn_lint_matches(self):
+        filepath = _get_fragment_file(self.fragment_dir)
+        try:
+            template = cfnlint.decode.cfn_json.load(filepath)
+        except json.decoder.JSONDecodeError:
+            template = cfnlint.decode.cfn_yaml.load(filepath)
```
priyanka-amzn, line 112: Nice!
We could add another `except` for a general exception here. That way if there is any linting failure, we could add a log and display a warning to the user instead of failing the `validate` operation.

src/rpdk/core/fragment/module_fragment_reader.py
```diff
@@ -0,0 +1,54 @@
+import logging
```
priyanka-amzn, line 1: Thanks for refactoring this

src/rpdk/core/fragment/module_fragment_reader.py
```diff
@@ -0,0 +1,54 @@
...
+                all_fragment_files.append(os.path.join(root, f))
+    if len(all_fragment_files) > 1:
+        raise FragmentValidationError(
+            "A Module can only consist of a "
+            "single template file, but there are "
+            + str(len(all_fragment_files))
+            + ": "
+            + str(all_fragment_files)
+        )
+    return all_fragment_files[0]
```
priyanka-amzn, line 46: This can cause `list index out of range error` if there are no fragment files since we aren't checking if `len(all_fragment_files)` < 1.

setup.py
MalikAtalla-AWS: It's [this](https://github.com/awslabs/aws-cfn-template-flip) which I believe is maintained by AWS.

src/rpdk/core/fragment/module_fragment_reader.py
MalikAtalla-AWS: Good catch. Added a condition in the latest commit.

src/rpdk/core/fragment/generator.py
MalikAtalla-AWS: Addressed this in the latest commit.

priyanka-amzn (approved):

miparnisari (approved):

PR #711, opened 2021-03-23
Update prune_properties to support pruning attributes for all sequence members

privetjr:
## Notes

Previously, contract tests could not prune an attribute from *all* sequence members, but pruning an attribute for a specific attribute was doable (by specifying a valid index in the path to prune). If supplying an asterisk (`*`) in the path for a sequence, the `traverse` method would run into a ValueError:
```
~/.virtualenvs/local_cfn/lib/python3.8/site-packages/rpdk/core/jsonutils/utils.py:114: in traverse
    part = int(part)
E   ValueError: invalid literal for int() with base 10: '*'
```

This pull request adds support for pruning attributes for all sequence members, when the index provided for the sequence is `*`.

(NOTE: it could easily be any other unpack identifier that isn't a valid property name, but asterisk felt like a good choice)

## Example

An example of a schema with readonly sequence members is `AWS::FraudDetector::EventType` ([schema json](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-frauddetector/blob/master/aws-frauddetector-eventtype/aws-frauddetector-eventtype.json#L278-L286))

## Related Issue:
https://github.com/aws-cloudformation/cloudformation-cli/issues/478

## Description of changes:

`prune_properties` was updated with a conditional that checks for `*` in the path, and uses a different version of `traverse` for those paths. This different version of traverse returns multiple resolved paths, which are then used to delete attributes from the original document.

The additional `traverse` (named `traverse_path_for_sequence_members`) uses recursion to continue to traverse multiple documents when the path "branches" for all sequence members.

## Testing:
`pre-commit run --all-files` is happy, and previously failing contract tests pass :)

---

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -22,9 +22,18 @@ from ..boto_helpers import (
     get_temporary_credentials,
 )
 from ..jsonutils.pointer import fragment_decode, fragment_list
-from ..jsonutils.utils import item_hash, traverse, traverse_raw_schema
+from ..jsonutils.utils import (
+    UNPACK_SEQUENCE_IDENTIFIER,
+    item_hash,
+    traverse,
+    traverse_path_for_sequence_members,
+    traverse_raw_schema,
+)
 
 LOG = logging.getLogger(__name__)
+LOOKUP_ERROR_MESSAGE_FORMAT = (
+    "Caught LookupError when pruning properties for document %s and path %s"
+)
 
 
 def prune_properties(document, paths):
@@ -35,13 +44,37 @@ def prune_properties(document, paths):
     for convenience. (The return value may be ignored.)
     """
     for path in paths:
+        # if '*' is in path, we need to prune more than one property (prune property for all members of array)
+        if UNPACK_SEQUENCE_IDENTIFIER in path:
+            document = _prune_properties_for_all_sequence_members(document, path)
+            continue
         try:
             _prop, resolved_path, parent = traverse(document, path)
         except LookupError:
-            pass  # not found means nothing to delete
+            # not found means nothing to delete
+            LOG.info(LOOKUP_ERROR_MESSAGE_FORMAT, document, path)
         else:
             key = resolved_path[-1]
             del parent[key]
+    return document
+
+
+def _prune_properties_for_all_sequence_members(document: dict, path: list) -> dict:
+    try:
+        # this returns multiple paths
+        _prop, resolved_paths = traverse_path_for_sequence_members(document, path)
+    except LookupError:
+        # not found means nothing to delete
+        LOG.info(LOOKUP_ERROR_MESSAGE_FORMAT, document, path)
+    else:
+        # paths with indices are gathered in increasing order, but we need to prune in reverse order
+        resolved_paths = resolved_paths[::-1]
+        for resolved_path in resolved_paths:
+            new_doc = document
+            for key in resolved_path[: len(resolved_path) - 1]:
+                new_doc = new_doc[key]
+            key = resolved_path[-1]
+            del new_doc[key]
     return document
 
 
--- a/src/rpdk/core/jsonutils/utils.py
+++ b/src/rpdk/core/jsonutils/utils.py
@@ -2,7 +2,7 @@ import hashlib
 import json
 import logging
 from collections.abc import Mapping, Sequence
-from typing import Any
+from typing import Any, List, Tuple
 
 from nested_lookup import nested_lookup
 from ordered_set import OrderedSet
@@ -14,6 +14,7 @@ LOG = logging.getLogger(__name__)
 NON_MERGABLE_KEYS = ("uniqueItems", "insertionOrder")
 TYPE = "type"
 REF = "$ref"
+UNPACK_SEQUENCE_IDENTIFIER = "*"
 
 
 class FlatteningError(Exception):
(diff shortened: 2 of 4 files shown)


src/rpdk/core/contract/resource_client.py
```diff
@@ -49,6 +57,24 @@ def prune_properties(document, paths):
...
+        pass  # not found means nothing to delete
+    else:
+        # paths with indices are gathered in increasing order, but we need to delete in reverse order
+        resolved_paths = resolved_paths[::-1]
+        for resolved_path in resolved_paths:
+            new_doc = document
+            for key in resolved_path[: len(resolved_path) - 1]:
+                new_doc = new_doc[key]
+            key = resolved_path[-1]
+            del new_doc[key]
```
privetjr, line 74: The pruning of the attributes happens a little differently, as `traverse_path_for_sequence_members` does not return a parent, and returns multiple resolved_paths. Due to the indices being gathered in sequential order, we remove attributes in reverse order to avoid IndexError.

src/rpdk/core/jsonutils/utils.py
```diff
@@ -111,6 +112,79 @@ def traverse(document, path_parts):
...
+            )[0]
+            new_documents.extend(new_document)
+        for i in range(len(new_documents)):  # pylint: disable=consider-using-enumerate
+            new_document = new_documents[i]
+            newer_documents, newer_paths = traverse_path_for_sequence_members(
+                new_document, path_parts, new_paths[i]
+            )
+            documents.extend(newer_documents)
+            resolved_paths.extend(newer_paths)
+        return documents, resolved_paths
```
privetjr, line 180: Thoughts on this bit? I was looking into refactoring this bit, but I could not figure out a way to do so without re-writing `traverse_path_for_sequence_members` to support multiple documents as input arguments, and the added complexity was not worth it IMHO.

tests/jsonutils/test_utils.py
```diff
@@ -58,3 +63,128 @@ def test_schema_merge_overwrite_and_merge():
...
+    case_2 = {"foo": {"bar": [42, 43, 44]}}, ["foo"]
+    result_2 = (
+        [{"bar": [42, 43, 44]}],
+        [tuple(["foo"])],  # noqa: C409
+    )
+    case_3 = {"foo": {"bar": [42, 43, 44]}}, ("foo", "bar")
+    result_3 = [[42, 43, 44]], [("foo", "bar")]
+    case_4_key_error = {}, ["foo"]
+    case_5_value_error = [], ["foo"]
+    case_6_index_error = [], [0]
```
privetjr, line 81: FYI: These tests are duplicated by DocTests

tests/contract/test_resource_client.py
```diff
@@ -302,6 +302,32 @@ def test_prune_properties():
     assert document == {"one": "two", "array": ["first"]}
 
 
+def test_prune_properties_for_all_sequence_members():
```
anshikg, line 305: Can we add an array of an object type and delete a property which is embedded in the object. This would give high confidence to the code

src/rpdk/core/jsonutils/utils.py
anshikg: I agree, supporting multiple documents as input arguments does not sounds right in this case

anshikg (commented):
The CR looks good to me. If we can add a test case for a complex array type we should be good to merge this

ammokhov (commented):
would you mind to add type annotations? I know we dont do a good job of annotating code in this project but maybe a good start. I would also ask you to add more examples to the PR details

privetjr:
> would you mind to add type annotations?

Sounds good, I'll rebase and add type annotations.

> I would also ask you to add more examples to the PR details

Also reasonable. I'll add more examples into the PR details, as well as additional unit tests (e.g. the complex array test Anshika suggested).

tests/contract/test_resource_client.py
privetjr: Tests have been added.

src/rpdk/core/jsonutils/utils.py
```diff
@@ -185,6 +186,133 @@ def traverse_raw_schema(schema: dict, path: tuple):
...
+    The next path part is added to the traversed path
+    """
+    sequence_part = current_path_parts.pop(0)
+    if sequence_part == UNPACK_SEQUENCE_IDENTIFIER:
+        return _handle_unpack_sequence_for_traverse(
+            current_document, current_path_parts, current_path
+        )
+    # otherwise, sequence part should be a valid index
+    current_sequence_part = int(sequence_part)
+    current_document = current_document[current_sequence_part]
```
ammokhov, line 276: would be great to handle index error as well

src/rpdk/core/jsonutils/utils.py
```diff
@@ -185,6 +186,133 @@ def traverse_raw_schema(schema: dict, path: tuple):
...
+    The new document is obtained from the current document (a sequence) using the new path part as the index.
+    The next path part is added to the traversed path
+    """
+    sequence_part = current_path_parts.pop(0)
+    if sequence_part == UNPACK_SEQUENCE_IDENTIFIER:
+        return _handle_unpack_sequence_for_traverse(
+            current_document, current_path_parts, current_path
+        )
+    # otherwise, sequence part should be a valid index
+    current_sequence_part = int(sequence_part)
```
ammokhov, line 275: does this mean that `sequence_part` can be of two values (`*` or `int`)? in other words we either handle all or just one

src/rpdk/core/jsonutils/utils.py
```diff
@@ -185,6 +186,133 @@ def traverse_raw_schema(schema: dict, path: tuple):
...
+    current_document = current_document[current_sequence_part]
+    current_path.append(current_sequence_part)
+    return [current_document], [tuple(current_path)]
+
+
+def _handle_unpack_sequence_for_traverse(
+    current_document, current_path_parts, current_path
```
ammokhov, line 282: nit: `current_document: Sequence, current_path_parts: list, current_path: list`

privetjr: Whoops, missed the type annotations for this helper. I will add.

src/rpdk/core/contract/resource_client.py
```diff
@@ -45,6 +55,24 @@ def prune_properties(document, paths):
     return document
 
 
+def _prune_properties_for_all_sequence_members(document: dict, path: list) -> dict:
+    try:
+        # this returns multiple paths
+        _prop, resolved_paths = traverse_path_for_sequence_members(document, path)
+    except LookupError:
+        pass  # not found means nothing to delete
```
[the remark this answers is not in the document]
privetjr: Yup, will revise

src/rpdk/core/jsonutils/utils.py
privetjr: Yup, `sequence_part` should either be `UNPACK_SEQUENCE_IDENTIFIER` (`*`) or a valid index (`int`).

src/rpdk/core/jsonutils/utils.py
privetjr: Hrm, I think that would be increasing the scope of this PR. As it stands today, `traverse` simply throws `IndexError`, it does not handle it. You'd know better than I, but since `traverse` is just throwing `IndexError`, I figured `traverse_path_for_sequence_members` should also just throw it, so as to keep scope reduced as small as possible.

ammokhov: fair enough

ammokhov (approved):

wbingli:
This supports a new syntax for `readOnlyProperties` or `createOnlyProperties`. Could you give more details about how to use this new syntax?

In the meanwhile, is this syntax with wildcard `*` the right syntax ?  What's other options? e..g should we allow to configure the `readOnly` and `writeOnly` for definition?  This can help to define `readOnlyProperties` and `createOnlyProperties` for multiple or a list of properties.  What's the cons and pros and why we need to use this option?

privetjr:
> This supports a new syntax for readOnlyProperties or createOnlyProperties. Could you give more details about how to use this new syntax?

The schema can use the sequence unpack identifier (currently `*`) to indicate that all sequence members should be marked as readonly or writeonly. Here are a couple of examples:
* [readonly properties in AWS::FraudDetector::Detector](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-frauddetector/blob/master/aws-frauddetector-detector/aws-frauddetector-detector.json#L473-L479)
* [readonly properties in AWS::CE::AnomalySubscription](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-cost-explorer/blob/master/anomalysubscription/aws-ce-anomalysubscription.json#L99)
* [writeonly properties in AWS::ElasticLoadBalancingV2::Listener](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-elasticloadbalancingv2/blob/master/aws-elbv2-listener/aws-elasticloadbalancingv2-listener.json#L278-L280)

> In the meanwhile, is this syntax with wildcard `*` the right syntax ? What's other options? e..g should we allow to configure the `readOnly` and `writeOnly` for definition? This can help to define `readOnlyProperties` and `createOnlyProperties` for multiple or a list of properties. What's the cons and pros and why we need to use this option?

We could use any other character not allowed in the name of a property, but wildcard/asterisk seems like the right character to use. It's intuitive, and as the examples above demonstrate, folks are using wildcard in their schemas and expecting the behavior to be s.t. each sequence member is included. I can't think of a use case where the sequence unpack identifier being different for `readonly` vs. `writeonly` would be valuable, so I think it's fine that we have a single unpack identifier. 

**Pros of merging these changes**
* Intuitive sequence unpack identifier (`*`), customers will use this without being asked to or being guided to
* Allows CFN registry customers to define sequence members as readonly, writeonly, etc.
* Contract tests can succeed for schemas w/ readonly or writeonly sequence members
* Schemas that do not have `*` in their declaration of readonly/writeonly properties will be unaffected

**Cons of merging these changes**
* ? Let me know if you can think of any

gandhek (approved):

wbingli:
> We could use any other character not allowed in the name of a property, but wildcard/asterisk seems like the right character to use. It's intuitive, and as the examples above demonstrate, folks are using wildcard in their schemas and expecting the behavior to be s.t. each sequence member is included. I can't think of a use case where the sequence unpack identifier being different for readonly vs. writeonly would be valuable, so I think it's fine that we have a single unpack identifier.

I'm thinking the option to define `readOnly` and `writeOnly` inside the definition.  Those `*` is trying to match with all the definitions used in the resource properties. Actually you want to define property inside a definition as `readOnly` or `writeOnly`. 

For example, instead of [following](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-frauddetector/blob/master/aws-frauddetector-detector/aws-frauddetector-detector.json#L473-L479)
```
    "/properties/Rules/*/RuleVersion",
    "/properties/Rules/*/Arn",
```

What we can do in the definition :

```
  "Rule": {
      "type": "object",
      "properties": {
        "RuleId": {
          "type": "string"
        },
        "RuleVersion": {
          "type": "string",
          "readOnly": true
        },
        "Arn": {
          "type": "string",
          "readOnly": true
        }
```

Besides that,  this change is not a simple contract test fix. It's about how we are going to support such use case in schema. We should provide support in https://github.com/aws-cloudformation/cloudformation-resource-schema as well.

wbingli:
As discussed, this won't cause any regression to other CFN features. It helps on contract test for now. We need to follow up to support it in the schema as well as other CFN features should support this synatx.

wbingli (approved):

PR #714, opened 2021-03-24
contract test integration test

PatMyron:
[Installing the AWS SAM CLI on Linux](https://docs.aws.amazon.com/serverless-application-model/latest/developerguide/serverless-sam-cli-install-linux.html)

[Testing resource types using contract tests](https://docs.aws.amazon.com/cloudformation-cli/latest/userguide/resource-type-test.html)

[`Encrypted secrets`](https://docs.github.com/en/free-pro-team@latest/actions/reference/encrypted-secrets) for IAM actions currently required for resource ([`AWS::LookoutVision::Project`](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-lookoutvision/blob/main/aws-lookoutvision-project/aws-lookoutvision-project.json)) ([set here](https://github.com/aws-cloudformation/cloudformation-cli/settings/secrets/actions))

### DIFF
--- a/.github/workflows/contract-integration.yaml
+++ b/.github/workflows/contract-integration.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [ master ]
 jobs:
-  job:
+  contract-integration:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
--- a/.github/workflows/contract-integration.yaml
+++ b/.github/workflows/contract-integration.yaml
@@ -1,0 +1,31 @@
+on: [push, pull_request, workflow_dispatch]
+jobs:
+  job:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


anshikg (commented):
This looks good to me. My only concern is consistency. If these are passing consistently this is super helpful but, we might want to reconsider if it starts creating noise

anshikg (approved):

PatMyron:
> This looks good to me. My only concern is consistency. If these are passing consistently this is super helpful but, we might want to reconsider if it starts creating noise

Share the same concern on current flakiness. One option is that we could merge and run this on PRs [but not require it to pass to merge](https://github.com/aws-cloudformation/cloudformation-cli/settings/branch_protection_rules/2167468) until contract tests pass more consistently

jotompki (approved):
Well done! this is great

PR #716, opened 2021-03-26
Adding a unit test for module schema generation

MalikAtalla-AWS:
*Description of changes:*
Adding this over-assertive unit test for modules schema generation. Asserting on the exact schema that is currently generated for this fairly complex module, so that we have one more way of being notified when something about the schema generation changes. The schema generation is coupled to certain logic in the CFN server-side and we should be cautious when making changes.

The module fragment is taken from the official sample templates in the CFN documentation: https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/sample-templates-services-us-west-2.html#w2ab1c35c58c13c27

### DIFF
--- a/tests/fragments/test_generator.py
+++ b/tests/fragments/test_generator.py
@@ -42,6 +42,23 @@ def test_schema_generator(template_fragment, tmpdir):
     __validate_against_meta_schema(schema)
 
 
+def test_schema_generator_with_complex_real_module(template_fragment, tmpdir):
+    schema = __generate_schema("OpsWorks.json", template_fragment)
+    schema_file = tmpdir.join("schema.json")
+    assert os.path.exists(schema_file)
+    generated_schema = __read_file(schema_file)
+    expected_schema = __read_file(
+        os.path.join(directory, "../data/sample_fragments/OpsWorksSchema.json")
+    )
+    assert generated_schema == expected_schema
+    __validate_against_meta_schema(schema)
+
+
+def __read_file(fragment_file):
+    with open(fragment_file, "r", encoding="utf-8") as f:
+        return f.read()
+
+
 def test_schema_generation_param_without_description(template_fragment):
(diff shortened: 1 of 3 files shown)


tests/fragments/test_generator.py
```diff
@@ -42,6 +42,23 @@ def test_schema_generator(template_fragment, tmpdir):
...
 
+def test_schema_generator_with_complex_real_module(template_fragment, tmpdir):
+    schema = __generate_schema("OpsWorks.json", template_fragment)
+    schema_file = tmpdir.join("schema.json")
+    assert os.path.exists(schema_file)
+    generated_schema = __read_file(schema_file)
+    expected_schema = __read_file(
+        os.path.join(directory, "../data/sample_fragments/OpsWorksSchema.json")
+    )
+    assert generated_schema == expected_schema
```
priyanka-amzn, line 53: According to [this answer](https://www.quora.com/How-do-I-compare-two-JSON-files-in-Python?share=1) for "How do I compare two JSON files in Python?"
> the == operator on two dicts doesn’t compare the ordering of the keys. 

Suggestion: If we find the JSON diff and print it as an error log here in case of failure, that'll help us understand what changed.

MalikAtalla-AWS: These are not dicts. This is a simple string comparison.

MalikAtalla-AWS: Sorry, I thin I know what you're saying. I think it's not necessary to make the test pass when the order in the json is different. There is no reason for that to happen without an intentional change to the schema generation.

priyanka-amzn: > I think it's not necessary to make the test pass when the order in the json is different.

Got it 👍

priyanka-amzn (approved):

miparnisari (approved):

PR #718, opened 2021-03-29
support regex comparison for transformed values

anshikg:
*Issue #, if available:*

*Description of changes:* property Transform allows users to compare regex with live state from read handler. Regex comparison fails in input_equals_output method as  we were equating regex to live state rather than checking for regex match. This CR adds a check to compare regex rather than equating strings.

Lets say a customer adds the name of a property when creating the resource. This property name is updated to an ARN. propertyTransform property in the schema will allow one to define the arn regex to ensure this property is not falsely drifted.
```
document_input: {
  "propertyName": "Name"
}
```
```
document_output: {
  "propertyName": "arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster/Name"
}
```
where the propertyTransform in the schema is:
```
"propertyTransform": {
    "/properties/propertyName": "$join([\"arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster\\/\", PropertyName])"
  }
```

*Test:* 
Added a property transform which compares a string against a regex and this check seems to pass now. 

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -268,17 +268,17 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
     def compare(document_input, document_output):
         try:
             if isinstance(document_input, str):
-                return re.match(f"^{document_input}$", document_output)
+                return bool(re.match(f"^{document_input}$", document_output))
             return document_input == document_output
         except re.error:
             return document_input == document_output
 
     def transform(self, property_transform_value, input_model):
-        LOG.warning("This is the transform %s", property_transform_value)
+        LOG.debug("This is the transform %s", property_transform_value)
         content = self.transformation_template.render(
             input_model=input_model, jsonata_expression=property_transform_value
         )
-        LOG.warning("This is the content %s", content)
+        LOG.debug("This is the content %s", content)
         file = tempfile.NamedTemporaryFile(
             mode="w+b",
             buffering=-1,
--- a/src/rpdk/core/contract/suite/handler_commons.py
+++ b/src/rpdk/core/contract/suite/handler_commons.py
@@ -180,8 +180,8 @@ def test_input_equals_output(resource_client, input_model, output_model):
                     pruned_input_model[key], pruned_output_model[key]
                 )
             else:
-                assert (
-                    pruned_input_model[key] == pruned_output_model[key]
+                assert resource_client.compare(
+                    pruned_input_model[key], pruned_output_model[key]
                 ), assertion_error_message
     except KeyError as e:
         raise AssertionError(assertion_error_message) from e


src/rpdk/core/contract/resource_client.py
```diff
@@ -268,17 +268,17 @@ def convert_type(document_input, transformed_property):
     def compare(document_input, document_output):
         try:
             if isinstance(document_input, str):
-                return re.match(f"^{document_input}$", document_output)
+                return bool(re.match(f"^{document_input}$", document_output))
```
ammokhov, line 271: could u provide and example of what you are trying to achieve? 
would like to see what form `document_input` takes and what is `document_output`. 

I can see that it is used as:
```python
resource_client.compare(
    pruned_input_model[key],
    pruned_output_model[key]
)
```

is **key** a property name, if so then why do you need to use regex match

anshikg: Lets say a customer adds the name of a property when creating the resource. This property name is updated to an ARN. ```propertyTransform ``` property in the schema will allow one to define the arn regex to ensure this property is not falsely drifted.

```
document_input: {
  "propertyName": "Name"
}
```

```
document_output: {
  "propertyName": "arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster/Name"
}
```

where the propertyTransform in the schema is:
```
"propertyTransform": {
    "/properties/propertyName": "$join([\"arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster\\/\", PropertyName])"
  }
```

ammokhov: great details, would be great to add it to PR details. 

but still trying to understand where regex is coming from? from property pattern like:
```json
{
  "properties": {
    "propertyA": {
      "type": "...",
      "pattern": "^arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster/Name$"
    }
  }
}
```
or it is coming from transform and is used only on comparison during contract test execution?

anshikg: It is coming from propertyTransform.

```
"propertyTransform": {
    "/properties/propertyName": "$join([\"arn:(aws)[-]{0,1}[a-z]{0,2}[-]{0,1}[a-z]{0,3}:ecs:[a-z]{2}[-]{1}[a-z]{4,10}[-]{1}[1-3]{1}:[0-9]{12}[:]{1}cluster\\/\", PropertyName])"
  }
```

Adding the pattern check will not allow customers to set the string ```Name``` for the property. ```propertyTransform``` is primarily used by Drift. Contract test is just replicating the drift functionality for testing purposes.

src/rpdk/core/contract/resource_client.py
```diff
@@ -268,17 +268,17 @@ def convert_type(document_input, transformed_property):
     def compare(document_input, document_output):
         try:
             if isinstance(document_input, str):
```
ammokhov, line 270: I am curious if this check is enough?

like if it is `str` then do regex match, otherwise just compare `==`. Should we have something more strict? what other alternatives there could be?

anshikg: As far as I know, regex match is only going to happen in case of a string in any other primitive type the value should be compared with ```===``` as we will have the exact value from the function.

PatMyron: I could see number inputs needing `propertyTransform` in addition to string inputs

anshikg: I think I didn't explain well, let me take another stab at it. I understand that propertyTransform can be used for all primitive types namely: bool, string, number etc.
My point above was if a regex string is formed for any primitive type. The output for```if isinstance(document_input, str):```  where, document_input is regex will always be true. Therefore in case of any primitive type if a regex is formed we will execute the regex match. Does this make more sense?

ammokhov (approved):

PatMyron (approved):

Issue #527, opened 2020-08-04
More public documentation/guidance on each contract test


ninavolkmuth:
We provide a contract for handlers to follow in the docs, but it will be useful for providers to know what each individual test within the contract tests is trying to accomplish, what the inputs are for the test, and what the expected outcome is without having to read through the code of the test. Might also be useful to have more context on the test failures when we're reporting the results as well (ie why it's bad the test is failing, suggestions on how to fix it).

advaj:
+1 - As someone who is migrating resources at scale. From personal experience, we have come up with some notes like - But if the tests have suggestions and what has failed specifically, that would be very useful. 

```
1. Delete Handlers should not return a Model. So, when the Handler is returning a Success Progress Event, make sure that the Resource Model attribute is set to null. 
2. The Update, Read and Delete Handlers should contain a check for the existence of a Primary Identifier in the Model. If the Primary Identifier doesn’t exist, then throw a NotFound Error Code. 
3. If your Schema contains create only properties, make sure you include a check in the Update Handler for verifying if these properties are not being changed. 
```

Similarly, most contract tests check for something specific like this. Hence, some documentation about what each test is checking for and a potential resolution would be very useful. It would save us a ton of time as well. In some cases, it is taking us more time to fix contract tests than write the handler code.

PatMyron:
The contract test names themselves need to be much more descriptive. Needed to fix my `ListHandler` to get `create_delete` to pass, which is not intuitive, even if other more thorough documentation existed

anshikg:
Re-opening this issue. It looks like we missed the documentation release here

PatMyron:
> 3. If your Schema contains create only properties, make sure you include a check in the Update Handler for verifying if these properties are not being changed. 

Why should every `UpdateHandler` have to handle this for every `createOnlyProperty`? I thought the point of [`createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics) was so we didn't have to worry about this

> 2. The Update, Read and Delete Handlers should contain a check for the existence of a Primary Identifier in the Model. If the Primary Identifier doesn’t exist, then throw a NotFound Error Code. 

Wondering the same for this as well. If this always must be done, why rely on all of these Handlers implementing this correctly anyway? This seems better handled above the individual resource provider level

ninavolkmuth:
[...]

PR #721, opened 2021-04-01
removing contract_update_create_only_property

PatMyron:
https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/120#pullrequestreview-625828185
https://github.com/aws-cloudformation/cloudformation-cli/issues/527#issuecomment-680696823

### DIFF
--- a/src/rpdk/core/contract/suite/handler_update_invalid.py
+++ b/src/rpdk/core/contract/suite/handler_update_invalid.py
@@ -11,41 +11,6 @@ from rpdk.core.contract.suite.contract_asserts import failed_event
 
 @pytest.mark.update
 @failed_event(
-    error_code=(HandlerErrorCode.NotUpdatable, HandlerErrorCode.NotFound),
-    msg="An update hander MUST return FAILED with a NotUpdatable error code \
-        if the difference between the previous state and the properties included \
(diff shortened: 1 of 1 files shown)


ammokhov (commented):
I think it's a good point. We intend to have schema as a single source of truth. Handlers really should not throw not updatable exception on immutable change. This defeats the purpose of the field. I am ok removing this tests. 

@aygold92 @jotompki @wbingli thoughts?

anshikg (changes requested):
I agree we need to remove this contract test. Lets remove this once we have closure on how we want to handle this check in the platform: https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/pull/120

aygold92:
> I think it's a good point. We intend to have schema as a single source of truth. Handlers really should not throw not updatable exception on immutable change. This defeats the purpose of the field. I am ok removing this tests.
> 
> @aygold92 @jotompki @wbingli thoughts?

Right.  Once we introduce conditionalCreateOnlyProperties, we should have essentially an identical contract test for that to replace this one.

aygold92 (approved):

anshikg (approved):

PR #723, opened 2021-04-06
contract test integration test (Python resource provider)

PatMyron:
expanding test beyond only [Java resource provider](https://github.com/aws-cloudformation/cloudformation-cli/pull/714) using [`AWS::FraudDetector::Outcome`](https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-frauddetector/blob/master/aws-frauddetector-outcome/aws-frauddetector-outcome.json)

### DIFF
--- a/.github/workflows/contract-integration.yaml
+++ b/.github/workflows/contract-integration.yaml
@@ -12,6 +12,10 @@ jobs:
         with:
           repository: aws-cloudformation/aws-cloudformation-resource-providers-lookoutvision
           path: lookoutvision
+      - uses: actions/checkout@v2
+        with:
+          repository: aws-cloudformation/aws-cloudformation-resource-providers-frauddetector
+          path: frauddetector
       - uses: actions/setup-java@v1
         with:
           java-version: 8
@@ -25,7 +29,7 @@ jobs:
           unzip -q sam.zip -d sam-install
           sudo ./sam-install/install
           sam local start-lambda &
-          pip install --quiet cloudformation-cli-java-plugin
+          pip install --quiet cloudformation-cli-java-plugin cloudformation-cli-python-plugin
           mvn package --no-transfer-progress -Dmaven.test.skip=true
           cfn test --enforce-timeout 240
           ! grep -q -E -i '(accessKeyId|secretAccessKey|sessionToken)' rpdk.log*
--- a/.github/workflows/contract-integration.yaml
+++ b/.github/workflows/contract-integration.yaml
@@ -4,7 +4,7 @@ on:
   pull_request:
     branches: [ master ]
 jobs:
-  contract-integration:
+  contract-integration-java:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


.github/workflows/contract-integration.yaml
```diff
@@ -20,16 +20,42 @@ jobs:
...
+        with:
+          python-version: 3
+      - run: |
+          pip install -e .
+          cd frauddetector/*outcome
+          curl -s -L https://github.com/aws/aws-sam-cli/releases/latest/download/aws-sam-cli-linux-x86_64.zip -o sam.zip
+          unzip -q sam.zip -d sam-install
+          sudo ./sam-install/install
+          sam local start-lambda &
+          pip install --quiet cloudformation-cli-java-plugin cloudformation-cli-python-plugin
```
privetjr, line 54: I see this uses `actions/setup-python@v2`, and not `actions/setup-java@v1`. So, does this need to `pip install ... cloudformation-cli-java-plugin` at all?

.github/workflows/contract-integration.yaml
```diff
@@ -20,16 +20,42 @@ jobs:
...
       - run: |
           pip install -e .
-          cd provider/*project
+          cd lookoutvision/*project
           curl -s -L https://github.com/aws/aws-sam-cli/releases/latest/download/aws-sam-cli-linux-x86_64.zip -o sam.zip
           unzip -q sam.zip -d sam-install
           sudo ./sam-install/install
           sam local start-lambda &
-          pip install --quiet cloudformation-cli-java-plugin
+          pip install --quiet cloudformation-cli-java-plugin cloudformation-cli-python-plugin
```
privetjr, line 28: Does this need both the java plugin and the python plugin?

PatMyron: Neither of these jobs need both plugins. Just kept the same line in both jobs in the hopes of being able to re-use some code next time we continue expanding the languages tested

.github/workflows/contract-integration.yaml
ammokhov, line 54: why do you need any of these? we can directly install dependencies like docker but to execute the resource we probably dont need plugin. Maybe `cloudformation-cli-python-lib` but even without it seems to be working

ammokhov (commented):
overall this is great addition but i am curious what happens if latest commits break contract tests in the middle of the execution and we have leaked resoruces. Should we clean them up?

.github/workflows/contract-integration.yaml
PatMyron: Java resource provider couldn't build until `cloudformation-cli-java-plugin` was installed. Didn't test, but assumed Python resource provider might need more than `cloudformation-cli` installed as well 🤷‍♂️

PatMyron: `rpdk.log` after `cfn submit --dry-run` fails for Python resource provider with only `cloudformation-cli` installed:
```shell
Traceback (most recent call last):
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/cli.py", line 100, in main
    args.command(args)
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/submit.py", line 14, in submit
    project.load()
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/project.py", line 412, in load
    self.load_settings()
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/project.py", line 201, in load_settings
    self.validate_and_load_resource_settings(raw_settings)
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/project.py", line 230, in validate_and_load_resource_settings
    self._plugin = load_plugin(raw_settings["language"])
  File "/workspace/.pip-modules/lib/python3.8/site-packages/rpdk/core/plugin_registry.py", line 27, in load_plugin
    return PLUGIN_REGISTRY[language]()()
KeyError: 'python37'
```

ammokhov: oh yea, that's right we need that for cfn generate etc

ammokhov (approved):

omkhegde (approved):

Issue #731, opened 2021-04-14
`additionalIdentifiers` in schema cause update example generation failure.


ptrucks:
When generating an update example for contract tests, the resource client looks up unique keys from the resource model, however if there are `additionalIdentifiers` the check (see [here](https://github.com/aws-cloudformation/cloudformation-cli/blob/a143ded46fe47690884eb1fb91659e3df396e3f5/src/rpdk/core/contract/resource_client.py#L420)) will fail when passing a set of tuples to [`fragment_list`](https://github.com/aws-cloudformation/cloudformation-cli/blob/a143ded46fe47690884eb1fb91659e3df396e3f5/src/rpdk/core/jsonutils/pointer.py#L100) which will try to call `split` on each tuple in the set, rather than on each string in the tuple for a correct input.

It looks like there is no test case for a schema with `additionalIdentifiers`, so this was not caught.

I will create a PR to fix this issue if that would be accepted.

ptrucks:
I opened a PR to fix the problematic check [here](https://github.com/aws-cloudformation/cloudformation-cli/pull/732).

anshikg:
Closing the issue as the code change has been pushed

PR #732, opened 2021-04-14
Fix update generation with additionalIdentifiers

ptrucks:
If the `additionalIdentifiers` property is present in a resource schema
the `generate_update_example` method in `resource_client` will not work,
because it checks for unique keys from the create_model and passes
`self._additional_identifiers_paths` to a method that expects a list of
tuples of string, whereas it receives a list of sets of tuples of
string and will fail with

```
AttributeError: 'tuple' object has no attribute 'split'
```

Instead, check if the key is in any of the additional identifiers.

Related issue
-------------

https://github.com/aws-cloudformation/cloudformation-cli/issues/731

*Issue #, if available:*

*Description of changes:*

### DIFF
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -80,6 +80,19 @@ SCHEMA_WITH_COMPOSITE_KEY = {
     "primaryIdentifier": ["/properties/c", "/properties/d"],
 }
 
+SCHEMA_WITH_ADDITIONAL_IDENTIFIERS = {
+    "properties": {
+        "a": {"type": "number"},
+        "b": {"type": "number"},
+        "c": {"type": "number"},
+        "d": {"type": "number"},
+    },
+    "readOnlyProperties": ["/properties/b"],
+    "createOnlyProperties": ["/properties/c"],
+    "primaryIdentifier": ["/properties/c"],
+    "additionalIdentifiers": [["/properties/b"]],
+}
+
 
 @pytest.fixture
 def resource_client():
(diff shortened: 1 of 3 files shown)


ptrucks:
Please note that contract tests are failing because the workflow can't access the secrets to retrieve credentials, not because the tests themselves are failing.

tests/contract/test_resource_client.py
```diff
@@ -80,6 +80,43 @@
     "primaryIdentifier": ["/properties/c", "/properties/d"],
 }
 
+SCHEMA_WITH_TRANSFORM = {
```
prerna-p, line 83: Why is this needed?

ptrucks: This is not part of my change, but got pulled into the diff due to a merge when rebasing.

ptrucks: Removed the extraneous lines from the commit, sorry for the noise.

prerna-p (approved):

tests/contract/test_resource_client.py
```diff
@@ -202,6 +215,51 @@ def resource_client_inputs_schema():
...
+                EMPTY_OVERRIDE,
+                {
+                    "CREATE": {"a": 111, "c": 2, "d": 3},
+                    "UPDATE": {"a": 1, "c": 2},
+                    "INVALID": {"c": 3},
+                },
+            )
+
+    mock_sesh.client.assert_called_once_with("lambda", endpoint_url=endpoint)
+    mock_creds.assert_called_once_with(mock_sesh, LOWER_CAMEL_CRED_KEYS, None)
```
anshikg, line 252: Can we reuse an existing resource client vs creating a new one?

ptrucks: Sure, I was following the style of the other tests at the time but it seems like that has been changed since. I'll update to the new style.

ptrucks: I have addressed this by parameterizing the fixture, let me know if that works for you.

anshikg (approved):

PR #741, opened 2021-04-27
Add support for nested comparison

anshikg:
*Issue #, if available:* 

*Description of changes:* This CR avoids comparing nested properties which were not present in the input model provided to the handler. This is done to mimic the drift functionality and avoid failing contract test for expected behavior.
In the below CR I am recursively checking the model of the input and output and comparing the properties present in input with output model and asserting only if a property present in input model is not present/changed in the output model.
For now, the functionality for un order types has been removed as this not supported in the back end service.

*Test:* Ran contract test for resources which failed this condition
```
platform darwin -- Python 3.8.9, pytest-6.2.3, py-1.9.0, pluggy-0.13.1 -- /Users/anshikg/workspace/cloudformation-cli/env/bin/python3
cachedir: .pytest_cache
hypothesis profile 'default' -> database=DirectoryBasedExampleDatabase('/Users/anshikg/workspace/aws-cloudformation-resource-providers-kinesisfirehose/deliverystream/.hypothesis/examples')
rootdir: /Users/anshikg/workspace/aws-cloudformation-resource-providers-kinesisfirehose/deliverystream, configfile: ../../../../../private/var/folders/zh/rhw_046j7bq4d88hjn671hqjy323j8/T/pytest_952xexz8.ini
plugins: random-order-1.0.4, localserver-0.5.0, cov-2.10.0, hypothesis-5.19.3, black-0.3.10
collected 14 items                                                                                                                                                                                                                                                                                                                                                                                                                                                                                    

handler_create.py::contract_create_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                
handler_create.py::contract_create_duplicate 	PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             
handler_create.py::contract_create_read_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          
handler_create.py::contract_create_list_success PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                          
handler_delete.py::contract_delete_read PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  
handler_delete.py::contract_delete_list PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                  
handler_delete.py::contract_delete_delete PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                
handler_delete.py::contract_delete_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                                
handler_misc.py::contract_check_asserts_work PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                             
handler_read.py::contract_read_without_create PASSED                                                                                                                                                                                                                                                                                                                                                                                                                                            

===========================================================================================
```


### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -334,6 +334,29 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         example = override_properties(self.invalid_strategy.example(), overrides)
         return {**create_model, **example}
 
+    def compare(self, inputs, outputs):
+        assertion_error_message = (
+            "All properties specified in the request MUST "
+            "be present in the model returned, and they MUST"
+            " match exactly, with the exception of properties"
+            " defined as writeOnlyProperties in the resource schema"
+        )
+        try:
+            for key in inputs:
+                if isinstance(inputs[key], dict):
+                    self.compare(inputs[key], outputs[key])
+                elif isinstance(inputs[key], list):
+                    assert len(inputs[key]) == len(outputs[key])
+                    self.compare_list(inputs[key], outputs[key])
+                else:
+                    assert inputs[key] == outputs[key], assertion_error_message
+        except KeyError as e:
+            raise AssertionError(assertion_error_message) from e
+
+    def compare_list(self, inputs, outputs):
+        for index in range(len(inputs)):  # pylint: disable=C0200
+            self.compare(inputs[index], outputs[index])
+
     @staticmethod
     def key_error_safe_traverse(resource_model, write_only_property):
         try:
--- a/src/rpdk/core/contract/suite/handler_commons.py
+++ b/src/rpdk/core/contract/suite/handler_commons.py
@@ -163,32 +163,7 @@ def test_input_equals_output(resource_client, input_model, output_model):
     pruned_output_model = prune_properties_if_not_exist_in_path(
         pruned_output_model, pruned_input_model, resource_client.create_only_paths
     )
-
-    assertion_error_message = (
-        "All properties specified in the request MUST "
-        "be present in the model returned, and they MUST"
-        " match exactly, with the exception of properties"
-        " defined as writeOnlyProperties in the resource schema"
+    resource_client.compare(
+        pruned_input_model,
+        pruned_output_model,
     )
-    # only comparing properties in input model to those in output model and
-    # ignoring extraneous properties that maybe present in output model.
-    try:
-        for key in pruned_input_model:
-            if key in resource_client.properties_without_insertion_order:
-                assert test_unordered_list_match(
-                    pruned_input_model[key], pruned_output_model[key]
-                )
-            else:
-                assert (
-                    pruned_input_model[key] == pruned_output_model[key]
-                ), assertion_error_message
-    except KeyError as e:
-        raise AssertionError(assertion_error_message) from e
-
-
-def test_unordered_list_match(inputs, outputs):
-    assert len(inputs) == len(outputs)
-    try:
-        assert all(input in outputs for input in inputs)
-    except KeyError as exception:
-        raise AssertionError("lists do not match") from exception
--- a/tests/contract/test_resource_client.py
+++ b/tests/contract/test_resource_client.py
@@ -68,6 +68,32 @@ SCHEMA_ = {
     "writeOnlyProperties": ["/properties/d"],
 }
 
+SCHEMA_WITH_NESTED_PROPERTIES = {
+    "properties": {
+        "a": {"type": "string"},
+        "g": {"type": "number"},
+        "b": {"$ref": "#/definitions/c"},
+        "f": {
+            "type": "array",
+            "items": {"$ref": "#/definitions/c"},
+        },
+        "h": {
+            "type": "array",
+            "insertionOrder": "false",
+            "items": {"$ref": "#/definitions/c"},
+        },
+    },
+    "definitions": {
+        "c": {
+            "type": "object",
+            "properties": {"d": {"type": "integer"}, "e": {"type": "integer"}},
+        }
+    },
+    "readOnlyProperties": ["/properties/a"],
+    "primaryIdentifier": ["/properties/a"],
+    "writeOnlyProperties": ["/properties/g"],
+}
+
 SCHEMA_WITH_COMPOSITE_KEY = {
     "properties": {
         "a": {"type": "number"},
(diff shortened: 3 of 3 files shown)


tests/contract/test_resource_client.py
```diff
@@ -1210,3 +1236,52 @@ def test_generate_update_example_with_composite_key(
...
+def test_compare_should_pass(resource_client):
+    resource_client._update_schema(SCHEMA_WITH_NESTED_PROPERTIES)
+    inputs = {"b": {"d": 1}, "f": [{"d": 1}], "h": [{"d": 1}, {"d": 2}]}
+
+    outputs = {
+        "b": {"d": 1, "e": 3},
+        "f": [{"d": 1, "e": 2}],
+        "h": [{"d": 1, "e": 3}, {"d": 2}],
+    }
+    resource_client.compare(inputs, outputs)
```
ammokhov, line 1250: wanna add any asserts? this test does not really test anything

src/rpdk/core/contract/resource_client.py
```diff
@@ -334,6 +334,29 @@ def generate_invalid_update_example(self, create_model):
         example = override_properties(self.invalid_strategy.example(), overrides)
         return {**create_model, **example}
 
+    def compare(self, inputs, outputs):
```
ammokhov, line 337: *thought: i think we should start using annotations (like `def compare(self, inputs: dict, outputs: dict) -> None:`) or something similar so you know what are the possible inputs.

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -163,32 +163,7 @@ def test_input_equals_output(resource_client, input_model, output_model):
     pruned_output_model = prune_properties_if_not_exist_in_path(
         pruned_output_model, pruned_input_model, resource_client.create_only_paths
     )
-
```
ammokhov, line 166: I cant attach a comment to the code above but I wanna make sure that `input_model` is the provided template and not the resource model from previous action, isnt it?

src/rpdk/core/contract/resource_client.py
```diff
@@ -334,6 +334,29 @@ def generate_invalid_update_example(self, create_model):
...
+            "be present in the model returned, and they MUST"
+            " match exactly, with the exception of properties"
+            " defined as writeOnlyProperties in the resource schema"
+        )
+        try:
+            for key in inputs:
+                if isinstance(inputs[key], dict):
+                    self.compare(inputs[key], outputs[key])
+                elif isinstance(inputs[key], list):
+                    assert len(inputs[key]) == len(outputs[key])
```
ammokhov, line 349: is that gonna be true for all cases?

```
input = {
  "key_1": [ {"a": 10}, {"b": 11} ]
},
output = {
  "key_1": [ {"a": 10}, {"b": 11}, {"c": 12} ]
}
```
this will fail assertion but `c` is absent from input - based on cr description this should pass assertion

src/rpdk/core/contract/resource_client.py
```diff
@@ -334,6 +334,29 @@ def generate_invalid_update_example(self, create_model):
...
+            "All properties specified in the request MUST "
+            "be present in the model returned, and they MUST"
+            " match exactly, with the exception of properties"
+            " defined as writeOnlyProperties in the resource schema"
+        )
+        try:
+            for key in inputs:
+                if isinstance(inputs[key], dict):
+                    self.compare(inputs[key], outputs[key])
+                elif isinstance(inputs[key], list):
```
ammokhov, line 348: is `list` safe to go with (no sets)? should you change it to `Collection`?

src/rpdk/core/contract/resource_client.py
```diff
@@ -334,6 +334,29 @@ def generate_invalid_update_example(self, create_model):
...
 
+    def compare(self, inputs, outputs):
+        assertion_error_message = (
+            "All properties specified in the request MUST "
+            "be present in the model returned, and they MUST"
+            " match exactly, with the exception of properties"
+            " defined as writeOnlyProperties in the resource schema"
+        )
+        try:
+            for key in inputs:
```
ammokhov, line 345: what if there as scalar values?

src/rpdk/core/contract/suite/handler_commons.py
anshikg: could you explain what you mean by ```resource model from previous action``` are you referring to response model of an action ? because when we compare for Update we do call the same function. This is an example: https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/suite/handler_update.py#L33

tests/contract/test_resource_client.py
anshikg: ```compare``` method is asserting on the values. If the input and output model is not equal we should getting assertion exception. Could you suggest an assert message which might be right for this test?

src/rpdk/core/contract/resource_client.py
anshikg: in my opinion it should be list. As we are dealing with an order list of items ```[]``` vs set being an unordered list ```{}``` 
Is there a specific example where you might see this to be probromatic?

tests/contract/test_resource_client.py
```diff
@@ -1210,3 +1236,52 @@ def test_generate_update_example_with_composite_key(
...
+
+    outputs = {
+        "b": {"d": 1, "e": 2},
+        "f": [{"d": 1}],
+        "h": [{"d": 1}],
+    }
+    try:
+        resource_client.compare(inputs, outputs)
+    except AssertionError:
+        logging.debug("Expected")
```
omkhegde, line 1265: Nit: would prefer if we have a more meaningful message here. Maybe just something like `Expected assertion in <testName>..`

omkhegde (approved):
Minor comment. Otherwise LGTM.

tests/contract/test_resource_client.py
anshikg: makes sense. Let me update that

src/rpdk/core/contract/resource_client.py
ammokhov: ah I see, the it's self-referenced on complex types only

src/rpdk/core/contract/resource_client.py
ammokhov: I think since it's json it's always either dict or list

src/rpdk/core/contract/resource_client.py
```diff
@@ -334,6 +334,29 @@ def generate_invalid_update_example(self, create_model):
...
+                if isinstance(inputs[key], dict):
+                    self.compare(inputs[key], outputs[key])
```
ammokhov, line 347: could you use [`nested_lookup`](https://pypi.org/project/nested-lookup/) instead of writing plain dfs?

src/rpdk/core/contract/resource_client.py
ammokhov: but then it will fail before the actual assertion..

src/rpdk/core/contract/suite/handler_commons.py
ammokhov: gotcha, yea I meant to make sure it's not a response from the handler we comparing

ammokhov (approved):
as chatted offline, the change looks good. we can improve it as a follow up PR

PR #750, opened 2021-05-03
removing duplicate module validation caught by cfn-lint

PatMyron:
https://github.com/aws-cloudformation/cloudformation-cli/pull/634
https://github.com/aws-cloudformation/cloudformation-cli/pull/644

left module-specific validation like restrictions on [exports](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-stack-exports.html), [macros/transforms](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html), and [nested stacks](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/using-cfn-nested-stacks.html)


src/rpdk/core/fragment/generator.py
```diff
@@ -91,24 +77,7 @@ def __validate_no_exports_present(raw_fragments):
...
-                raise FragmentValidationError(
-                    "The Module template fragment has "
-                    + str(output_count)
-                    + " outputs but must not exceed the limit of "
-                    + str(self.output_limit)
-                    + " outputs"
-                )
-
     def __validate_resources(self, raw_fragments):
-        if "Resources" not in raw_fragments:
```
MalikAtalla-AWS, line 107: Validations like this have been added to avoid parsing errors that are impossible to understand by the user.  I'm not sure if we're regressing on the user experience here.

MalikAtalla-AWS (changes requested):

src/rpdk/core/fragment/generator.py
PatMyron: `cfn-lint` will throw `E1001 Missing top level template section Resources` for this validation

miparnisari: @PatMyron For all the validations that you removed in this PR can you show us what it looks like right now and what it will look like after? Wanna make sure that the error messages in cfn-lint are on par with the ones in the CLI

miparnisari:
The `fragments` folder cannot have more than 1 json or yaml file. Can we add this validation? Otherwise it'll fail at runtime during registration

MalikAtalla-AWS:
That validation should already exist, Maria. I had added that a while ago.

miparnisari:
> That validation should already exist, Maria. I had added that a while ago.

I was able to submit a module today with two files inside `fragments`: `s3.json` and `blabla.json.old`

miparnisari (approved):

kddejong (approved):

mircealam:
@MalikAtalla-AWS @PatMyron does this change still apply and/or what rework is needed here?

MalikAtalla-AWS:
My comments are addressed, but the build is failing @mircealam

PatMyron:
> My comments are addressed, but the build is failing @mircealam

looks like that test was removed in https://github.com/aws-cloudformation/cloudformation-cli/pull/1028 already

PR #754, opened 2021-05-06
Add contract tests cloudwatch log delivery support.

AugustXiaodm:
*Issue #, if available:* N/A

*Description of changes:*
This change added two new options for cfn test: `--log-group-name` and `--log-role-arn`. When both of these two options are specified with valid values, contract tests generated handler logs will be delivered into the destination log group using credentials from the specified log role.


yualexan (approved):

src/rpdk/core/contract/resource_client.py
```diff
@@ -435,8 +444,13 @@ def make_request(
             "action": action,
             "callbackContext": callback_context,
             "bearerToken": token,
+            "resourceType": type_name,
             **kwargs,
         }
+        if log_group_name and log_creds:
+            request_body["requestData"]["providerCredentials"] = log_creds
```
anshikg, line 451: please make sure you are redacting creds before logging

README.md
```diff
@@ -56,6 +56,7 @@ cfn test -- -k contract_delete_update #to run a single test
 cfn test --tb=long #exhaustive, informative traceback formatting
 cfn test --enforce-timeout 60 #set the RL handler timeout to 60 seconds and CUD handler timeout to 120 seconds.
 cfn test --enforce-timeout 60 -- -k contract_delete_update # combine args
+cfn test --log-group-name cw_log_group --log-role-arn log_delivery_role_arn # Handler logs generated by contract tests will be delivered to the specified cw_log_group using the credentials from log_delivery_role_arn
```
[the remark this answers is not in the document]
anshikg: If this is creating the log group if one does not exist this is a non issue

src/rpdk/core/contract/resource_client.py
```diff
@@ -419,11 +425,14 @@ def make_request(
         account,
         action,
         creds,
+        type_name,
```
ammokhov, line 428: the change makes sense, I am a bit unsure why do you need type_name? could you provide more details?

AugustXiaodm: Type name will be used for constructing cloudwatch metric name if there are valid logging credential passed in. Without this, wrapper side will fail for NPE when doing so.

README.md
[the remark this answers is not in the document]
aygold92: I imagine that we would want to at least encourage, if not assume, that providers will always want logging.  Can we make specifying this easier or perhaps even not required?

Some options are to have them specify it in the rpdk.config, create it for them, have a flag to create it for them.  We could also just use the AWS credentials of whoever is running the command, or generate a log group name if they don't specify it

aygold92 (changes requested):

PatMyron:
Do we want something similar for `cfn validate`?

PR #766, opened 2021-05-19
Use deep copy for model compare

wbingli:
*Issue #, if available:*

*Description of changes:*.   Currently it uses shallow copy for model comparison. However,  for any read only property is nested, it will modify the actual model during the comparison.  For example, if a primary ID is nested property,  the contract test will break as the primary id will be pruned in this method.

### Testing
Verified using a nested primary ID, now the primary Id can pass to delete handler correctly.

### DIFF
--- a/src/rpdk/core/contract/suite/handler_commons.py
+++ b/src/rpdk/core/contract/suite/handler_commons.py
@@ -1,5 +1,5 @@
-import logging
 import copy
+import logging
 
 from rpdk.core.contract.interface import Action, HandlerErrorCode, OperationStatus
 from rpdk.core.contract.resource_client import (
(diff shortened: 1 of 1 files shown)


anshikg (approved):

src/rpdk/core/contract/suite/handler_commons.py
```diff
@@ -1,4 +1,5 @@
 import logging
+import copy
```
PatMyron, line 2: think `isort` is failing on this
```suggestion
import copy
import logging
```

ammokhov (approved):

src/rpdk/core/contract/suite/handler_commons.py
wbingli: Good catcj. Fixed.

PR #779, opened 2021-06-04
Remove test contract_read_without_create

sungkkim:
*Issue #, if available:*

In https://github.com/aws-cloudformation/cloudformation-cli/pull/515, we switched to using invalid input file for read without create test. This was because the primary id generation strategy for `generate_create_example` was insufficient. As a side effect, I noticed teams mapping InvalidInput exceptions to NotFound in their read handler as a workaround. I don't think this is correct and will just cause confusion.

1. As mentioned in https://github.com/aws-cloudformation/cloudformation-cli/pull/515, `contract_delete_read` should be enough to test read handler returning NotFound for non-existent resources. The comment mentioned a concern for soft-deleting resources (such as KMS::Key marked for deletion state), but that is an edge case and resource providers should mark those transitionary states as deleted. If they don't do this, they would have failed `contract_delete_read` anyway.

2. Even if we use invalid input file, we are still hardcoding primary id. This means for resources where primary id is ARN, the tests will fail across different partitions.

*Description of changes:*

- Remove `contract_read_without_create` test


ammokhov (approved):
I think it should be ok to remove this tests, as we have `create_delete_read` test, which does similar check. 

There might be an argument that during delete at `create_delete_read` some resources take time to delete (e.g kms key takes at least 7d to fully delete), which means they are still readable. However, such readable resource should still have a certain state (e.g. `PENDING_DELETION` etc) which could be used inside read handler to return not found exception

wbingli:
Agree that we have `create_delete_read` test to cover the read after delete.    The `contract_read_without_create` is more duplicated as well as more hard to figure out if a resource indeed not be created.

wbingli (approved):

anshikg (approved):

sungkkim:
AWS creds can't be fetched for now from fork PRs so I'll continue with merging

PR #786, opened 2021-06-22
Test config

anshikg:
*Issue #, if available:*

*Description of changes:* Merging Type configuration support for cfn invoke and contract test


src/rpdk/core/data/schema/provider.definition.schema.v1.json
```diff
@@ -274,65 +87,53 @@
...
         },
         "title": {
-            "$ref": "http://json-schema.org/draft-07/schema#/properties/title"
+            "$ref": "base.definition.schema.v1.json#/properties/title"
         },
         "description": {
             "$comment": "A short description of the resource provider. This will be shown in the AWS CloudFormation console.",
-            "$ref": "http://json-schema.org/draft-07/schema#/properties/description"
+            "$ref": "base.definition.schema.v1.json#/properties/description"
         },
```
PatMyron, line 101: our definitions are still the same anyways, right? why switch these to our own definitions?

ammokhov (approved):

src/rpdk/core/data/schema/cloudformation-resource-type-configuration-schema.v1.json
```diff
@@ -0,0 +1,32 @@
+{
```
srujithpoondla03, line 1: We should not release this with Public Registry launch. For more info check with @rajdivya

src/rpdk/core/data/schema/base.definition.schema.v1.json
```diff
@@ -0,0 +1,302 @@
+{
```
srujithpoondla03, line 1: Can we not add this manually here, instead update https://github.com/aws-cloudformation/cloudformation-cli/blob/master/.github/workflows/schema-updater.yaml. Same with other schema definitions too, we should include in the above action to automatically pull from the resource-schema package.

srujithpoondla03 (changes requested):
Left some feedback on schemas

src/rpdk/core/data/schema/cloudformation-resource-type-configuration-schema.v1.json
anshikg: Ack

srujithpoondla03 (approved):

src/rpdk/core/data/schema/base.definition.schema.v1.json
PatMyron: +1, this reverted commits like @Yunhao-Jiao's https://github.com/aws-cloudformation/cloudformation-resource-schema/pull/123

internal/setup.py
```diff
@@ -0,0 +1,21 @@
...
+    name="AWSCloudFormationRPDK",
+    version="1.0",
+    packages=["rpdk.core"],
+    package_dir={"": "src"},
+    include_package_data=True,
+    entry_points={
+        "console_scripts": ["cfn-cli = rpdk.core.cli:main", "cfn = rpdk.core.cli:main"]
+    },
+    # Not all tests passed in Brazil due to some test dependency versions missing in Brazil
+    # Since all test are run in GitHub/CodeBuild, only verify one test
```
PatMyron, line 13: Should this file be public? Also think test dependencies should be imported into Brazil rather than just disabling tests that depend on them

Issue #811, opened 2021-08-23
Contract tests fail on out-of-order collections when insertionOrder=false


osdrv:
Hey folks,

I'm developing a resource where one of the attributes is an out-of-order collection (practically, a set). The service preserves no specific order on the items in this collection, hence I'm being explicit on setting `insertionOrder=false`:

```json
"Foo": {
  "type": "array",
  "insertionOrder": false,
  "items": {
    "$ref": "#/definitions/Foo"
  }
}
```

My CFN version:
```
cfn --version
cfn 0.2.15
```

My contract test input file declares this collection in a particular order. The service response model implies on no specific order returned.

When I run `cfn test`, I'm getting this error:
```
AssertionError: All properties specified in the request MUST be present in the model returned, and they MUST match exactly, with the exception of properties defined as writeOnlyProperties in the resource schema
```

The tool says nothing what specific object failed the validation, but from rpdk.log I see the order of the input and the response collections is distinct.

I checked the contract test executor code (https://github.com/aws-cloudformation/cloudformation-cli/blob/master/src/rpdk/core/contract/resource_client.py#L351) and it seems it doesn't respect `insertionOrder` attribute:

```python
351     def compare(self, inputs, outputs):
...
360                 for key in inputs:
...
363                     elif isinstance(inputs[key], list):
364                         assert len(inputs[key]) == len(outputs[key])
365                         self.compare_list(inputs[key], outputs[key])
...
373     def compare_list(self, inputs, outputs):
374         for index in range(len(inputs)):
375             self.compare(inputs[index], outputs[index])
```

Would you please assist on this case and let me know if this attribute would be supported by the tool in the future revisions?

Thank you.

ammokhov:
Thanks for filing this issue, it looks that there is an assumption in the code that any collection has an order built-in. Fairly simple change, CFN will prioritize the work withing next couple of weeks. Feel free to contribute as well.

ammokhov:
@osdrv PR is attached to the issue (#812) plz take a look at the change

PR #812, opened 2021-08-24
[BugFix] Collection Comparison for Contract Tests

ammokhov:
*Issue #, if available:* https://github.com/aws-cloudformation/cloudformation-cli/issues/811

*Description of changes:*

Collection Comparison for contract tests. Method `test_input_equals_output` contains a naive comparison of the collections, which generates a false positive for unordered sets.

Change: instead of comparing lists by values, added `compare_collection`. If `insertionOrder` is `False` then transform list items (since it's json each item is scalar/array/json) into a set of hash values, which does naive comparison after that. For hashing I'm using MD5, which is not secure hashing but provides a good hashing for contract test purpose.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -22,7 +22,7 @@ from ..boto_helpers import (
     get_temporary_credentials,
 )
 from ..jsonutils.pointer import fragment_decode, fragment_list
-from ..jsonutils.utils import traverse
+from ..jsonutils.utils import item_hash, traverse
 
 LOG = logging.getLogger(__name__)
 
@@ -362,7 +362,10 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
                         self.compare(inputs[key], outputs[key])
                     elif isinstance(inputs[key], list):
                         assert len(inputs[key]) == len(outputs[key])
-                        self.compare_list(inputs[key], outputs[key])
+                        is_ordered = self._schema["properties"][key].get(
+                            "insertionOrder", True
+                        )
+                        self.compare_collection(inputs[key], outputs[key], is_ordered)
                     else:
                         assert inputs[key] == outputs[key], assertion_error_message
             else:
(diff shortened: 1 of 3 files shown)


tests/contract/test_resource_client.py
```diff
@@ -1373,6 +1373,30 @@ def test_compare_should_throw_exception(resource_client):
         logging.debug("This test expects Assertion Exception to be thrown")
 
 
+@pytest.mark.parametrize(
+    "inputs,outputs,schema_fragment",
+    [
+        (
+            {"CollectionToCompare": ["item1", "item2", "item3"]},
+            {"CollectionToCompare": ["item3", "item2", "item1"]},
+            {"properties": {"CollectionToCompare": {"insetionOrder": False}}},
```
gandhek, line 1382: Typo in insertion. How did the tests work correctly with this Typo!

src/rpdk/core/contract/resource_client.py
```diff
@@ -22,7 +22,7 @@
     get_temporary_credentials,
 )
 from ..jsonutils.pointer import fragment_decode, fragment_list
-from ..jsonutils.utils import traverse
+from ..jsonutils.utils import item_hash, traverse
```
gandhek, line 25: Can we use Counter instead of generating our own hash?

tests/contract/test_resource_client.py
ammokhov: `try-catch` actually caught the exception...

src/rpdk/core/contract/resource_client.py
ammokhov: counter transforms into dictionary but in case collection has complex types (dictionaries) then we wont be able to hash then, that is why using md5 to custom hash dictionaries

gandhek (approved):

src/rpdk/core/contract/resource_client.py
```diff
@@ -362,17 +362,26 @@ def compare(self, inputs, outputs):
                         self.compare(inputs[key], outputs[key])
                     elif isinstance(inputs[key], list):
                         assert len(inputs[key]) == len(outputs[key])
-                        self.compare_list(inputs[key], outputs[key])
+                        is_ordered = self._schema["properties"][key].get(
+                            "insertionOrder", True
```
sungkkim, line 366: Just FYI, we default to False for features like Drift Detection. I am fine with this for the purpose of contract tests though.

sungkkim (approved):

PR #824, opened 2021-09-10
Add contract test for tagging metadata enforcement

Yunhao-Jiao:
*Issue #, if available:*

*Description of changes:*

Customers migrated to new version of tagging metadata will need to go through some new contract tests. However, If "taggable" is marked false, they will be able to skip them.
```
"tagging": {
            "type": "object",
            "properties": {
                "taggable": {
                    "description": "A boolean flag indicating whether this resource type supports tagging.",
                    "type": "boolean",
                    "default": true
                },
                "tagOnCreate": {
                    "description": "A boolean flag indicating whether this resource type supports tagging resources upon creation.",
                    "type": "boolean",
                    "default": true
                },
                "tagUpdatable": {
                    "description": "A boolean flag indicating whether this resource type supports updatable tagging.",
                    "type": "boolean",
                    "default": true
                },
                "cloudFormationSystemTags": {
                    "description": "A boolean flag indicating whether this resource type supports CloudFormation system tags.",
                    "type": "boolean",
                    "default": true
                },
                "tagProperty": {
                    "description": "A reference to the Tags property in the schema.",
                    "$ref": "http://json-schema.org/draft-07/schema#/properties/$ref"
                }
            },
            "required": [
                "taggable"
            ],
            "additionalProperties": false
 }
```

### DIFF
--- a/src/rpdk/core/contract/suite/handler_create.py
+++ b/src/rpdk/core/contract/suite/handler_create.py
@@ -7,7 +7,10 @@ import pytest
 # WARNING: contract tests should use fully qualified imports to avoid issues
 # when being loaded by pytest
 from rpdk.core.contract.interface import Action, OperationStatus
-from rpdk.core.contract.suite.contract_asserts import skip_not_writable_identifier
+from rpdk.core.contract.suite.contract_asserts import (
+    skip_not_taggable,
+    skip_not_writable_identifier,
+)
 from rpdk.core.contract.suite.handler_commons import (
     test_create_failure_if_repeat_writeable_id,
     test_create_success,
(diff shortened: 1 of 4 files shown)


src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -73,3 +76,15 @@ def contract_create_list(created_resource, resource_client):
...
+
+
+@pytest.mark.create
+@skip_not_taggable
+def contract_create_taggable(resource_client):
+    assert (
+        resource_client.metadata_contains_tag_property()
+    ), "Resource marked taggable but tagProperty attribute missing in tagging metadata."
+    input_model = resource_client.generate_create_example()
+    assert resource_client.validate_model_contain_tags(
```
wbingli, line 88: Should we move it as part of the normal `contract_create_delete`  instead of a new contract test case?  

Also if we introduce this new, will it break some customers if they don't include tags in their contract test input data? how should we rollout this change without breaking customers?

src/rpdk/core/contract/suite/handler_create.py
```diff
@@ -73,3 +76,15 @@ def contract_create_list(created_resource, resource_client):
     _input_model, created_model, _request = created_resource
     assert test_model_in_list(resource_client, created_model)
     test_read_success(resource_client, created_model)
+
+
+@pytest.mark.create
+@skip_not_taggable
+def contract_create_taggable(resource_client):
+    assert (
+        resource_client.metadata_contains_tag_property()
```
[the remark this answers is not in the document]
Yunhao-Jiao: yeah agree, have revised this checking

src/rpdk/core/contract/suite/handler_create.py
Yunhao-Jiao: have added @skip_no_tagging so that if tagging metadata is missing, we will skip those new test

Issue #841, opened 2021-10-26
cfn validate doesn't validate missing properties


richardhboyd:
If I have the following snippet in my schema

```json
    "additionalProperties": false,
    "readOnlyProperties": [
        "/properties/TPSCode"
    ],
    "primaryIdentifier": [
        "/properties/TPSCode"
    ],
```

But I haven't actually defined a `TPSCode` in `properties`, I would expect `cfn validate` to fail.

PR #892, opened 2022-07-09
841 Validate properties are defined on cfn validate

mgentry612:
*Issue #, if available:*
[841](https://github.com/aws-cloudformation/cloudformation-cli/issues/841)

*Description of changes:*
***This PR is not ready for production*** - it makes several of the current test cases fail. I wanted to start a conversation about whether all of this validation is needed before I considered modifying current test cases. The main reason I made this PR was to introduce myself to the codebase. I realize this PR may or may not be helpful to your team. If you find that this PR is heading in the right direction, please provide feedback. If you're not interested in adding this functionality, no problem, it was fun diving in.

This PR implements validation for the top level of properties in the resource schema. For the following resource schema keys, a validation function ensures that properties at the top level of the "property" object are defined.
`
        "readOnlyProperties",
        "createOnlyProperties",
        "primaryIdentifier",
        "required",
        "additionalIdentifiers",
        "deprecatedProperties",
        "writeOnlyProperties",
`

The following assumptions are made in these changes, please confirm:
- Properties are case sensitive
- additionalIdentifiers, when defined, are always two dimensional arrays.

These changes handle examples like this:
`
    "createOnlyProperties": [
        "/properties/Name"
    ]
`

but not nested properties like this, see below section about traversing the json structure:
`
    "createOnlyProperties": [
        "/properties/obj1/obj2/str1"
    ]
`

As mentioned above, this PR omits nested properties. I started diving into nested properties but I decided I would need discuss the requirements before doing any more work. There were several complexities which I had questions about including the array data type and circular references. If you think it would be helpful to show you what I have so far for this on another branch, let me know and I'll share.

I've included several new test cases which run successfully only in isolation at this time.
`pytest --cov="rpdk.core" --doctest-modules --random-order-bucket="parent" "tests/test_data_loaders.py" -k test_load_resource_spec_property_not_defined --pyargs "rpdk.core" --verbose`

### DIFF
--- a/src/rpdk/core/data_loaders.py
+++ b/src/rpdk/core/data_loaders.py
@@ -390,7 +390,7 @@ def verify_top_level_properties_are_defined(resource_spec, schema_key):
 
     properties = get_properties_from_schema_key(resource_spec, schema_key)
 
-    nested_properties = filter(lambda property: "/" in property, properties)
+    nested_properties = set(filter(lambda property: "/" in property, properties))
     top_level_properties = properties - nested_properties
 
     if len(top_level_properties):


src/rpdk/core/data_loaders.py
```diff
@@ -368,6 +370,54 @@ def load_resource_spec(resource_spec_file):  # pylint: disable=R # noqa: C901
 
     return inlined
 
+def verify_listed_properties_are_defined(resource_spec):
+    LOG.debug("verify_listed_properties_are_defined")
+
+    schema_keys = [
```
PatMyron, line 376: Some newer JSON pointer property types
```suggestion
    schema_keys = [
        "nonPublicProperties",
        "conditionalCreateOnlyProperties",
```
https://github.com/aws-cloudformation/cloudformation-resource-schema/blob/master/src/main/resources/schema/provider.definition.schema.v1.json

mircealam:
To get all the validations included and keep them up to date over time is not a trivial task. May I suggest you give https://github.com/aws-cloudformation/resource-schema-guard-rail a try and see if it helps solve your problems? 
In the medium to long term I believe integrating guard rail into the cli is the way to go to keep the validations consistent.

PR #901, opened 2022-09-21
Add custom property arrayType

gsurbhi:
arrayType is only applicable for property of type array. When set to AttributeList, it indicates that the array is of nested type objects, and when set to Standard it indicates that the array is of primitive types


benbridts:
The changes seem unrelated to the title / description?

gsurbhi:
Incorrect pull request

PR #911, opened 2022-10-17
Add python 3.9 and 3.10 support

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add support for Python 3.9 and 3.19 for the cloudformation-cli 


tests/test_type_schema_loader.py
```diff
@@ -135,7 +135,7 @@ def test_load_type_schema_from_file(loader):
         type_schema = loader.load_type_schema(TEST_TARGET_SCHEMA_FILE_PATH)
 
     assert_dict_equals(TEST_TARGET_SCHEMA, type_schema)
-    mock_path_is_file.assert_called_with(TEST_TARGET_SCHEMA_FILE_PATH)
+    mock_path_is_file.assert_any_call(TEST_TARGET_SCHEMA_FILE_PATH)
```
kddejong, line 138: not sure why but the call list for `isFile` also includes `rpdk.log`.  (in Python 3.9+) This would only be called deep in the configuration of the logger.  So not sure why that is getting patched here as it shouldn't be. 

Also the call to `rpdk.log` when running pytest -k the tests pass.  It will only fail on a full test run using `pytest tests/`

kddejong: Switched it to any call because I'm not sure it matters as long as the correct isFile is in the list.

mmaeng (approved):

cmgorton (approved):

Issue #286, opened 2019-09-11
Support `package` for cfn-cli


wbingli:
For a resource development CI/CD pipeline,  customer need to make a RPDK package only during build time. 

Currently `cf-cli submit --dry-run` will generate a package without submit. But it would be great to have `package`  or `build` command to do only build and package to generate artifact.

richardhboyd:
`submit —dry-run` implies that it will also validate if I have the appropriate permissions to execute the submission. I believe `package` more accurately reflects what the cfn CLI is actually doing.

ikben:
This is very similar to https://github.com/aws-cloudformation/cloudformation-cli/issues/368

richardhboyd:
I think the difference between the two is the name of the command. `cfn submit --dry-run` makes a user think it will validate that a bucket exists with the right permissions and the current user has sufficient privileges to perform a submit. `package` is an entirely local operation that should be able to execute even without IAM credentials.

PR #913, opened 2022-10-18
feature(cli) - Support `package` as an alias for the `submit --dry-run` argument in the cfn-cli 

mmaeng:
Issue #286 

Request to create a `package` argument which would be an alias to the `submit --dry-run` argument.  This would be to more accurately describe what is occurring during a `cfn submit --dry-run`


src/rpdk/core/package.py
```diff
@@ -0,0 +1,27 @@
+"""This is the same as cfn submit --dryrun, it will create a local package without uploading.
+
+Projects can be created via the 'init' sub command.
```
kddejong, line 3: ?

tests/test_package.py
```diff
@@ -0,0 +1,17 @@
...
+from rpdk.core.project import Project
+
+
+def test_package_command_valid_schema(capsys):
+    mock_project = Mock(spec=Project)
+
+    with patch("rpdk.core.package.Project", autospec=True, return_value=mock_project):
+        main(args_in=["package"])
+
+    mock_project.load.assert_called_once_with()
```
kddejong, line 13: This seems light doesn't it?  Is this actually called twice and should we evaluate the parameters?

mmaeng: Added more explicit tests

mircealam:
@mmaeng could you please update this to solve conflicts? thank you.

mircealam (approved):

kddejong (approved):

PR #918, opened 2022-10-18
chore(build): Update workflow to not fail-fast and continue with other builds even if 1 fails

### LINKED ISSUE #917: Github Action workflows canceling all builds after first failure.
Looks like the fail-fast workflow strategy is set for true by default.  This is causing all builds to get canceled after the first failure.  While builds are in an inconsistent state fail-fast should be disabled so we can re-run just the failed builds.

mmaeng:
Closes #917 

By default github actions fail fast, where if one build fails it cancels all other in progress builds.  Current build process is somewhat inconsistent and until that is resolved disabling fail-fast so we can just rerun failed builds.

### DIFF
--- a/.github/workflows/pr-ci.yaml
+++ b/.github/workflows/pr-ci.yaml
@@ -13,6 +13,7 @@ jobs:
     env:
       AWS_DEFAULT_REGION: us-east-1
     strategy:
+      fail-fast: false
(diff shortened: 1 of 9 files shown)


mmaeng:
Includes unintended commits.  Will open new PR

Issue #118, opened 2018-11-20
Resource generation outstanding issues


johnttompkins:
The current resource generator in #104 handles simpler cases of resource generation. The below items are necessary to bring the generator to a more operable state that can handle most if not all valid resource definition schemas. 
- [x] Refs are not handled. This is fine, but needs some kind of step prior to make sure it works
- [ ] Type keywords. Keywords that can be specified in jsonschema draft 07 for primitive types should be accounted for in resource generation. #104 handles some of this but not all (e.g. minLength and maxLength is not supported for arrays).
- [ ] Assuming type is a string. We assume the schema type is a string and cannot be an array. Arrays are valid values for type keys and we need to generate values based off all of the types listed.
- [ ]  String formats. jsonschema has a "format" keyword that allows for specific string formats like json pointers, domain names, email addresses, etc. that should be able to be generated by the resource generator. The logic to do this is in place, but we need to construct regular expression strategies that we can map to format names.
- [x] Options to generate more than the required properties. Currently the generator only generates those that are required. Having an option to generate a maximal set of properties would be beneficial for testing the robustness of handlers.
- [x] Combiners. Currently allOf and anyOf are not supported by the generator. anyOf should be a simple addition because it is handled in the same way as oneOf. allOf will require to merge the child schemas. 
- [x]  Const and enum. Constant values and enum schemas are not supported currently.
- [x] Needs to only generate a resource blob (i.e. the stuff under properties)
- [x] Refactor generation functions for readability. 
- [ ] https://github.com/aws-cloudformation/cloudformation-cli/issues/287

Also, some research needs to be done to see if the flattener and resource generator can be combined in some way. They are both walking the schema in a similar way, so it would be nice to remove duplicate implementations of this to lessen the amount of code to be maintained.

iann0036:
Relating to type is string / type keywords, here's the trace for when type is an array of simpleTypes:

```[2019-12-09T03:35:45Z] DEBUG    - Writing Types
[2019-12-09T03:35:45Z] DEBUG    - Unhandled exception
Traceback (most recent call last):
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/cli.py", line 98, in main
    args.command(args)
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/generate.py", line 15, in generate
    project.generate()
  File "/usr/local/lib/python3.7/site-packages/rpdk/core/project.py", line 243, in generate
    return self._plugin.generate(self)
[...]

PR #930, opened 2022-10-27
fix(contract): added regex for JSON string formats date-time, date, time, and email 

mmaeng:
Part of Issue #118 :

cfn was throwing `Key Error` during `cfn test` when using valid JSON string format types. e.g. "date-time" which is one of the examples generated during `cfn init`

added regex for

date-time
date
time
email


src/rpdk/core/contract/resource_generator.py
```diff
@@ -32,12 +33,12 @@
...
 # time is extraction from date-time
 STRING_FORMATS = {
-    "arn": "^arn:aws(-(cn|gov))?:[a-z-]+:(([a-z]+-)+[0-9])?:([0-9]{12})?:[^.]+$",
+    "arn": "^arn:aws(-(cn|us-gov))?:[a-z-]+:([a-z]{2}-([a-z]+-)+[0-9])+:([0-9]{12}){1}:[a-z0-9-:/*]+$",
     "uri": "^(https?|ftp|file)://[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
     "date-time": r"^(\d{4})-(0[1-9]|1[0-2])-(\d{2})T(?:[01]\d|2[0123]):(?:[0-5]\d):(?:[0-5]\d)(?:\.\d+)?(?:Z|[+-](?:[01]\d|2[0123]):[0-5]\d)$",
     "date": r"^(\d{4})-(0[1-9]|1[0-2])-(\d{2})$",
     "time": r"^(?:[01]\d|2[0123]):(?:[0-5]\d):(?:[0-5]\d)(?:\.\d+)?(?:Z|[+-](?:[01]\d|2[0123]):[0-5]\d)$",
-    "email": r"^.+@[^\.].*\.[a-z]{2,}$",
+    "email": r"^[a-zA-Z0-9+_-]+@[a-zA-Z0-9_-]+\.[a-zA-Z]{2,}$",
```
kddejong, line 41: I think there are a lot more characters that make an email address valid.  What is wrong with the current regex?

src/rpdk/core/contract/resource_generator.py
```diff
@@ -32,12 +33,12 @@
 # date is extraction from date-time
 # time is extraction from date-time
 STRING_FORMATS = {
-    "arn": "^arn:aws(-(cn|gov))?:[a-z-]+:(([a-z]+-)+[0-9])?:([0-9]{12})?:[^.]+$",
+    "arn": "^arn:aws(-(cn|us-gov))?:[a-z-]+:([a-z]{2}-([a-z]+-)+[0-9])+:([0-9]{12}){1}:[a-z0-9-:/*]+$",
```
kddejong, line 36: `([a-z]{2}-([a-z]+-)+[0-9])+` I think ? is more appropriate here? What am I missing?  Region can either be blank or a single region.

src/rpdk/core/contract/resource_generator.py
mmaeng: We can revert to the previous email regex.  Was attempting to reduce some invalid characters.

src/rpdk/core/contract/resource_generator.py
kddejong: `([0-9]{12}){1}` I think `?` is more accurate here.  There are valid ARNs that have no account IDs.
For example IAM AWS Managed policies have an ARN like `arn:aws:iam::aws:policy/AmazonSQSFullAccess`

mmaeng: I was basing it off https://docs.aws.amazon.com/general/latest/gr/aws-arns-and-namespaces.html i seem to have overlooked the global resources dont need the region.

Will fix.

kddejong: `[a-z0-9-:/*]+`  I think this is too restrictive too and I believe its missing a slash `[a-z0-9-:\/*]`

As an example a IAM role ARN typically includes the name:
`arn:aws:iam::account:role/role-name-with-path`
and from the documentation:
```
Names of users, groups, roles, policies, instance profiles, and server certificates must be alphanumeric, including the following common characters: plus (+), equal (=), comma (,), period (.), at (@), underscore (_), and hyphen (-).
```

I believe there is a unicode character thing here too.  I would make it more like what it was before.

src/rpdk/core/contract/resource_generator.py
```diff
@@ -257,4 +268,4 @@ def generate_string_strategy(schema):
             LOG.warning("found maxLength used with format")
 
         regex = STRING_FORMATS[string_format]
-        return from_regex(regex)
+        return from_regex(re.compile(regex, re.ASCII), fullmatch=True)
```
kddejong, line 271: fullmatch?

src/rpdk/core/contract/resource_generator.py
```diff
@@ -27,9 +28,17 @@
...
+# format list taken from https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.FormatChecker.checkers
+# date-time regex from https://github.com/naimetti/rfc3339-validator
+# date is extraction from date-time
+# time is extraction from date-time
 STRING_FORMATS = {
-    "arn": "^arn:aws(-(cn|gov))?:[a-z-]+:(([a-z]+-)+[0-9])?:([0-9]{12})?:[^.]+$",
-    "uri": "^(https?|ftp|file)://[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
+    "arn": r"^arn:aws(-(cn|us-gov))?:[a-z-]+:([a-z]{2}-([a-z]+-)+[0-9])?:([0-9]{12}){1}:[^.]+$",
+    "uri": r"^(https?|ftp|file)://[0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
+    "date-time": r"^(\d{4})-(0[1-9]|1[0-2])-(\d{2})T(?:[01]\d|2[0123]):(?:[0-5]\d):(?:[0-5]\d)(?:\.\d+)?(?:Z|[+-](?:[01]\d|2[0123]):[0-5]\d)$",
```
kddejong, line 38: I'm okay with these regex patterns just need to make sure the re.ASCII flag is set

src/rpdk/core/contract/resource_generator.py
```diff
@@ -247,7 +256,9 @@ def generate_string_strategy(schema):
             if "maxLength" in schema:  # pragma: no cover
                 LOG.warning("found maxLength used with pattern")
 
-            return from_regex(terminate_regex(regex))
+            return from_regex(
+                re.compile(terminate_regex(regex), re.ASCII), fullmatch=True
```
kddejong, line 260: terminate_regex replaces `$` with `\Z`... I wonder if should also replace `^` with `\A` then we would drop full match?

```
If you want the generated string to match the whole regex you should use boundary markers.
So e.g. r"\A.\Z" will return a single character string, while "." will return any string, and r"\A.$" 
will return a single character optionally followed by a "\n". Alternatively, passing 
fullmatch=True will ensure that the whole string is a match, as if you had used 
the \A and \Z markers.
```

src/rpdk/core/contract/resource_generator.py
```diff
@@ -247,7 +256,7 @@ def generate_string_strategy(schema):
             if "maxLength" in schema:  # pragma: no cover
                 LOG.warning("found maxLength used with pattern")
 
-            return from_regex(terminate_regex(regex))
+            return from_regex(re.compile(terminate_regex(regex), re.ASCII))
```
kddejong, line 259: should terminate regex also account for the `\A`

mmaeng: terminate_regex also set to replace `^` with `\A`

src/rpdk/core/contract/resource_generator.py
```diff
@@ -27,18 +28,28 @@
...
+# date is extraction from date-time
+# time is extraction from date-time
 STRING_FORMATS = {
     "arn": "^arn:aws(-(cn|gov))?:[a-z-]+:(([a-z]+-)+[0-9])?:([0-9]{12})?:[^.]+$",
-    "uri": "^(https?|ftp|file)://[0-9a-zA-Z]([-.\\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
+    "uri": r"^(https?|ftp|file)://[0-9a-zA-Z]([-.\w]*[0-9a-zA-Z])(:[0-9]*)*([?/#].*)?$",
+    "date-time": r"^(\d{4})-(0[1-9]|1[0-2])-(\d{2})T(?:[01]\d|2[0123]):(?:[0-5]\d):(?:[0-5]\d)(?:\.\d+)?(?:Z|[+-](?:[01]\d|2[0123]):[0-5]\d)$",
+    "date": r"^(\d{4})-(0[1-9]|1[0-2])-(\d{2})$",
+    "time": r"^(?:[01]\d|2[0123]):(?:[0-5]\d):(?:[0-5]\d)(?:\.\d+)?(?:Z|[+-](?:[01]\d|2[0123]):[0-5]\d)$",
+    "email": r"^.+@[^\.].*\.[a-z]{2,}$",
```
kddejong, line 41: I'm struggling finding these formats in the spec.  Can you help me find the list?

src/rpdk/core/contract/resource_generator.py
```diff
@@ -27,18 +28,28 @@
 # https://github.com/aws-cloudformation/aws-cloudformation-rpdk/issues/118
 
 # Arn is just a placeholder for testing
+# format list taken from https://python-jsonschema.readthedocs.io/en/stable/validate/#jsonschema.FormatChecker.checkers
+# date-time regex from https://github.com/naimetti/rfc3339-validator
+# date is extraction from date-time
+# time is extraction from date-time
 STRING_FORMATS = {
     "arn": "^arn:aws(-(cn|gov))?:[a-z-]+:(([a-z]+-)+[0-9])?:([0-9]{12})?:[^.]+$",
```
kddejong, line 36: from here should we be adding the rest of the formats? 
https://json-schema.org/understanding-json-schema/reference/string.html#id8

Missing ipv4, ipv6, hostname, and the internationalized versions.  We can table it as they come up just curious

kddejong (approved):

ericzbeard (approved):

src/rpdk/core/contract/resource_generator.py
mmaeng: We should add in the rest that are supported by cloudformation.  Let me create an issue for tracking about it.  The immediate fix was to stop the errors from the examples provided.

mmaeng: We should probably also move away from the .example setup since hypothesis docs recommends the @given syntax.

PR #931, opened 2022-11-01
fix(schema) - Use ASCII characters for \w in pattern regex

kddejong:
*Issue #, if available:*

*Description of changes:*
- \w, \W, \b, \B, \d, \D, \s and \S flags in regex should be ASCII based not unicode
- Unicode is the default in python so this switches it to be ascii based

### DIFF
--- a/src/rpdk/core/contract/resource_generator.py
+++ b/src/rpdk/core/contract/resource_generator.py
@@ -1,4 +1,5 @@
 import logging
+import re
 from collections.abc import Sequence
 
 from hypothesis.strategies import (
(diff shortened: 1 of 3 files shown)


kddejong:
More details:
https://docs.python.org/3/library/re.html#re.ASCII
https://262.ecma-international.org/5.1/#sec-15.10.2.12
http://json-schema.org/understanding-json-schema/reference/regular_expressions.html

mmaeng (approved):

ericzbeard (approved):

PR #934, opened 2022-11-03
Adding headers in STS Calls for Confused Deputy

himanshs23:
*Description of changes:*

1. Adding Headers in STS Calls for Confused Deputy Protection.
2. Test cases added
3. Will add more Unit Test post Integration Test

### DIFF
--- a/src/rpdk/core/boto_helpers.py
+++ b/src/rpdk/core/boto_helpers.py
@@ -32,19 +32,32 @@ def create_sdk_session(region_name=None):
     return session
 
 
-def get_temporary_credentials(session, key_names=BOTO_CRED_KEYS, role_arn=None):
+def get_temporary_credentials(
+    session, key_names=BOTO_CRED_KEYS, role_arn=None, headers=None
+):
     sts_client = session.client(
         "sts",
         endpoint_url=get_service_endpoint("sts", session.region_name),
         region_name=session.region_name,
     )
+
+    if headers and len(headers) == 2:
+        # Inject headers through the event system.
+        def inject_header(params):
+            params["headers"]["x-amz-source-account"] = headers[0]
+            params["headers"]["x-amz-source-arn"] = headers[1]
+
+        sts_client.meta.events.register("before-call", inject_header)
+
     if role_arn:
         session_name = "CloudFormationContractTest-{:%Y%m%d%H%M%S}".format(
             datetime.now()
         )
         try:
             response = sts_client.assume_role(
-                RoleArn=role_arn, RoleSessionName=session_name, DurationSeconds=900
+                RoleArn=role_arn,
+                RoleSessionName=session_name,
+                DurationSeconds=900,
             )
         except ClientError:
             # pylint: disable=W1201
--- a/src/rpdk/core/boto_helpers.py
+++ b/src/rpdk/core/boto_helpers.py
@@ -1,5 +1,6 @@
 import logging
 from datetime import datetime
+from typing import TypedDict, Union
 
 import botocore.loaders
 import botocore.regions
@@ -12,6 +13,11 @@ LOG = logging.getLogger(__name__)
 
 BOTO_CRED_KEYS = ("aws_access_key_id", "aws_secret_access_key", "aws_session_token")
 LOWER_CAMEL_CRED_KEYS = ("accessKeyId", "secretAccessKey", "sessionToken")
+
+
+class Headers(TypedDict):
+    source_account: str
+    source_arn: str
 
 
 def create_sdk_session(region_name=None):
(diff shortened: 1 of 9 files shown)


src/rpdk/core/boto_helpers.py
```diff
@@ -32,19 +32,32 @@ def _known_error(msg):
...
         endpoint_url=get_service_endpoint("sts", session.region_name),
         region_name=session.region_name,
     )
+
+    if headers and len(headers) == 2:
+        # Inject headers through the event system.
+        def inject_header(params):
+            params["headers"]["x-amz-source-account"] = headers[0]
+            params["headers"]["x-amz-source-arn"] = headers[1]
+
```
mmaeng, line 49: Wouldn't using a dictionary be preferable here to increase readability?  e.g. headers["source-account"] and headers["source-arn"]

src/rpdk/core/test.py
```diff
@@ -320,6 +320,7 @@ def get_contract_plugin_client(args, project, overrides, inputs):
             project.type_name,
             args.log_group_name,
             args.log_role_arn,
+            headers=(args.source_account, args.source_arn),
```
mmaeng, line 323: I believe headers will always not be None and have a length of 2 in this case (None, None) so it would always go into the inject_headers in boto_helpers.py

src/rpdk/core/test.py
```diff
@@ -341,6 +342,7 @@ def get_contract_plugin_client(args, project, overrides, inputs):
         project.type_name,
         args.log_group_name,
         args.log_role_arn,
+        headers=(args.source_account, args.source_arn),
```
mmaeng, line 345: I believe headers will always not be None and have a length of 2 in this case (None, None) so it would always go into the inject_headers in boto_helpers.py

tests/test_boto_helpers.py
```diff
@@ -200,6 +202,43 @@ def test_get_temporary_credentials_assume_role():
     assert tuple(creds.values()) == (access_key, secret_key, token)
 
 
+def test_get_temporary_credentials_assume_role_with_headers():
```
mmaeng, line 205: I believe there is two unit test scenarios not accounted for:
1. source_account is provided but no source_arn
2. source_arn is provided but no source_account

src/rpdk/core/test.py
```diff
@@ -441,6 +443,16 @@ def setup_subparser(subparsers, parents):
...
 
+    parser.add_argument(
+        "--source-account",
+        help="Source Account key used for Assume Role to Run Contract Tests",
+    )
+
+    parser.add_argument(
+        "--source-arn",
+        help="Source Type Version Arn key used for Assume Role to Run Contract Tests",
+    )
```
ammokhov, line 454: do you need to add basic validation? what are the implications if incorrect entities are provided

ericzbeard (changes requested):
There is a lot of context here that is invisible to the public. Since this is an open source repo, we should clearly document all of that context in the PR. We should also clearly document how this changes the behavior of the CLI, and how users need to adjust their usage of it.

src/rpdk/core/test.py
kddejong: why do you think this?

kddejong:
Adding documentation links for this:
https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/registry.html#cross-service-confused-deputy-prevention

src/rpdk/core/test.py
kddejong: Good catch.  (None, None) will still be a length of 2.

src/rpdk/core/test.py
kddejong: got it now :)

src/rpdk/core/boto_helpers.py
```diff
@@ -33,19 +39,24 @@ def _known_error(msg):
...
     sts_client = session.client(
         "sts",
         endpoint_url=get_service_endpoint("sts", session.region_name),
         region_name=session.region_name,
     )
 
-    if headers and len(headers) == 2:
+    if headers and headers.get("source_account") and headers.get("source_arn"):
         # Inject headers through the event system.
+        # https://docs.aws.amazon.com/IAM/latest/UserGuide/confused-deputy.html
```
ericzbeard, line 55: These links do not tell me why we are adding these headers.

ericzbeard:
@himanshs23  Please add a detailed description to the PR documenting why we need the change, what effect it has on the CLI, what changes users need to make, etc. We need to understand the specific threat that is being mitigated, and how it changes the configuration and behavior of resource handlers.

ericzbeard:
Closing for now while we re-evaluate this fix.

PR #970, opened 2023-01-19
Add in no-docker and use-docker override to submit command

### LINKED ISSUE #969: Allow specifying --[no-]-use-docker via CLI during packaging / submit
When building hooks or resources we often want to use a different build configuration on a developers desktop versus in CI. Today I can only specify whether or not builds should happen in a docker container during `cfn init` or by modifying the `.rpdk-config` file, which makes it difficult to configure developers to build using docker on their local machine but have our CI workflow build without docker, as the CI workflow may already be running in a container. I can achieve this today by writing a wrapper that injects different config depending on the build environment, but it would be much easier if I could override the `use_docker` flag at build / packaging time.

Example CLI usage:
```
# Create a new project, setting use_docker to true in the initial config
cfn init

# Validate and package using the project config (build in docker container)
cfn submit --dry-run

# Validate and package without using docker, overriding the project settings, e.g. in a CI container
cfn submit --dry-run --no-use-docker
```

mmaeng:
Closes #969 

Add in CLI packaging overrides to submit command for `cfn submit --no-docker` and `cfn submit --use-docker`

This will override the use_docker settings in .rpdk-config file when building the package.

### DIFF
--- a/src/rpdk/core/submit.py
+++ b/src/rpdk/core/submit.py
@@ -12,7 +12,7 @@ LOG = logging.getLogger(__name__)
 def submit(args):
     project = Project()
     project.load()
-    # Update config file if use-docker or no-docker used in submit call
+    # Use CLI override opposed to config file if use-docker or no-docker switch used
     if args.use_docker or args.no_docker:
         project.settings["use_docker"] = args.use_docker
         project.settings["no_docker"] = args.no_docker
--- a/src/rpdk/core/submit.py
+++ b/src/rpdk/core/submit.py
@@ -12,6 +12,10 @@ LOG = logging.getLogger(__name__)
 def submit(args):
     project = Project()
     project.load()
+    # Update config file if use-docker or no-docker used in submit call
+    if args.use_docker or args.no_docker:
+        project.settings["use_docker"] = args.use_docker
+        project.settings["no_docker"] = args.no_docker
     project.submit(
         args.dry_run,
         args.endpoint_url,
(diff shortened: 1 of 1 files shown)


src/rpdk/core/submit.py
```diff
@@ -12,6 +12,10 @@
 def submit(args):
     project = Project()
     project.load()
+    # Update config file if use-docker or no-docker used in submit call
+    if args.use_docker or args.no_docker:
+        project.settings["use_docker"] = args.use_docker
```
seawildau, line 17: If I understand correctly, feeding this CLI arg into `project.settings` will write the new flags to `.rpdk-config.json` and get picked up by future CLI commands. This may not be desirable if the user only wants to override this setting for a single run. Perhaps a better solution would be to extend project settings with a flag or something that lets you control whether the setting gets persisted or not.

Also, I notice that only the python plugin uses the `use_docker` setting today. The typescript plugin uses `useDocker`, and as far as I can tell neither the java or go plugins support building in a docker container at all. I think it's worth considering whether this behaviour can be standardised across all plugins (but that could be a future enhancement / PR).

mmaeng: Typescript was updated to match the other two plugins a little while back and I tested this update on python and typescript extensions (hooks/resources)  I don't have a golang extension handy but it should also apply.

Since config file actually only gets written during the "cfn init" portion of things.  The override won't update the .rpdk-config file.

mmaeng: BTW I'm basing this off the current master branch of the typescript plugin.  The next release of the typescript plugin will have this use_docker update.

mmaeng: Whoops just noticed that comment, let me update that to be accurate :)

ericzbeard (approved):

kddejong (approved):

Issue #960, opened 2022-12-29
During cfn init for hook - obtuse error when no valid creds


mmaeng:
Clean up the very verbose error when no creds are detected during cfn init hook.

Issue #959, opened 2023-01-03
cfn init shouldn't need AWS access or credentials to create example templates for hook/resource/modules


mmaeng:
cfn init

should be stand alone and able to be done offline.  It has no need to access AWS accounts but currently not having your environment set up will cause an exception for all languages.

benbridts:
Isn't this already the case (for `init`)?

on my machine:
```
~$ cfn --version
cfn 0.2.28

~$ aws sts get-caller-identity
Unable to locate credentials. You can configure credentials by running "aws configure".

~$ cfn init
Initializing new project
Do you want to develop a new resource(r) or a module(m) or a hook(h)?.
>> m
What's the name of your module type?
(<Organization>::<Service>::<Name>::MODULE)
>> Example::Service::Name::MODULE
Directory  /Users/ben/tmp/fragments  Created
Initialized a new project in /Users/ben/tmp

~$ ls
fragments/ rpdk.log
```

mmaeng:
Thanks Ben!

For modules and resources that looks to be the case but for hooks it errors, and then proceeds to create the example anyway :P

```
$ cfn init -a hook -t test::tes~t::test python38 --use-docker
Initializing new project
No region specified. Please ensure your AWS CLI is configured correctly: https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html

$ ls
README.md           hook-role.yaml      requirements.txt    rpdk.log            src                 template.yml        test-test-test.json

$ cfn init --region us-east-1 -a hook -t test::test::test python38 --use-docker
Initializing new project
No credentials specified. Please ensure your AWS CLI is configured correctly: https://docs.aws.amazon.com/cli/latest/userguide/cli-chap-configure.html

$ ls
README.md           hook-role.yaml      requirements.txt    rpdk.log            src                 template.yml        test-test-test.json
```

PR #995, opened 2023-05-02
Update hook project init to use only plugins that support hooks and fixing issue with init when no creds are specified

Brianwithay21:
*Issue #, if available:*
* https://github.com/aws-cloudformation/cloudformation-cli/issues/959
* https://github.com/aws-cloudformation/cloudformation-cli/issues/960

*Description of changes:*
Updated the `cfn init` command for `HOOK` type projects to only allow projects to be initialized with language plugins that support `HOOK` types.

This PR also fixes an issue with error being thrown when project was initialized with no AWS credentials set

### DIFF
--- a/tests/hook/test_init_hook.py
+++ b/tests/hook/test_init_hook.py
@@ -71,7 +71,7 @@ def test_validate_plugin_choice_greater_than_choice():
 
 
 def test_validate_plugin_choice_valid():
-    choices = ["python38", PROMPT, "pyton39"]
+    choices = ["python38", PROMPT, "python39"]
     validator = ValidatePluginChoice(choices)
     assert validator("2") == PROMPT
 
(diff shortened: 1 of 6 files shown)


tests/hook/test_init_hook.py
```diff
@@ -70,7 +71,7 @@ def test_validate_plugin_choice_greater_than_choice():
 
 
 def test_validate_plugin_choice_valid():
-    choices = ["1", PROMPT, "2"]
+    choices = ["python38", PROMPT, "pyton39"]
```
reza-aws, line 74: Last choice should be "python39" -- missing an 'h'

reza-aws (changes requested):

tests/hook/test_init_hook.py
Brianwithay21: fixed typo

reza-aws (approved):

wkexiang (approved):

Issue #998, opened 2023-05-11
Unable to set stackId when testing with cfn invoke


chamcca:
When doing local testing of RESOURCEs, `cfn invoke` does not provide the ability to set a `stackId` on the request. This causes any handlers relying on the `stackId` to fail or incorporate special handling for missing `stackId` values. 

Ideally, should be able to pass JSON like:
```json
{
    "stackId": "test-stack",
    "desiredResourceState": {
        "Key1": "Value1",
        "Key2": "Value2"
    },
    "previousResourceState": {},
    "logicalResourceIdentifier": "SomeIdentifier"
}
```

PR #999, opened 2023-05-11
fix: stackId not setable during 'cfn invoke' testing of RESOURCE

chamcca:
*Issue #, if available:* Fixes: #998

*Description of changes:*
- optionally create a request containing `stackId` during `cfn invoke` if it is included in the request JSON

### DIFF
--- a/src/rpdk/core/invoke.py
+++ b/src/rpdk/core/invoke.py
@@ -110,6 +110,7 @@ def invoke(args):
             request["desiredResourceState"],
             request["previousResourceState"],
             request.get("typeConfiguration"),
+            stackId=request.get("stackId")
         )  # pylint: disable=too-many-function-args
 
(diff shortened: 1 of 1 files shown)


src/rpdk/core/invoke.py
```diff
@@ -110,6 +110,7 @@ def invoke(args):
             request["desiredResourceState"],
             request["previousResourceState"],
             request.get("typeConfiguration"),
+            stackId=request.get("stackId")
```
ammokhov, line 113: Stack id is not always present, as this is purely CloudFormation concept. Resource Handlers are also used by CloudControl, which does not provide value for this field. I would prefer to left it out of `cfn invoke` payload.

ammokhov (changes requested):

PR #1003, opened 2023-05-25
Add permissions to tagging schema

yueful27:
Update tagging schema with permissions field.

Add unit test for tagging permissions field.

### DIFF
--- a/src/rpdk/core/contract/resource_client.py
+++ b/src/rpdk/core/contract/resource_client.py
@@ -837,6 +837,12 @@ class ResourceClient:  # pylint: disable=too-many-instance-attributes
         except KeyError:
             return False
 
+    def metadata_contains_tag_permissions(self):
+        try:
+            return "permissions" in self._schema["tagging"]
+        except KeyError:
+            return False
+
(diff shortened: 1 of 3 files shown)


WaelA (approved):

AidenAtSea (approved):

src/rpdk/core/contract/resource_client.py
```diff
@@ -837,6 +837,12 @@ def metadata_contains_tag_property(self):
         except KeyError:
             return False
 
+    def metadata_contains_tag_permissions(self):
+        try:
+            return "permissions" in self._schema["tagging"]
+        except KeyError:
+            return False
+
```
ammokhov, line 845: do we need it?
```suggestion
    def metadata_contains_tag_permissions(self):
        return "permission" in self._schema.get("tagging", {})
```

ammokhov: lets add a method that can access actual permission from the list + add appropriate unit test for that. Other than that it looks good to me

ammokhov (approved):

PR #1020, opened 2023-08-23
Support for extending cfn cli with custom commands

TheChouzanOne:
*Issue #, if available:* NA

*Description of changes:* 

This change enables custom plugins to be developed under the `rpdk.v1.extensions` entrypoint to extend `cfn` functionality. E.g. developing a new `hello` command:
```
cfn hello --name Adrian

Hello, Adrian!
```

In this change, I introduce the `ExtensionPlugin` abstract base class, which has 2 methods. `command_name` determines what the command will be (e.g. `hello` in the above example). This registers a subparser under `cfn`. The second method is `setup_parser` which should be implemented in the child class to setup any arguments and behaviors required (e.g. `--name` parameter and printing `Hello, Adrian!` in examble above)

The way a new extension can be setup is the same way as other plugins already developed, where you simply `pip install` the package in your environment.

### DIFF
--- a/src/rpdk/core/extensions.py
+++ b/src/rpdk/core/extensions.py
@@ -1,0 +1,8 @@
+from .plugin_registry import get_extensions
+
+def setup_subparsers(subparsers, parents):
+    extensions = get_extensions()
+
+    for extension_cls in extensions.values():
+        extension = extension_cls()()
+        extension.setup_subparser(subparsers, parents)
--- a/src/rpdk/core/extensions.py
+++ b/src/rpdk/core/extensions.py
@@ -1,10 +1,18 @@
 from .plugin_registry import get_extensions
 
 
+def _check_command_name_collision(subparsers, command_name):
+    if command_name in subparsers.choices:
+        raise RuntimeError(
+            f'"{command_name}" is already registered as an extension. Please use a different name.'
+        )
+
+
 def setup_subparsers(subparsers, parents):
     extensions = get_extensions()
 
     for extension_cls in extensions.values():
         extension = extension_cls()()
+        _check_command_name_collision(subparsers, extension.command_name)
         parser = subparsers.add_parser(extension.command_name, parents=parents)
         extension.setup_parser(parser)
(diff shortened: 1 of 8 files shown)


src/rpdk/core/extensions.py
```diff
@@ -0,0 +1,10 @@
+from .plugin_registry import get_extensions
+
+
+def setup_subparsers(subparsers, parents):
+    extensions = get_extensions()
+
+    for extension_cls in extensions.values():
+        extension = extension_cls()()
+        parser = subparsers.add_parser(extension.command_name, parents=parents)
```
joshuadeanhall, line 9: Few comments about this.

1. Is there any risk of name collision?
2. Can an extension hijack a subparser "name" and override its behavior

src/rpdk/core/extensions.py
```diff
@@ -0,0 +1,10 @@
...
+        extension.setup_parser(parser)
```
joshuadeanhall, line 10: What can the extension do given the subparser?

Can it access the parent from the subparse and start mutating the parent in unexpected ways?

src/rpdk/core/extensions.py
TheChouzanOne: Yes to both.

1. If there is a name collision, the latest one that was executed (`.add_parser`) will take precedence.
2. Given the answer from 1, yes, it is possible.

I can add error handling to verify all the options that have been added and raise an exception if there's a collision

src/rpdk/core/extensions.py
TheChouzanOne: From `argparse` [documentation](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_subparsers), it doesn't look like a subparser can access the parent parser. The closest is the `parents` argument which just tells the subparser to inherit the parents' arguments as well (this is something we might want to remove for this case, actually).

src/rpdk/core/extensions.py
TheChouzanOne: Added check in [37b45ce](https://github.com/aws-cloudformation/cloudformation-cli/pull/1020/commits/37b45ce2888cc2dc0ad17b919d4adbdf094037f8)

kddejong (approved):

joshuadeanhall (approved):

PR #1027, opened 2023-10-02
Fix prepare_payload_for_reinvocation in invoke

mircealam:
*Description of changes:*

This is for the use case where we are leveraging cfn invoke.
Before the fix only the callbackContext was passedin leading to a situation where if the model is updated and the handler emits a progress event we would end up losing recoding the updates that were made (and making the handler fail for those instances).  
One such example is when we are creating a resources and there are multiple API calls needed to materialize this resource with a progress event being emitted between this API calls and one of the later API calls needed info returned by the earlier calls and persisted in the model.

    Validation:
    seed isort known_third_party.............................................Passed
    isort....................................................................Passed
    black....................................................................Passed
    check for case conflicts.................................................Passed
    fix end of files.........................................................Passed
    mixed line ending........................................................Passed
    trim trailing whitespace.................................................Passed
    pretty format json...................................(no files to check)Skipped
    check for merge conflicts................................................Passed
    flake8...................................................................Passed
    check blanket noqa.......................................................Passed
    check for not-real mock methods..........................................Passed
    use logger.warning(......................................................Passed
    bandit...................................................................Passed
    pylint-local.............................................................Passed
    pytest-local.............................................................Passed

    In addition ran the cfn invoke and verified that the model is correctly passed back in

### DIFF
--- a/src/rpdk/core/invoke.py
+++ b/src/rpdk/core/invoke.py
@@ -76,7 +76,7 @@ def get_contract_client(args, project):
 def prepare_payload_for_reinvocation(payload, response, artifact_type):
     if artifact_type == ARTIFACT_TYPE_RESOURCE:
         payload["callbackContext"] = response.get("callbackContext")
-        if response.get("resourceModel") and payload.get("requestData"):
+        if ("resourceModel" in response) and ("requestData" in payload):
             payload["requestData"]["resourceProperties"] = response.get("resourceModel")
 
     return payload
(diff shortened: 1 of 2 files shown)


src/rpdk/core/invoke.py
```diff
@@ -76,6 +76,8 @@ def get_contract_client(args, project):
 def prepare_payload_for_reinvocation(payload, response, artifact_type):
     if artifact_type == ARTIFACT_TYPE_RESOURCE:
         payload["callbackContext"] = response.get("callbackContext")
+        if ("resourceModel" in response) and ("requestData" in payload):
+            payload["requestData"]["resourceProperties"] = response.get("resourceModel")
```
ammokhov, line 80: Should we default to a previous one of its not returned? None properties might lead to confusing results. 
`payload["requestData"]["resourceProperties"] = response.get("resourceModel") or payload["requestData"]["resourceProperties"]`

ammokhov (approved):

src/rpdk/core/invoke.py
mircealam: If it's not returned we wouldn't touch the payload.

kddejong (approved):

PR #1044, opened 2023-11-15
Add python 3.12 support

kddejong:
*Issue #, if available:*

*Description of changes:*
- Add python 3.12 support

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -61,6 +61,8 @@ RUNTIME = random.choice(
         "go1.x",
         "python3.8",
         "python3.9",
+        "python3.10",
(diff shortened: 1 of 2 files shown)


tests/test_project.py
```diff
@@ -61,6 +61,8 @@
         "go1.x",
         "python3.8",
         "python3.9",
+        "python3.10",
+        "python3.11",
```
ericzbeard, line 65: What about 12?

kddejong: Weird dependency issue.  It requires a released version of the python plugin that has support for python 3.12 which isn't released yet.

PR #663, opened 2021-01-08
catch common resource schema issues

PatMyron:
**resource types with LIST API options like `MaxResults`, `NextToken`, `Filters`, etc as properties of the resource itself:**
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-licensemanager/pull/3
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-iotwireless/pull/3
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-ssm/pull/75

```shell
botocore $ grep -h 'input_token' botocore/data/*/*/paginators-1.json | tr '[:upper:]' '[:lower:]' | sort | uniq -ic | sort -nr | head
1147       "input_token": "nexttoken",
 328       "input_token": "marker",
  40       "input_token": "pagetoken",
  18       "input_token": "position",
  16       "input_token": "nextmarker",
   7       "input_token": "nextpagetoken",
   5       "input_token": "paginationtoken",

botocore $ grep -h 'output_token' botocore/data/*/*/paginators-1.json | tr '[:upper:]' '[:lower:]' | sort | uniq -ic | sort -nr | head
1147       "output_token": "nexttoken",
 199       "output_token": "marker",
  57       "output_token": "nextmarker",
  45       "output_token": "nextpagetoken",

botocore $ grep -h 'limit_key' botocore/data/*/*/paginators-1.json | tr '[:upper:]' '[:lower:]' | sort | uniq -ic | sort -nr | head
 931       "limit_key": "maxresults",
 159       "limit_key": "maxrecords",
 144       "limit_key": "limit",
 129       "limit_key": "maxitems",
  42       "limit_key": "pagesize",
  ```

---

**[`readOnlyProperties` overlapping with `writeOnlyProperties` or `createOnlyProperties`](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#resource-semantics):**
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-auditmanager/pull/2
https://github.com/aws-cloudformation/aws-cloudformation-resource-providers-codeartifact/pull/38
[`AWS::Events::Archive.ArchiveName`](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-events-archive.html#cfn-events-archive-name)

---

**wildcards in [handler permissions](https://github.com/aws-cloudformation/aws-cloudformation-resource-schema/#permissions)**:
[...]

PR #1061, opened 2024-03-07
Adding headers in STS Calls for Confused Deputy

saieshwarm:
Adding headers in STS Calls for Confused Deputy to accept account id and the source ARN

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -12,6 +12,7 @@ disable=
     duplicate-code,             # finds dupes between tests and plugins
     too-few-public-methods,     # triggers when inheriting
     ungrouped-imports,          # clashes with isort
+    W0613                       # Unused argument 'kwargs'
 
 [BASIC]
 
@@ -23,4 +24,5 @@ indent-string='    '
 max-line-length=160
 
 [DESIGN]
-max-locals=16
+max-locals=17
+max-args=6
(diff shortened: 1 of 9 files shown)


kddejong (approved):

.pylintrc
```diff
@@ -23,4 +24,5 @@ indent-string='    '
 max-line-length=160
```
ericzbeard, line 24: Can we add an exception to just the one line if it's not possible to break it down? 160 is very long.

ericzbeard (commented):
Looks good, just one question about pylintrc

.pylintrc
saieshwarm: This exception is from a different PR(PR No:663).
https://github.com/aws-cloudformation/cloudformation-cli/pull/663/files#diff-ba293004ae8f18975051e825bd8d5c70137a8b5d42d16e0f7c9354aa634c446c

src/rpdk/core/boto_helpers.py
```diff
@@ -32,17 +32,35 @@ def _known_error(msg):
...
+        and headers["account_id"]
+        and headers["source_arn"]
+    ):
+        # Inject headers through the event system.
+        def inject_confused_deputy_headers(params, **kwargs):
+            params["headers"]["x-amz-source-account"] = headers["account_id"]
+            params["headers"]["x-amz-source-arn"] = headers["source_arn"]
+
+        sts_client.meta.events.register("before-call", inject_confused_deputy_headers)
+        LOG.info(headers)
```
himanshs23, line 56: do we need to still log it

PR #1074, opened 2024-06-05
Implement canary file generation functionality from contract test PatchInputs

marc-1010:
*Issue #, if available:*

*Description of changes:*
Add canary file generation functionality for Patch Operation canaries from contract test inputs.
canarySettings will be shared with generating Create Operation canaries. 

Testing Scenarios:
1. Validate that the correct number of output files are generated. Validated that file naming and structure are correct.
2. Validate that an existing field in the `CreateInputs` will be updated by a `replace`, `add` or `remove` operation in `PatchInputs`. 
3. Validate that only `replace`, `add` and `remove` operations result in an update. Other operations such as `move`, `copy` and `test` are ignored.
4. Validate that a field can be updated if it was not included in `CreateInputs` but included in `PatchInputs` for `add` and `replace` operations. 
5. Validate that nested inputs are correctly updated. 

### DIFF
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -1,4 +1,5 @@
 # pylint: disable=too-many-lines
+import copy
 import json
 import logging
 import os
@@ -76,7 +77,6 @@ CONTRACT_TEST_DEPENDENCY_FILE_NAME = "dependencies.yml"
 FILE_GENERATION_ENABLED = "file_generation_enabled"
 TYPE_NAME = "typeName"
 CONTRACT_TEST_FILE_NAMES = "contract_test_file_names"
-INPUT1_FILE_NAME = "inputs_1.json"
 FN_SUB = "Fn::Sub"
 FN_IMPORT_VALUE = "Fn::ImportValue"
 UUID = "uuid"
@@ -1345,21 +1345,64 @@ class Project:  # pylint: disable=too-many-instance-attributes,too-many-public-m
             with ct_file.open("r") as f:
                 json_data = json.load(f)
             resource_name = self.type_info[2]
-            stack_template_data = {
-                "Description": f"Template for {self.type_name}",
-                "Resources": {
-                    f"{resource_name}": {
-                        "Type": self.type_name,
-                        "Properties": self._replace_dynamic_values(
-                            json_data["CreateInputs"]
-                        ),
-                    }
-                },
-            }
-            stack_template_file_name = f"{CANARY_FILE_PREFIX}{count}_001.yaml"
-            stack_template_file_path = stack_template_folder / stack_template_file_name
-            with stack_template_file_path.open("w") as stack_template_file:
-                yaml.dump(stack_template_data, stack_template_file, indent=2)
+
+            if "PatchInputs" in json_data:
+                deepcopy_create_input = copy.deepcopy(json_data["CreateInputs"])
+                self._save_stack_template_data(
+                    resource_name,
+                    count,
+                    stack_template_folder,
+                    self._apply_patch_inputs_to_create_inputs(
+                        json_data["PatchInputs"], deepcopy_create_input
+                    ),
+                    "002",
+                )
+            self._save_stack_template_data(
+                resource_name,
+                count,
+                stack_template_folder,
+                json_data["CreateInputs"],
+                "001",
+            )
+
+    def _save_stack_template_data(
+        self,
+        resource_name,
+        contract_test_input_count,
+        stack_template_folder,
+        properties_data,
+        suffix,
+    ):
+        stack_template_data = {
+            "Description": f"Template for {self.type_name}",
+            "Resources": {
+                f"{resource_name}": {
+                    "Type": self.type_name,
+                    "Properties": self._replace_dynamic_values(
+                        properties_data,
+                    ),
+                }
+            },
+        }
+        stack_template_file_name = (
+            f"{CANARY_FILE_PREFIX}{contract_test_input_count}_{suffix}.yaml"
+        )
+        stack_template_file_path = stack_template_folder / stack_template_file_name
+        with stack_template_file_path.open("w") as stack_template_file:
+            yaml.dump(stack_template_data, stack_template_file, indent=2)
+
+    def _apply_patch_inputs_to_create_inputs(
+        self, patch_inputs: Dict[str, Any], create_inputs: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        output = create_inputs
+        for patch_input in patch_inputs:
+            if patch_input["op"] == "replace":
+                key_list = patch_input["path"].split("/")
+                current_output = output
+                for key in key_list[:-1]:
+                    current_output = current_output.setdefault(key, {})
+                current_output[key_list[-1]] = patch_input["value"]
+        return output
 
     def _replace_dynamic_values(self, properties: Dict[str, Any]) -> Dict[str, Any]:
         for key, value in properties.items():
--- a/src/rpdk/core/project.py
+++ b/src/rpdk/core/project.py
@@ -1,5 +1,4 @@
 # pylint: disable=too-many-lines
-import copy
 import json
 import logging
 import os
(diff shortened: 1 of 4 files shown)


rajdnp (approved):

src/rpdk/core/project.py
```diff
@@ -1345,21 +1345,91 @@ def _generate_stack_template_files(self) -> None:
...
-                yaml.dump(stack_template_data, stack_template_file, indent=2)
+
+            self._save_stack_template_data(
+                resource_name,
+                count,
+                stack_template_folder,
+                self._replace_dynamic_values(
+                    json_data["CreateInputs"],
+                ),
+                "001",
```
joshuadeanhall, line 1356: What does 001 and 002 represent?

src/rpdk/core/project.py
```diff
@@ -1345,21 +1345,91 @@ def _generate_stack_template_files(self) -> None:
...
+        self, patch_inputs: Dict[str, Any], create_inputs: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        for patch_input in patch_inputs:
+            self._apply_patch_input_to_create_input(patch_input, create_inputs)
+        return create_inputs
+
+    def _apply_patch_input_to_create_input(
+        self, patch_input: Any, create_inputs: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        op = patch_input.get("op")
```
joshuadeanhall, line 1405: Instead of manually handling patch operation was any investigation done into using existing tools for this? For example `jsonpatch`?

src/rpdk/core/project.py
```diff
@@ -1345,21 +1345,91 @@ def _generate_stack_template_files(self) -> None:
...
+                self._save_stack_template_data(
+                    resource_name,
+                    count,
+                    stack_template_folder,
+                    self._apply_patch_inputs_to_create_inputs(
+                        json_data["PatchInputs"],
+                        deepcopy_create_input,
+                    ),
+                    "002",
+                )
```
ammokhov, line 1369: This should not be so hardcoded

src/rpdk/core/project.py
```diff
@@ -1345,21 +1345,91 @@ def _generate_stack_template_files(self) -> None:
...
+    def _apply_patch_inputs_to_create_inputs(
+        self, patch_inputs: Dict[str, Any], create_inputs: Dict[str, Any]
+    ) -> Dict[str, Any]:
+        for patch_input in patch_inputs:
+            self._apply_patch_input_to_create_input(patch_input, create_inputs)
+        return create_inputs
+
+    def _apply_patch_input_to_create_input(
+        self, patch_input: Any, create_inputs: Dict[str, Any]
+    ) -> Dict[str, Any]:
```
ammokhov, line 1404: please use jsonpatch library instead

src/rpdk/core/project.py
marc-1010: `001` represents a canary associated with a "create" operation. `002` represents a canary associated with an "update" operation. I'll refactor to have these values as constants.

ammokhov (changes requested):
This PR needs some work. I think we should have a method that will translate desired state (dict) into a stack template (which we already have). Then you should prepare a list of such structures. You should iterate over each `CreateInputs` and apply `PatchInputs` then we can loop through all of them and generate stack template with names added incrementally.


Example:
```json
{
    "CreateInputs": {
        "property1": "value1",
        "property2": "value1",
        "property3": "value1",
    },
    "PatchInputs": [
        {
            "op": "add",
            "path": "/property4",
            "value": "value1"
        },
        {
            "op": "replace",
            "path": "/property1",
            "value": "value2"
        }
    ]
}
```
those inputs will be translated into this
```
[
    {
        "property1": "value1",
        "property2": "value1",
        "property3": "value1"
    },
    {
        "property1": "value2",
        "property2": "value1",
        "property3": "value1",
        "property4": "value1",
    }
]
```
then you can iterate and run it through stack generation method

```python
def apply_patch(document:dict, patch_document: Sequence[Dict]):
    modified_document = jsonpatch.JsonPatch(patch_document).apply(document)
    ...

def generate_canaries(...):
    inputs = get_list_of_inputs(...)
    for input in inputs:
        canary_file = generate_canary(input)
```

src/rpdk/core/project.py
marc-1010: i considered using `pydash`. i think `jsonpatch` would be best here

src/rpdk/core/project.py
marc-1010: refactored to use defined constants

marc-1010:
Summary of Updates:
- Refactored hardcoded values into defined constants
- Refactored to use jsonpatch and removed unnecessary functions. A noteworthy changes are listed below:
	- jsonpath will enforce path validation. If a path is invalid, an exception will be raised. Previous behavior of skipping updates with invalid paths could lead consumers to ignore operations that are not passing.
	- Contract test PatchInputs.path values should be relative to the “Properties” key and include a leading slash(“/“). This behavior will be enforced in canary generation. 
- Revised algorithm for PatchInputs
	- Filter for supported “op” (ie remove, replace and add)
	- Translate “value” to dynamic values
	- use jsonpatch to apply translated patch(which includes dynamic values) to CreateInputs
	- Save

src/rpdk/core/project.py
```diff
@@ -1345,21 +1352,68 @@ def _generate_stack_template_files(self) -> None:
...
+            "Description": f"Template for {self.type_name}",
+            "Resources": {
+                f"{resource_name}": {
+                    "Type": self.type_name,
+                    "Properties": properties_data,
+                }
+            },
+        }
+        stack_template_file_name = (
+            f"{CANARY_FILE_PREFIX}{contract_test_input_count}_{suffix}.yaml"
```
joshuadeanhall, line 1398: This will create files that look like

- For Create
    - AWS::S3::Bucket1001.yaml
    - AWS::S3::Bucket2001.yaml
    - AWS::S3::Bucket3001.yaml
- For Update
    - AWS::S3::Bucket1002.yaml
    - AWS::S3::Bucket2002.yaml
    - AWS::S3::Bucket3002.yaml
    
Is there a reason we didn't instead using a convention more like `{CANARY_FILE_PREFIX}[UPDATE|CREATE]{contract_test_input_count::padding3}.yaml`?

That way names would look like `AWS::S3::BucketUpdateCREATE001`, `AWS::S3::BucketUpdateCREATE002` ...

marc-1010: We choose to favor backwards compatibility over changing naming. Since existing canaries all use the `canary1_001`, `canary1_002`, etc. convention, we wanted auto generated canaries to be consistent with the existing naming convention. If we were to change naming conventions, I think it would be advisable to migrate historical canaries to the new naming convention.

joshuadeanhall (approved):

src/rpdk/core/project.py
```diff
@@ -1345,21 +1352,68 @@ def _generate_stack_template_files(self) -> None:
...
-                yaml.dump(stack_template_data, stack_template_file, indent=2)
+
+            self._save_stack_template_data(
+                resource_name,
+                count,
+                stack_template_folder,
+                self._replace_dynamic_values(
+                    json_data[CREATE_INPUTS_KEY],
+                ),
+                CANARY_FILE_CREATE_SUFFIX,
```
ammokhov, line 1363: could you please explain why this needs to be hardcoded? Not sure this is the best customer experience. Can this be something configurable by the customer?


Ideally canaries should not test every functionality of the resource, this is what integration tests are for. However, if we dynamically assign names `001/00*` then it might got too far as providers have no upper limit of how many inputs they can specify. Can we introduce a setting inside rpdk config file where providers can configure how many canaries they can generate? Sometime you need more than just 2 files

ammokhov (changes requested):

ammokhov (approved):

src/rpdk/core/project.py
marc-1010: Some customer experience optimizations will be proposed/reviewed in a separate PR

PR #1085, opened 2024-07-30
Refactor canary auto-generation cli semantics

marc-1010:
*Issue #, if available:*

*Description of changes:*
- Canary auto generation will be disabled for empty canary settings. This is intended to prevent auto generation when a key is added to the `.rpdk-config` when providers are upgrading their rpdk version set.
- Add an argument `local-code-generation` to the `cfn generate` command to enable canary auto generation. This intended to restrict canary autogeneration to specific development use cases. Canary auto-generation should not occur for remote builds. 

Unit Test Cases:
- Canary autogeneration will not occur if the `local-code-generation` flag is missing or False. 
- Add `local-code-generation=True` to existing unit tests

### DIFF
--- a/src/rpdk/core/generate.py
+++ b/src/rpdk/core/generate.py
@@ -20,7 +20,9 @@ def generate(args):
         args.profile,
     )
     project.generate_docs()
-    project.generate_canary_files()
+    project.generate_canary_files(
+        args.local_code_generation,
+    )
     LOG.warning("Generated files for %s", project.type_name)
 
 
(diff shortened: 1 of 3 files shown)


ammokhov (commented):
I am confused what benefits we get from having `local-code-generation`

marc-1010:
> I am confused what benefits we get from having `local-code-generation`

There is no valid use case/need for re-generating canary code during remote builds. Canaries are not assumed to be a build time generated-source. This cli argument will prevent remote builds from re-generation/overriding canary code during remote host build time. I don't think the burden should be on providers to manually to disable canary generation completely in .rpdk-config to prevent behavior that we know to be unnecessary or problematic.

agarikana (approved):

AidenAtSea (approved):

